PR #7, opened 2019-04-10
Feature/registration

breneser:


trbateman:
Can you use: https://pypi.org/project/django-environ/ for the setings?

trbateman:
Can we make separate settings for production.py, staging.py and development.py using the common pattern where base.py exists?

trbateman:
My docker container dies on docker up as postgres hasn't yet started. Can you add https://github.com/jwilder/dockerize with a timeout so the web doesn't start until the db is up?

PR #31, opened 2019-11-04
Feature/import application create

breneser:


web/static/web/js/main.js
```diff
@@ -170,6 +170,19 @@
...
+            var form =$(this).closest('form');
+            $('<input />')
+              .attr('type','hidden')
+              .attr('name', 'change')
+              .attr('value', 'True')
+              .appendTo(form);
+            form.submit();
+          }; 
+          $('#id_application_type, #id_importer').change(post_form);
+        }
```
bsk: I'm curious why is this needed.

PR #33, opened 2019-11-11
Use password_disposition, add unsuccessfull_login_attepmts, login error messages

bsk:


web/domains/user/views.py
```diff
@@ -33,7 +33,8 @@ def details_update(request, action, pk):
                 messages.error(request,
                                'Please correct the highlighted errors.')
 
-    return render(request, 'web/user/details.html', forms)
+    return render(request, 'web/user/details.html' if request.user.pk == pk
+                  else 'web/user/admin-view-details.html', forms)
```
breneser: Does this mean both web user accounts administration and user details served through same url?

bsk: No, current user details are at /user/ and account admin is at /users/user_id. Do you think it's better to replace this with a URL check or pass a flag through the stack?

PR #42, opened 2020-01-20
Request exporter access

bsk:


web/domains/case/access/forms.py
```diff
@@ -1,10 +1,29 @@
...
 
 
 class AccessRequestForm(ViewFlowModelEditForm):
+    def validate_fields(self, fields):
+        for field in fields:
+            if self.data[field] == '':
+                self.add_error(field, 'You must enter this item')
+
+    def clean(self):
+        request_type = self.data['request_type']
```
rossmiller: cleaned_data = super().clean()
request_type = cleaned_data.get("request_type")

(the individual clean functions will have been called by this point - see https://docs.djangoproject.com/en/3.0/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

web/static/web/js/request-access.js
```diff
@@ -21,19 +22,48 @@ document.addEventListener("DOMContentLoaded", function (event) { // doesn't work
...
+    if (dropdown.value !== '') {
+        // this is a POST request
+        showOrHide(uiConfig, dropdown, reasonRowClasses, agentNameRowClasses, agentAddressRowClasses);
+        doRemoveOptionalLabels();
+        container.scrollIntoView(true);
+    }
 });
 
+function showOrHide(uiConfig, dropdown, reasonRowClasses, agentNameRowClasses, agentAddressRowClasses) {
+    'use strict';
```
rossmiller: Is 'use strict' this supposed to be here?

web/static/web/js/request-access.js
```diff
@@ -21,19 +22,48 @@ document.addEventListener("DOMContentLoaded", function (event) { // doesn't work
...
+
 function getClasses(id) {
     'use strict';
     return document.getElementById(id).parentElement.parentElement.classList;
-}
\ No newline at end of file
+}
+
+function doRemoveOptionalLabels() {
+    'use strict';
```
rossmiller: As above

web/static/web/js/request-access.js
```diff
@@ -21,19 +22,48 @@ document.addEventListener("DOMContentLoaded", function (event) { // doesn't work
...
+    removeOptionalLabels();
+    removeOptionalLabels(); // needs two calls, for some unknown reason, to fully do the job ...
+}
+
+function removeOptionalLabels() {
+    'use strict';
```
rossmiller: As above

web/tests/domains/case/access/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def populate_fields(access_request):
+    access_request.agent_name = 'Agent Smith'
+    access_request.agent_address = '50 VS'
+    access_request.request_reason = 'Import/Export'
+
+
+class AccessRequestViewsTest(TestCase):
+    def testMissingRequestType(self):
```
rossmiller: The name of these functions should be in the format "test_missing_request_type" (it's just the Test class name that uses Pascal case)

web/static/web/js/request-access.js
bsk: yeah, it's a little bit strange, but see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

web/static/web/js/request-access.js
bsk: As above

web/static/web/js/request-access.js
bsk: As above

PR #52, opened 2020-02-17
FIR page

francodacosta:
list current FIR entries for current access request

we should discuss __how to display formatted dates in a consistent way across application ?__

Screenshot_2020-02-17 Importer Exporter Access Request - ICMS


web/templates/web/access-request/access-request-fir-list.html
```diff
@@ -0,0 +1,51 @@
+{% extends "layout/sidebar.html" %}
+
+{% import "forms/forms.html" as forms %}
+{% import "forms/fields.html" as fields %}
```
breneser: You probably don't need these imported in this template?

PR #55, opened 2020-02-19
styling display forms + show edit form in pace + use post actions

francodacosta:
style read only forms to look like normal text
click on a form edit button triggers a post action and shows the edit form in place

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -91,7 +91,11 @@ class AccessRequestFirListView(TemplateView, PostActionMixin):
 
     def create_display_or_edit_form(self, fir, selected_fir):
         """
-        creates an edit form if fir is the selected fir otherwise creates a display form
+        This function either returns an Further Information Request (FIR) form, or a read only version of it.
+
+        By default returns a read only version of the FIR form (for display puposes).
+
+        If `fir.id` is is the same as `selected_fir` then a "editable" version of the form is returned
         """
         if selected_fir and fir.id == selected_fir:
             return FurtherInformationRequestForm(instance=fir)
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -74,10 +74,10 @@ class AccessRequestFirListView(TemplateView, PostActionMixin):
     def edit(self, request, process_id, *args, **kwargs):
         """
         Edits the FIR selected by the user.
-        The selected fir comes from the id property in request body
+        The selected FIR comes from the id property in the request body
         """
 
-        data = request.POST.copy() if request.POST else None
+        data = request.POST if request.POST else None
         if not data or 'id' not in data:
             return HttpResponseBadRequest('Invalid body received')
 
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -2,11 +2,14 @@ from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.generic import TemplateView
 from django.views.generic.edit import FormView
+from django.http import HttpResponseBadRequest
+from django.shortcuts import render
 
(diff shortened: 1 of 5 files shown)


web/domains/case/access/views.py
```diff
@@ -62,26 +65,53 @@ def get_form(self):
...
     template_name = 'web/access-request/access-request-fir-list.html'
 
-    def get_context_data(self, process_id, *args, **kwargs):
+    def edit(self, request, process_id, *args, **kwargs):
+        """
+        Edits the FIR selected by the user.
+        The selected fir comes from the id property in request body
+        """
+
+        data = request.POST.copy() if request.POST else None
```
rossmiller: What's the point of the copy?

web/domains/case/access/views.py
```diff
@@ -62,26 +65,53 @@ def get_form(self):
...
     """
     Access Request Further Information Request - list
     """
     template_name = 'web/access-request/access-request-fir-list.html'
 
-    def get_context_data(self, process_id, *args, **kwargs):
+    def edit(self, request, process_id, *args, **kwargs):
+        """
+        Edits the FIR selected by the user.
+        The selected fir comes from the id property in request body
```
rossmiller: I would change to: "The selected FIR comes from the id property in the request body"

web/domains/case/access/views.py
```diff
@@ -62,26 +65,53 @@ def get_form(self):
...
+
+        return render(
+            request,
+            self.template_name,
+            self.get_context_data(process_id, selected_fir=fir_id, action="edit")
+        )
+
+    def create_display_or_edit_form(self, fir, selected_fir):
+        """
+        creates an edit form if fir is the selected fir otherwise creates a display form
```
rossmiller: I don't quite understand what the comment means.

web/domains/case/access/views.py
francodacosta: none, but it is being done like this on other view, i thought to keep it for consistency

web/domains/case/access/views.py
francodacosta: I've updated its description, hope it is clear now

web/domains/case/access/views.py
```diff
@@ -62,26 +65,57 @@ def get_form(self):
...
+        )
+
+    def create_display_or_edit_form(self, fir, selected_fir):
+        """
+        This function either returns an Further Information Request (FIR) form, or a read only version of it.
+
+        By default returns a read only version of the FIR form (for display puposes).
+
+        If `fir.id` is is the same as `selected_fir` then a "editable" version of the form is returned
+        """
```
rossmiller: "This function either returns a Further Information Request (FIR) form, or a read only version of it.
        By default it returns a read only version of the FIR form (for display purposes).
        If `fir.id` is is the same as `selected_fir` then a "editable" version of the form is returned"

web/templates/web/access-request/partials/fir-item.html
```diff
@@ -3,9 +3,24 @@
 
 
 <fieldset>
-    <legend><strong>FIR {{ idx }}</strong></legend>
+    <legend><strong>FIR {{ idx }} </strong></legend>
+
+
+    {% call forms.form(method='post', csrf_input=csrf_input, container_class=action) -%}
+        {{ fields.hidden('id', fir.instance.id) }}
```
breneser: I would just make one suggestion. With detail views I never really wrap the fields around with a form. Just fields for display purposes. I guess that would break your styling though. Maybe you could use a div around them? 

Otherwise there will be a form that can be submitted, it would still be safe as inputs are disabled and Django won't accept any changes.

Hopefully that makes sense?

PR #56, opened 2020-02-25
Add and Edit FIR

francodacosta:
* add new FIR
* edit existing FIR

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -13,8 +13,8 @@ from web.domains.case.forms import FurtherInformationRequestDisplayForm, Further
 from web.views.mixins import PostActionMixin
 from web.domains.case.models import FurtherInformationRequest
 
-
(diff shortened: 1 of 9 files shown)


web/domains/case/access/views.py
```diff
@@ -86,7 +90,39 @@ def edit(self, request, process_id, *args, **kwargs):
...
+        """
+        Saves the FIR being editted by the user
+        """
+        model = FurtherInformationRequest.objects.get(pk=request.POST['id'])
+        form = FurtherInformationRequestForm(data=request.POST, instance=model)
+
+        if form.is_valid():
+            form.save()
+        else:
+            return HttpResponseBadRequest('invalid form')
```
rossmiller: I'm not sure how this behaves in the UI but would it be useful to return the form errors to the client?

francodacosta: There is a different story for that, which was just pushed to this PR

PR #61, opened 2020-02-26
improvements

francodacosta:
* Renamed `AccessRequestFirListView` to `AccessRequestFirView` as we are doing more than just listing FIRs

*  `AccessRequestFirView` not extending `TemplateView` for consistency so that all view actions use similar logic to display data

* More accurate docblocks

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -132,7 +132,18 @@ class AccessRequestFirView(PostActionMixin):
         Params:
             process_id - Access Request id
         """
-        self.set_fir_status(request.POST['id'], FurtherInformationRequest.CLOSED)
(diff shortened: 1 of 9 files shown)


web/domains/case/access/views.py
```diff
@@ -70,17 +70,35 @@ def get_form(self):
...
+    Access Request Further Information Request
+
+    This view class handles all actions that can be performed on a FIR
     """
     template_name = 'web/access-request/access-request-fir-list.html'
     FIR_TEMPLATE_CODE = 'IAR_RFI_EMAIL'
 
+    def get(self, request, process_id):
+        """
+        Lists all FIRs associated to the acces request
```
rossmiller: access

PR #64, opened 2020-03-02
unit tests

francodacosta:


web/tests/domains/case/access/test_views.py
```diff
@@ -49,3 +54,42 @@ def test_exporter_agent(self):
...
+        form = self.view.create_display_or_edit_form(self.model, self.model.pk, expected_return_class())
+        self.assertIsInstance(form, expected_return_class)
+
+    def test_create_display_or_edit_form_return_display_form(self):
+        """
+        Asserts that when the fir id DO NOT match the id passed a display form is returned
+        """
+        form = self.view.create_display_or_edit_form(self.model, self.model.pk + 1, None)
+        self.assertIsInstance(form, FurtherInformationRequestDisplayForm)
+        self.assertEquals(form.instance, self.model)
```
bsk: A lot of warnings:
`web/tests/domains/case/access/test_views.py::AccessRequestFirViewTest::test_create_display_or_edit_form_return_edit_form
  /root/code/web/tests/domains/case/access/test_views.py:79: DeprecationWarning: Please use assertEqual instead.
    self.assertEquals(form.instance, self.model)`

francodacosta: I've fixed it on my tests, but i think we should have a new story to fix the warnings for existing tests, resolving some of the warnings might not be as simple as deleting a letter

PR #73, opened 2020-03-18
Feature/link importer

breneser:


web/domains/case/access/actions.py
```diff
@@ -0,0 +1,36 @@
...
+
+import structlog as logging
+from django.contrib import messages
+from django.shortcuts import redirect
+from django.urls import reverse
+
+from web.domains.case.access.models import AccessRequestProcess
+from web.views.actions import PostAction
+
+# from viewflow.decorators import flow_func
```
francodacosta: should this be deleted?

web/utils/utils.py
```diff
@@ -80,6 +47,7 @@ class FilevalidationService():
         'wsc',
         'wsf',
         'wsh',
+        'txt'
```
francodacosta: this should be removed, we allow for .txt files

breneser: Oh, guess merge wasn't as right as I though :D I will fix this

web/domains/case/access/actions.py
breneser: the comment? Or all  those imports?

francodacosta: the comment

PR #76, opened 2020-03-24
restore deleted file

francodacosta:
restore deleted files

Screenshot_2020-03-24 Importer Exporter Access Request - ICMS

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -269,6 +269,16 @@ class AccessRequestFirView(PostActionMixin):
         file_model.is_active = False
         file_model.save()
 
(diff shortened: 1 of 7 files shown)


breneser:
Could it be possible to change the `Expected Value` message to something like `Empty file not allowed` ? I think at some point someone needs to give us content for new messages and texts.

francodacosta:
> 
> 
> Could it be possible to change the `Expected Value` message to something like `Empty file not allowed` ? I think at some point someone needs to give us content for new messages and texts.

Are you talking about the ```HttpResponseBadRequest('Invalid body received')``` responses ?

PR #77, opened 2020-03-30
Feature/user management

breneser:


breneser:
Closing temporarily as I am applying a workaround for broken pytest-sugar library

PR #78, opened 2020-03-30
return to edit mode on file upload actions

francodacosta:
file upload actions are done as part of the save action, to ensure any changes to form are saved if a file action is performed

click the save button redirects to list view, performing file actions keeps user in edit view

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -220,11 +220,13 @@ class AccessRequestFirView(PostActionMixin):
 
         return redirect('access_request_fir_list', process_id=process_id)
 
-
     @csrf_exempt
     def save(self, request, process_id, *args, **kwargs):
         """
         Saves the FIR being editted by the user
+
+        user is redirected to list view if no file actions are performed, otherwise user is redirected
+        to edit view
 
         Params:
             process_id - Access Request id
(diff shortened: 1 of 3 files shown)


web/domains/case/access/views.py
```diff
@@ -220,24 +220,49 @@ def delete(self, request, process_id):
...
+            )
+
+        form.save()
+
+        if 'uploaded_file' in request.FILES:
+            self.upload(request, process_id)
+            return self.edit(request, process_id)
+
+        if 'delete_file_id' in request.POST:
+            logger.debug('going to mark file %s as deleted' % request.POST['delete_file_id'])
```
rossmiller: A matter of taste to some degree but I would use f-strings for formatting these days.

web/domains/case/access/views.py
```diff
@@ -220,24 +220,49 @@ def delete(self, request, process_id):
 
         return redirect('access_request_fir_list', process_id=process_id)
 
+
+    @csrf_exempt
```
rossmiller: Is this a security risk in any way?

francodacosta: a bit, but s3 chunk uploader documentation explicitly says to disable csrf

rossmiller: We agreed to add a spike to investigate why they made this recommendation and if it's wholesome.

PR #84, opened 2020-04-01
Add redis and celery for async tasks

breneser:
Sending mailshots will send emails to thousands of users which will be
done asynchronously via Celery


breneser:
Closing the request as I made email template rendering logic through Celery as well.

PR #85, opened 2020-04-02
search templates

francodacosta:
added:
* new template types
* display template view

very basic view, no styling as of yet
image

### DIFF
--- a/web/domains/template/views.py
+++ b/web/domains/template/views.py
@@ -1,8 +1,10 @@
-from web.views import ModelFilterView
+from django.urls import reverse_lazy
+
+from web.views import ModelFilterView, ModelDetailView
 from web.views.actions import Archive, Unarchive, Edit
 
 from .models import Template
-from .forms import TemplatesFilter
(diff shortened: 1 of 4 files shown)


web/domains/template/views.py
```diff
@@ -34,3 +34,11 @@ class Display:
             }
         }
         actions = [Archive(), Unarchive(), Edit()]
+
+
+class TemplateDetailView(ModelDetailView):
+    form_class = GenericTemplate
+    model = Template
+    permission_required = 'web.IMP_ADMIN:MAINTAIN_ALL:IMP_MAINTAIN_ALL'
+    success_url = "javascript:history.go(-1)"
```
breneser, line 43: I don't think you need success url for detail view. If is used only with Django's generic UpdateView which ModelEditView implements. This would still work but might be confusing? If it is in any other detail view it is my bad :)

francodacosta: as agreed this is going to be addressed in a different pr

PR #89, opened 2020-04-06
Feature/async emails

breneser:
We should probably create some fixtures for test cases as you can see I created some random  importer/exporter data with member relations and random emails to test mailshots.


config/settings/base.py
```diff
@@ -203,3 +207,16 @@
     wrapper_class=structlog.stdlib.BoundLogger,
     cache_logger_on_first_use=True,
 )
+
+# Celery & Redis shared configuration
+if 'redis' in VCAP_SERVICES:
```
francodacosta, line 212: would it make sense to default `VACP_SERVICES` to local settings? that way we avoid the this logic

web/templates/email/registration/registration.html
```diff
@@ -2,21 +2,29 @@
...
-        </tr>
-        <tr>
-            <td align="left" valign="top"><b><u>Password:</u></b></td>
-            <td><b>{{password}}</b><br><i>(please note, this is a temporary password and you will be required to change it on first login)</i></td>
-        </tr>
-    </table>
-    <p>Please <a href="https://www.ilb.trade.gov.uk/icms/fox/icms/IMP_LOGIN/login">click here to login</a>.</p>
-    <p/>
-    For assistance or support email: <a href="mailto:enquiries.ilb@trade.gov.uk">enquiries.ilb@trade.gov.uk</a>.
+<p>
```
francodacosta, line 5: should this be on the database, as a normal template?

breneser: That bit is deleted. Now it is using `{{url}}`. If you mean the full email template though it is not in the templates on current system.

config/settings/base.py
breneser: Initially I was thinking of doing that but VCAP_SERVICES comes from Gov PaaS. I copied this logic from another project. I think the thought around this one is that it should work with `REDIS_URL` in env vars even if there is no VCAP_SERVICES ( I guess just in case it is deployed elsewhere, e.g. Heroku, etc. )

PR #98, opened 2020-04-17
create endorsement + change edit + detail endorsement views to hide t…

francodacosta:
create endorsement
edit and  detail endorsement views to hide template title field
hidden fields do not reder the label


web/domains/template/forms.py
```diff
@@ -50,6 +51,27 @@ class Meta:
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['template_type'].widget = forms.HiddenInput()
+        self.fields['template_type'].choices = (Template.get_choice_entry(Template.TYPES, Template.ENDORSEMENT),)
+        self.fields['template_type'].initial = Template.ENDORSEMENT
+
+        self.fields['application_domain'].widget = forms.HiddenInput()
+        self.fields['application_domain'].choices = (Template.get_choice_entry(Template.DOMAINS, Template.IMPORT_APPLICATION),)
+        self.fields['application_domain'].initial = Template.IMPORT_APPLICATION
```
breneser, line 64: Is it possible to disable the field as well? If it was changed in the markup I think it might be allowing the change on the hidden field.

breneser: Or maybe we could save set these on save if we can't disable? I mean rather than as hidden inputs?

web/domains/template/models.py
```diff
@@ -48,14 +48,14 @@ class Template(Archivable, models.Model):
...
-    start_datetime = models.DateTimeField(blank=False, null=False)
+    start_datetime = models.DateTimeField(auto_now_add=True, blank=False, null=False)
     end_datetime = models.DateTimeField(blank=True, null=True)
     is_active = models.BooleanField(blank=False, null=False, default=True)
     template_name = models.CharField(max_length=100, blank=False, null=False)
     template_code = models.CharField(max_length=50, blank=True, null=True)
     template_type = models.CharField(max_length=50, choices=TYPES, blank=False, null=False)
     application_domain = models.CharField(max_length=20, choices=DOMAINS, blank=False, null=False)
-    template_title = models.CharField(max_length=4000, blank=False, null=True)
+    template_title = models.CharField(max_length=4000, blank=True, null=True)
```
breneser, line 58: I think this might have an effect on other templates' creation as it wouldn't validate without blank=True ? Not 100% sure though.

web/domains/template/forms.py
```diff
@@ -50,6 +51,27 @@ class Meta:
         fields = ['template_name', 'template_title', 'template_content']
 
 
+class EnsorsemenCreateTemplateForm(GenericTemplate):
```
breneser, line 54: Typo here ? :) It should be `EndorsementCreateTemplateForm`

francodacosta: :+1

web/domains/template/forms.py
francodacosta: I will try to assign the value on save.
I think they need to be hidden inputs though

francodacosta: I can disable it, but if we are talking about changing the html markup, it won't prevetn any tampering.
I think the safest bet is to change the value on save, that will effectively ignore any user tampering on that field

francodacosta: @breneser From what i could understand, since we define only one choice, any oher value will fail form validation

francodacosta: Thanks Bren, I've put the readonly fields ont he view form_valid, that seems to work fine

web/domains/template/views.py
```diff
@@ -52,3 +52,24 @@ class TemplateEditView(ModelUpdateView):
     success_url = reverse_lazy('template-list')
     cancel_url = success_url
     permission_required = 'web.IMP_ADMIN:MAINTAIN_ALL:IMP_MAINTAIN_ALL'
+
+
+class EnsorsemenCreateView(ModelCreateView):
```
breneser, line 57: Sorry, missed this. Same typo here

PR #103, opened 2020-04-21
use js to submit action buttons

francodacosta:
* Action buttons no longer need to have hardcoded form and js sumission code, that can be replaced with ```<a role="action-button" data-csrf="{{ csrf_token }}" href="..."/>```
* small change to docker to use named volumes for postgres as a bug with postgres prevents the use of mapped volumes in windows hosts
* replaces call of FOSjs.action with new style action buttons on /workbasked


web/static/web/js/action-buttons.js
```diff
@@ -0,0 +1,41 @@
...
+        evt.preventDefault();
+        var url    = $(this).attr('href');
+        var action = $(this).attr('data-action');
+        var csrf   = $(this).attr('data-csrf');
+        var method = $(this).attr('data-method') || 'POST';
+
+        var confirmMessage =  $(this).attr('data-confirm');
+
+        var doAction = function(){
+            console.log('running action button callback')
```
breneser, line 13: should we delete this log?

PR #115, opened 2020-05-06
Feature/access request flow

breneser:


web/jinja2.py
```diff
@@ -26,6 +31,44 @@ def show_all_attrs(value):
     return '\n'.join(res)
 
 
+def capitalize(value):
```
francodacosta, line 34: should this be handled by css `text-transform` property?

breneser: Absolutely right. Let me do that

breneser: Pushed an update

PR #125, opened 2020-05-13
base circle ci config file

francodacosta:


features/steps/login.py
```diff
@@ -23,10 +23,15 @@ def user_login(context, username, password):
 
 @then(u'the user is presented with an invalid login message')
 def login_error_is_displayed(context):
-    assert context.browser.find_element(By.ID, 'login-error'), "Login Error Message Not Found"
-    assert 'Invalid username or password' in context.browser.find_element(By.ID, 'login-error').text, "Login Error Message Is not the one we expected"
+    print('checking for login error')
```
breneser, line 26: Could you remove print ? Or use logger.debug ?

Makefile
```diff
@@ -105,35 +105,38 @@ down: ## Stops and downs containers
...
-	unset UID && \
-	DJANGO_SETTINGS_MODULE=config.settings.test \
-	docker-compose run --rm web pytest --flake8 -v
-
 behave: down
-	docker-compose run web sh -c " \
-		echo 'drop  database test_postgres; create database test_postgres;' | python manage.py dbshell && \
-		python manage.py migrate \
-	"
+	docker-compose run web sh -c "echo 'drop  database test_postgres; create database test_postgres;' | python manage.py dbshell"
```
breneser, line 126: If there are two concurrent builds would it cause any issues? Could it be possible to attach some build id to the database?

breneser: CIRCLE_CI_BRANCH env var might also work

Makefile
```diff
@@ -105,35 +105,38 @@ down: ## Stops and downs containers
...
+	ICMS_DEBUG=False \
+	docker-compose run --rm web python -m flake8 -v
+
+test: ## run tests
+	mkdir -p test-reports
 	unset UID && \
 	ICMS_DEBUG=False \
 	TEST_TARGET='web/tests' \
 	DJANGO_SETTINGS_MODULE=config.settings.test \
-	docker-compose run --rm web pytest --verbose --cov=web --cov=config $(TEST_TARGET)
```
breneser, line 113: TEST_TARGET variable really helps with testing locally. I set TEST_TARGET when I am working on a specific domain's tests to only run them.

Makefile
francodacosta: if there are 2 concurrents build you will get an error message when trying to drop the db
postgress will refuse to drop it if there are active connections to the db

Makefile
francodacosta: ok, I will add it and default to `web/tests`

Makefile
francodacosta: the `CIRCLE_CI` will fail locally as it's not present.

Since this runs on docker, and circle ci starts a new docker machine per build, even if other build is running it will use its own containers and not interfere with this one

PR #141, opened 2020-06-09
filter by importer status

francodacosta:


web/domains/importer/models.py
```diff
@@ -23,11 +23,26 @@ class Importer(Archivable, BaseTeam):
     NON_EUROPEAN = "O"
     REGIONS = ((UK, "UK"), (EUROPE, 'Europe'), (NON_EUROPEAN, 'Non-European'))
 
+    # Statuses
```
breneser, line 26: `is_active` field on importers holds the status information as active/archived. Migrated data does not have draft status anymore, and there isn't a way to create a DRAFT importer, we should be using is_active to create the filter

PR #143, opened 2020-06-11
show importer offices

francodacosta:
* view import action on importer list page
* display importer page
* show current importer offices
* show archived importer offices

### DIFF
--- a/web/templates/forms/fields.html
+++ b/web/templates/forms/fields.html
@@ -91,9 +91,18 @@
 {%- endmacro %}
 
 {% macro field_as_text(field) -%}
-  {% if field.field.widget.__class__.__name__ in ['Select', 'TextInput', 'EmailInput', 'Textarea', 'DateInput', 'NumberInput'] %}
-    {{ field.value()|safe|nl2br }}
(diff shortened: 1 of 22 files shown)


web/templates/forms/fields.html
```diff
@@ -91,7 +91,7 @@
 {%- endmacro %}
 
 {% macro field_as_text(field) -%}
-  {% if field.field.widget.__class__.__name__ in ['TextInput', 'EmailInput', 'Textarea', 'DateInput', 'NumberInput'] %}
+  {% if field.field.widget.__class__.__name__ in ['Select', 'TextInput', 'EmailInput', 'Textarea', 'DateInput', 'NumberInput'] %}
```
breneser, line 94: This affects other view pages as well. Constabulary is one example. Select will display the option value rather than label. Currently it displays a disabled select box, which is how current icms does it.

francodacosta: That's a good point, I didn't notice that, I will try to find a way to display the label if not I will rever to the diables select box (and style it to look like text if possible)

PR #159, opened 2020-07-27
Feature/icmslst 397 fir with viewflow

breneser:


web/notify/utils.py
```diff
@@ -1,30 +1,35 @@
...
-):
+def send_email(subject, message, recipients, html_message=None, cc_list=None):
     """
         Sends emails to given recipients.
+
+        cc_list: ";" separated email list
     """
-    send_mail(subject, message, settings.EMAIL_FROM, recipients, html_message=html_message)
+    if cc_list:
+        cc_list = cc_list.split(";").join(",")
```
kevinetienne, line 18: It looks like it could be me but should this line be:
```python
        cc_list = ",",join(cc_list.split(";"))
```

web/domains/template/models.py
```diff
@@ -96,11 +96,34 @@ def get_content(self, replacements={}):
...
+        title = self.template_title
+
+        if title is None:
+            return ""
+
+        if replacements is None:
+            return title
+
+        for replacement, value in replacements.items():
+            title = re.sub(r"\[\[{}\]\]".format(replacement), str(value), title)
```
kevinetienne, line 123: As the regex is used for the title and the content is it possible to create a constant/variable on top of the file please?

web/domains/case/access/models.py
```diff
@@ -114,24 +141,78 @@ def save(self, *args, **kwargs):
...
+
     access_request = models.ForeignKey(AccessRequest, null=True, on_delete=models.SET_NULL)
     approval_required = models.BooleanField(blank=False, null=False, default=False)
 
     restart_approval = models.BooleanField(blank=False, null=False, default=False)
     re_link = models.BooleanField(blank=False, null=False, default=False)
+    fir_processes = GenericRelation(FurtherInformationRequestProcess)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
kevinetienne, line 162: You might not need to override the `__init__`.

web/domains/case/access/models.py
```diff
@@ -114,24 +141,78 @@ def save(self, *args, **kwargs):
...
+
+    def fir_content(self, request):
+        template = _get_fir_template()
```
kevinetienne, line 165: I've seen this pattern used a few times in this PR. Is there a principle or a rule guiding when to use a class variable vs a method. I guess this method is used at least two times so it could be a matter of having a mixin vs defining a function.

web/views/mixins.py
```diff
@@ -1,6 +1,18 @@
...
+from django.views.generic.edit import FormView
 from django.views.generic.list import ListView
+from s3chunkuploader.file_handler import s3_client
+
+from web.domains.file.models import File
+from web.utils import FilevalidationService
+from web.utils.s3upload import InvalidFileException, S3UploadService
+from web.utils.virus import ClamAV, InfectedFileException
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
breneser: Fixing this

web/notify/utils.py
breneser: Good catch, fixing this

web/domains/template/models.py
breneser: changing this accordingly

web/domains/case/access/models.py
breneser: Yea, forgot to delete. Deleting this

web/views/mixins.py
```diff
@@ -53,3 +65,127 @@ def post(self, request, *args, **kwargs):
...
+            return self.file_queryset
+
+        return File.objects.none()
+
+    def get_files(self):
+        request = self.request
+        if request.POST:
+            # use posted file list with POST requests
+            file_ids = request.POST.getlist("files")
+            files = list(File.objects.filter(pk__in=file_ids).all())
```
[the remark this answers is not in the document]
breneser: The thing is users can add new files. So files might not even be related to resource beforehand.

It will reuse the ids if there are existing files though. The pages are restricted to users with certain permissions. It would be those users tempering with it, and it kind of is the purpose of the file upload component to add/delete/restore files.

web/domains/case/access/models.py
breneser: There is no principle really. Import and Export access requests use the same template for Further Information Requests (FIR) , it is used twice, that's why I added it there.

Mind you I am not very experienced with Python, if this pattern would cause an issue I can change it to a more appropriate one.

web/domains/case/access/models.py
```diff
@@ -85,6 +111,7 @@ class AccessRequest(models.Model):
 
     @property
     def request_type_verbose(self):
+        logger.debug(f"request type is: {self.request_type}")
```
oskusalerma, line 114: probably want to take that out before merging? lots of these around as well in your changes

web/domains/case/access/models.py
oskusalerma: i would say this is fine. if something can be a stand alone function, it probably should be.

web/domains/case/access/models.py
```diff
@@ -114,24 +141,75 @@ def save(self, *args, **kwargs):
...
+    FIR_TEMPLATE_CODE = "IAR_RFI_EMAIL"
+
     access_request = models.ForeignKey(AccessRequest, null=True, on_delete=models.SET_NULL)
     approval_required = models.BooleanField(blank=False, null=False, default=False)
 
     restart_approval = models.BooleanField(blank=False, null=False, default=False)
     re_link = models.BooleanField(blank=False, null=False, default=False)
+    fir_processes = GenericRelation(FurtherInformationRequestProcess)
+
+    def __init__(self, *args, **kwargs):
```
oskusalerma, line 196: is there a point to this? it just forwards to baseclass, which is the default behavior

web/domains/case/access/urls.py
```diff
@@ -16,10 +16,15 @@
 exporter_access_request_urls = FlowViewSet(ExporterAccessRequestFlow).urls
 approval_request_urls = FlowViewSet(ApprovalRequestFlow).urls
 
+logger.debug(importer_access_request_urls)
```
oskusalerma, line 19: remove?

web/views/mixins.py
```diff
@@ -53,3 +62,127 @@ def post(self, request, *args, **kwargs):
...
+            )
+
+            file_path = upload_service.process_uploaded_file(
+                settings.AWS_STORAGE_BUCKET_NAME, uploaded_file
+            )
+
+        except (InvalidFileException, InfectedFileException) as e:
+            error_message = str(e)
+        except Exception as e:
+            logger.exception(e)
```
oskusalerma, line 125: we really need to get sentry integrated to icms

web/utils/s3upload.py
```diff
@@ -82,23 +79,21 @@ def process_uploaded_file(self, bucket, uploaded_file, destination):
...
-            self.s3_client.copy_object(
-                Bucket=bucket,
-                CopySource=url_path_join(bucket, uploaded_file.name),
-                Key=new_file_path,
-            )
-
-            return new_file_path
+            return url_path_join(bucket, uploaded_file.name)
 
+            # move file to final destination
```
oskusalerma, line 84: why are you leaving commented out code in?

web/domains/case/access/models.py
```diff
@@ -1,15 +1,41 @@
...
+from django.utils import timezone
 from viewflow.models import Process
 
-from web.domains.case.fir.models import FurtherInformationRequest
 from web.domains.exporter.models import Exporter
 from web.domains.importer.models import Importer
+from web.domains.template.models import Template
 from web.domains.user.models import User
 
+from ..fir.mixins import FurtherInformationProcessMixin
```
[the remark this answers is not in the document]
breneser: changing this accordingly

web/domains/case/access/models.py
```diff
@@ -1,15 +1,41 @@
...
+from ..fir.models import FurtherInformationRequest, FurtherInformationRequestProcess
 from .managers import AccessRequestQuerySet
 
+logger = logging.getLogger(__name__)
+
+
+def _get_fir_template():
+    """
```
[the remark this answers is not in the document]
breneser: making this change

web/domains/case/access/models.py
breneser: on it

web/domains/case/access/models.py
breneser: With generic foreign keys you need a `GenericRelation` to enable querying from parent process. Otherwise it doesn't seem to resolve `importeraccessrequestprocess.fir_processes` as it is Generic Foreign Key.

I actually wanted to add that to `FurtherInformationProcessMixin` , but for that to work I need to make the mixin extend from `Process` and mark it as abstract.

But unfortunately that's where I hit a Viewflow limitation. Email the author and he says that would be too complex. Even though Django is happy with abstract models extending a concrete one ( `Process` in this case)

web/views/mixins.py
breneser: Yes, that's a long term plan as fat as I understand.

web/utils/s3upload.py
breneser: ah, that's the file upload that wasn't working. It shouldn't be moving it to its place until the model is saved. I can add a TODO marker there?

PR #166, opened 2020-08-11
Feature/icmslst 316 create importer

kevinetienne:
This change request improve the form to create importer for an ICMS admin. There is now a sidebar to create either an individual or an organisational importer. Both entities have different fields but share the same models.

An individual importer is related to a person and can have an EORI number and many offices.
An organisation importer is related to a company name, it has an EORI number and can have many offices.

Create an individual importer:
importer_individual
Associate a Person with an individual importer:
importer_individual_find_person

Create an organisational importer:
importer_org
Associate a company from companies house to an organisational importer:
importer_org_find_company


config/settings/base.py
```diff
@@ -223,3 +226,6 @@
 
 # Default domain is used in email templates to pint users to ICMS from emails
 DEFAULT_DOMAIN = env.str("ICMS_DEFAULT_DOMAIN", "http://localhost:8080/")
+
+COMPANIES_HOUSE_TOKEN = os.environ.get("COMPANIES_HOUSE_TOKEN", "changeme")
+COMPANIES_HOUSE_TOKEN = base64.b64encode(bytes(COMPANIES_HOUSE_TOKEN, "utf-8")).decode("utf-8")
```
oskusalerma, line 231: any particular reason not to just have the environment variable already set to the base64-encoded value? would eliminate the need for the second line

requirements-base.txt
```diff
@@ -1,4 +1,5 @@
 --extra-index-url https://pypi.viewflow.io/${ICMS_VIEWFLOW_LICENSE}/simple/
+django-autocomplete-light~=3.5.1
```
oskusalerma, line 2: as discussed offline, lets have a discussion about dependency pinning strategies

web/static/3rdparty/select2/select2.min.css
```diff
@@ -0,0 +1 @@
...ius:0}.select2-container--classic.select2-container--open.select2-container--below .select2-selection--multiple{border-bottom:none;border-bottom-left-radius:0;border-bottom-right-radius:0}.select2-container--classic .select2-search--dropdown .select2-search__field{border:1px solid #aaa;outline:0}.select2-container--classic .select2-search--inline .select2-search__field{outline:0;box-shadow:none}.select2-container--classic .select2-dropdown{background-color:#fff;border:1px solid transparent}.select2-container--classic .select2-dropdown--above{border-bottom:none}.select2-container--classic .select2-dropdown--below{border-top:none}.select2-container--classic .select2-results>.select2-results__options{max-height:200px;overflow-y:auto}.select2-container--classic .select2-results__option[role=group]{padding:0}.select2-container--classic .select2-results__option[aria-disabled=true]{color:grey}.select2-container--classic .select2-results__option--highlighted[aria-selected]{background-color:#3875d7;color:#fff}.select2-container--classic .select2-results__group{cursor:default;display:block;padding:6px}.select2-container--classic.select2-container--open .select2-dropdown{border-color:#5897fb}
```
oskusalerma, line 1: if you add stuff to web/static/3rdparty, you have to modify NPM_FILE_PATTERNS in development.py so the files are copied correctly. see https://github.com/uktrade/icms/commit/b94dbce1c97033a86f6653c2d4c187c98296f4b3 for an example (which, btw, also adds select2).

web/templates/base.html
```diff
@@ -22,6 +22,10 @@
         <link rel="stylesheet" type="text/css" href="{{ static('web/css/fox.css') }}">
         <link rel="stylesheet" type="text/css" href="{{ static('web/css/fox_default_layout.css') }}">
         <link rel="stylesheet" type="text/css" href="{{ static('3rdparty/jodit/jodit.min.css') }}">
+
+        <link href="/static/vendor/select2/dist/css/select2.css" type="text/css" media="screen" rel="stylesheet">
+        <link href="/static/admin/css/autocomplete.css" type="text/css" media="screen" rel="stylesheet">
+        <link href="/static/autocomplete_light/select2.css" type="text/css" media="screen" rel="stylesheet">
```
oskusalerma, line 28: do these need to go into base.html? could they just go into the css block on pages that need them? see https://github.com/uktrade/icms/commit/85e6e2845e8fb6e0f26fc1ce40b4e53c71d9472c#diff-1f4f3b951ee5a330db42e2854fb17ad4 for an example.

also why is it hardcoding /static instead of using the django static function?

web/templates/base.html
```diff
@@ -72,6 +76,14 @@
             <script src="{{ static('web/js/components/formset-component.js') }}" type="text/javascript"></script>
             <script src="{{ static('web/js/components/django-formset.js') }}" type="text/javascript"></script>
             <script src="{{ static('web/js/main.js') }}" type="text/javascript"></script>
+
+            <script type="text/javascript" src="/static/autocomplete_light/jquery.init.js"></script>
```
oskusalerma, line 80: do these need to go into base.html? could they just go into the page_js block on pages that need them? see https://github.com/uktrade/icms/commit/85e6e2845e8fb6e0f26fc1ce40b4e53c71d9472c#diff-1f4f3b951ee5a330db42e2854fb17ad4 for an example.

also why is it hardcoding /static instead of using the django static function?

web/static/3rdparty/select2/select2.min.css
kevinetienne: thanks, these two files (`select2.min.css` and `select2.min.js`) shouldn't have been there.
Regarding https://github.com/uktrade/icms/commit/b94dbce1c97033a86f6653c2d4c187c98296f4b3 I suggest we evaluate both libraries and pick one.

web/templates/base.html
kevinetienne: Thanks, yes we don't need them in `base.html` and the relevant css and js files for the autocomplete field should be available on the form.

web/templates/base.html
kevinetienne: see https://github.com/uktrade/icms/pull/166#discussion_r469935591

web/static/3rdparty/select2/select2.min.css
kevinetienne: fixed in https://github.com/uktrade/icms/pull/166/commits/7ca510f5de6dbe6842f5394a064f902c3bddd749

web/templates/base.html
kevinetienne: fixed in https://github.com/uktrade/icms/pull/166/commits/c8d401112190feda932cb3d6b9cd96e776b4e344

web/templates/base.html
kevinetienne: fixed in https://github.com/uktrade/icms/pull/166/commits/c8d401112190feda932cb3d6b9cd96e776b4e344

config/settings/base.py
kevinetienne: I was torn about it, I'll go with your suggestion.

requirements-base.txt
kevinetienne: fixed in https://github.com/uktrade/icms/pull/166/commits/c8101a53f7bc03685b616569aa482278861ac349

config/settings/base.py
kevinetienne: fixed it https://github.com/uktrade/icms/pull/166/commits/2cecbd0a4e4ab7e225752d99f399245eeb3acbb8

PR #168, opened 2020-08-19
Add eori number for Northern Ireland

kevinetienne:
And add verbose names in preparation of adding importer with view forms


web/migrations/0010_auto_20200819_1352.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 2.2.13 on 2020-08-19 13:52
```
kevinetienne, line 1: How do you feel about giving more explicit name to migrations? I think our goal is to squash those migrations before going live but if you have any opinion let me know.

web/domains/importer/models.py
```diff
@@ -30,16 +30,30 @@ class Importer(Archivable, BaseTeam):
...
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        related_name="own_importers",
+        verbose_name="Person",
     )
 
     # set for both
-    comments = models.CharField(max_length=4000, blank=True, null=True)
+    comments = models.TextField(blank=True, null=True)
```
kevinetienne, line 56: Not sure why we didn't use TextField for a comment field, let me know if I'm breaking anything. Using `TextField` should allow forms to display an html `textarea` by default and avoid adding a form widget to render it.

web/domains/importer/models.py
```diff
@@ -30,16 +30,30 @@ class Importer(Archivable, BaseTeam):
     # these are only set for organisations
     name = models.CharField(max_length=4000, blank=True, null=True)
     registered_number = models.CharField(max_length=15, blank=True, null=True)
-    eori_number = models.CharField(max_length=20, blank=True, null=True)
+    eori_number = models.CharField(
+        max_length=20,
+        blank=True,
+        null=True,
+        verbose_name="EORI Number",
+        help_text="EORI number should include the GB prefix",
```
kevinetienne, line 38: `verbose_name` and `help_text` are used by django forms. Using them means I don't need to keep overriding help_text and label for forms. The drawbacks is you need a migration when you alter them.

web/migrations/0010_auto_20200819_1352.py
rossmiller: If you think it will be helpful, do it. Generally, I don't think it matters if you're going to squash them.

rossmiller (approved):
Seems legit!

PR #174, opened 2020-08-25
Feature/icmslst 316 importer update office form

kevinetienne:
We are using formset to add offices to importer. This PR uses:
 - django management form helper instead of hard coding the values
 - remove `show_offices_form` by using `extra=0`


web/domains/importer/views.py
```diff
@@ -74,26 +74,15 @@ def get(self, request, pk, offices_form=None, form=None):
         # should the offices formset be shown on the edit page
```
rossmiller, line 74: I don't understand this comment, might be worth updating whilst you are here.

web/templates/partial/offices/formset-script.html
```diff
@@ -0,0 +1,101 @@
+<script id="office-postcode-lookup-ui" type="text/html">
+  <div class="modal-popover-container">
+    <div class="modal-popover small-popover modal-alert-confirm">
+        <div class="modal-popover-content" role="alertdialog">
+        <div class="modal-popover-icon icon-question"></div>
+        <div class="modal-popover-text">
+          <p>Select Address:</p>
```
rossmiller, line 7: Not sure how much attention you need to pay to accessibility but this could do with a more formal structure - label, fieldset legend etc. https://www.w3.org/WAI/tutorials/forms/

web/templates/partial/offices/formset-script.html
```diff
@@ -0,0 +1,101 @@
...
+                <button class="link-button abort" >Cancel</button>
+            </li>
+            </ul>
+        </div>
+    </div>
+    </div>
+</script>
+
+<script>
+  function updateidx() {
```
rossmiller, line 26: Do you have a naming convention you need to follow with the JS? updateIDx or updateIdx perhaps. 

Does the JS need a corresponding BDD test?

rossmiller (changes requested):
Looks very good! Some minor changes required.

web/templates/partial/offices/formset-script.html
kevinetienne: Thanks, yeah, I'm not sure there was a convention. I guess going forward we should have a generic one. Ideally the jquery code would disappear in the near future but let's try to maintain it as best as we can.

Regarding BDD it looks like there are already [two tests to display and add offices](https://github.com/uktrade/icms/blob/develop/features/importer/importers-edit.feature#L30-L61).

web/templates/partial/offices/formset-script.html
kevinetienne: Code was moved to a new template and I totally agree we need to respect semantics.

On another note those custom elements might be removed and ideally the project would be using generic components  from https://design-system.service.gov.uk/components/select/ (once discussion would be made with skate holders).

web/domains/importer/views.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/174/commits/32023ab02fd844edde53113b368c4cd66d77db30

kevinetienne: This is still a work in progress, I'd like to remove the base classes and use `form_valid`, `form_invalid`, `get_context_data` and `post` instead of `get`, `edit`.

web/templates/partial/offices/formset-script.html
kevinetienne: fixed in https://github.com/uktrade/icms/pull/174/commits/30a8511fa717ef2f03f8c372b473fa865f307bf4

web/templates/partial/offices/formset-script.html
kevinetienne: fixed in https://github.com/uktrade/icms/pull/174/commits/d683624baab866757435efb87f23a600cb93d32d

rossmiller (approved):
LGTM!

PR #178, opened 2020-08-28
Update django from 2.2.13 to 3.1

kevinetienne:
Minimal required updates:
 - `s3chunkuploader` (see https://github.com/uktrade/s3chunkuploader)
 - migration for `django` first name (see https://docs.djangoproject.com/en/3.1/releases/3.1/#abstractuser-first-name-max-length-increased-to-150)

Other updates include version support for django3.1 and/or bump to latest bugfix/version


oskusalerma (approved):
approved, but don't merge until you're back, and please collapse down to one commit when merging

PR #180, opened 2020-09-08
ICMSLST-414 Add gevent and psycogreen to gunicorn

kevinetienne:
Gevent allows to use the same thread for concurrent requests

set worker_class to gevent
set worker_connections to 1000
monkey patch psycopg2


rossmiller (approved):
LGTM!

PR #187, opened 2020-09-17
Feature/icmslst 408 Export case officer can take ownership of an export application

kevinetienne:
Add ability for an export case officer to take ownership of an export case.


web/domains/case/export/models.py
```diff
@@ -138,6 +138,13 @@ class ExportApplication(WorkbasketBase, Process):
         null=True,
         related_name="agent_office_export_applications",
     )
+    case_owner = models.ForeignKey(
+        User,
+        on_delete=models.PROTECT,
+        blank=True,
+        null=True,
+        related_name="owned_export_applications",
```
oskusalerma: discussed offline

oskusalerma (approved):

web/domains/case/export/models.py
kevinetienne: Good suggestion, we don't need the reverse relationship fixed in https://github.com/uktrade/icms/pull/187/commits/0d433235e8eae70877821481332413ceb668fb2d

PR #188, opened 2020-09-22
Export case officer can close case

kevinetienne:


web/domains/case/export/views.py
```diff
@@ -199,7 +204,39 @@ def release_ownership(request, pk):
     return redirect(reverse("workbasket"))
 
 
-class Management(PermissionRequiredMixin, DetailView):
-    model = ExportApplication
-    permission_required = export_case_officer_permission
-    template_name = "web/domains/case/export/management.html"
+@login_required
+@permission_required(permissions, raise_exception=True)
```
oskusalerma, line 208: shouldn't this be export_case_officer_permission?

oskusalerma (changes requested):

web/domains/case/export/views.py
kevinetienne: Fixed in https://github.com/uktrade/icms/pull/188/commits/383aff7e8b86b7e97b641b2c373ed56f33fe4f15

oskusalerma (approved):

PR #190, opened 2020-09-28
ICMSLST-434 Add notes to case

kevinetienne:
As an export case officer add ability to add note to a case


web/domains/case/export/views.py
```diff
@@ -240,3 +247,128 @@ def management(request, pk):
     return render(
         request=request, template_name="web/domains/case/export/management.html", context=context
     )
+
+
+@login_required
+@permission_required(permissions, raise_exception=True)
+def management_notes(request, pk):
+    with transaction.atomic():
+        application = get_object_or_404(klass=ExportApplication.objects.select_for_update(), pk=pk)
```
oskusalerma, line 256: we've agreed before not to name the first param:

`        application = get_object_or_404(ExportApplication.objects.select_for_update(), pk=pk)
`

web/domains/case/export/views.py
```diff
@@ -240,3 +247,128 @@ def management(request, pk):
     return render(
         request=request, template_name="web/domains/case/export/management.html", context=context
     )
+
+
+@login_required
+@permission_required(permissions, raise_exception=True)
```
oskusalerma, line 253: are we sure who is supposed to see notes? i don't know if notes are for case officers to pass along between themselves, in which case this permission is wrong, or whether the people allowed to see applications for the exporter are supposed to see the notes as well.

in the latter case, this should have a TODO comment about using django-guardian, since you can't have people seeing notes for exporter A who only have access to exporter B

web/domains/case/export/views.py
```diff
@@ -240,3 +247,128 @@ def management(request, pk):
...
+
+
+@login_required
+@permission_required(export_case_officer_permission, raise_exception=True)
+@require_POST
+def unarchive_note(request, pk, note_pk):
+    with transaction.atomic():
+        application = get_object_or_404(ExportApplication.objects.select_for_update(), pk=pk)
+        application.get_task(ExportApplication.SUBMITTED, "process")
+        application.case_notes.filter(pk=note_pk).update(status=True)
```
oskusalerma, line 300: should this be `is_active=True`?

web/domains/case/export/views.py
```diff
@@ -240,3 +247,128 @@ def management(request, pk):
...
+                settings.CLAM_AV_USERNAME, settings.CLAM_AV_PASSWORD, settings.CLAM_AV_URL
+            ),
+            file_validator=FilevalidationService(),
+        )
+
+        file_path = upload_service.process_uploaded_file(settings.AWS_STORAGE_BUCKET_NAME, f)
+    except (InvalidFileException, InfectedFileException) as e:
+        error_message = str(e)
+    except Exception as e:
+        logger.exception(e)
```
oskusalerma, line 321: this shuold use sentry to log the exception

oskusalerma (changes requested):

web/domains/case/export/views.py
kevinetienne: Thanks good spot and yes notes are internal, only case officer should see them. Further Information Request and Update Request are the one shared with the applicants.

web/domains/case/export/views.py
kevinetienne: Correct, thanks

web/domains/case/export/views.py
kevinetienne: Updated in 691dcc2f0288d8d427661292af502a6a030dfa61

web/domains/case/export/views.py
kevinetienne: updated in 691dcc2f0288d8d427661292af502a6a030dfa61

web/domains/case/export/views.py
kevinetienne: Updated in 691dcc2f0288d8d427661292af502a6a030dfa61

web/domains/case/export/views.py
kevinetienne: Updated in 691dcc2f0288d8d427661292af502a6a030dfa61

oskusalerma (approved):

PR #193, opened 2020-10-06
ICMSLST-424 Add new permission scheme, use it for export applications.

oskusalerma:
In more detail:

* Add web/permissions.py for global permissions.
* Add django-guardian for object-level permissions.
* Add `make pip-check` to check dependencies for possible updates.
* Add `make psql` to easily browse the database locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,9 @@ mypy: ## run mypy
 	ICMS_DEBUG=False \
 	docker-compose run --rm web mypy web config
 
+pip-check:
+	docker-compose run --rm web pip-check
+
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -98,6 +101,10 @@ shell: ## Starts the Python interactive interpreter
 	ICMS_MIGRATE=False \
 	docker-compose run --rm web python ./manage.py shell
 
+psql: ## Starts psql
+	@unset UID && \
```
kevinetienne, line 105: did you find out why we need to unset UID? Or is it copied from other command?

kevinetienne (approved):
lgtm

Makefile
oskusalerma: just copied from other commands, don't really understand why it's needed (or not)

PR #261, opened 2021-02-03
ICMSLST-492 Allow Importer to create/view/retract withdrawal

kevinetienne:
Withdraw application for Firearms and Ammunition (OIL)


web/domains/case/_import/views.py
```diff
@@ -467,7 +468,10 @@ def case_oil_view(request, pk):
             OpenIndividualLicenceApplication.objects.select_for_update(), pk=pk
         )
 
-        task = application.get_task(ImportApplication.SUBMITTED, "process")
+        try:
+            task = application.get_task(ImportApplication.SUBMITTED, "process")
+        except Exception:
```
oskusalerma, line 473: This is too generic, can you make get_task raise specific exceptions for the different errors it can encounter and then catch the one(s) you want here? you can add the new exception classes in `web/flow/errors.py` or something

web/domains/case/_import/views.py
```diff
@@ -481,6 +485,84 @@ def case_oil_view(request, pk):
...
+@permission_required("web.importer_access", raise_exception=True)
+def case_oil_withdraw(request, pk):
+    with transaction.atomic():
+        application = get_object_or_404(
+            OpenIndividualLicenceApplication.objects.select_for_update(), pk=pk
+        )
+
+        try:
+            task = application.get_task(ImportApplication.SUBMITTED, "process")
+        except Exception:
```
oskusalerma, line 498: same here

oskusalerma (changes requested):

trbateman (approved):

trbateman (changes requested):
Change

web/domains/case/_import/views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/261/commits/97d79ef62e93a09fc832f8251c4587b0143bda9d, we decided to allow get_task to accept one or many expected states.

web/domains/case/_import/views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/261/commits/97d79ef62e93a09fc832f8251c4587b0143bda9d, we decided to allow get_task to accept one or many expected states.

oskusalerma (approved):

PR #263, opened 2021-02-05
ICMSLST-494 Take/release ownership to manage case by ILB Admin

kevinetienne:
Take and release ownership of a case for Firearms and Ammunition (OIL).

I've copied here the same approach used for export certificate.
An ILB Admin 'takes ownership' of an import application by clicking 'Take Ownership' from the workbasket and an ILB Admin 'releases ownership' of an import application by clicking the 'Release Ownership' action from within the case management interface.

Being the scene an ILB Admin is case owner when `ImportApplication.case_owner == ILBAdmin`.


web/domains/case/_import/views.py
```diff
@@ -613,3 +613,51 @@ def view_verified_firearms(request, application_pk, firearms_pk):
...
+        application.get_task(ImportApplication.SUBMITTED, "process")
+        application.case_owner = None
+        application.save()
+
+        return redirect(reverse("workbasket"))
+
+
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def management(request, pk):
```
oskusalerma, line 648: in general function names should be verbs that indicate what they do. i'm not sure "management" is conveying what this view is actually doing.

oskusalerma (changes requested):

web/domains/case/_import/views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/263/commits/d7b4fab7541357a9dc033db45324ee668ea5185b. `management` becomes `manage_case`.

oskusalerma (approved):

trbateman (approved):

PR #264, opened 2021-02-09
ICMSLST-504 Stop OIL case (ILB Admin)

kevinetienne:
Allow ILB Admin to stop a case and to email participants.

image


oskusalerma (approved):

trbateman (approved):
Looks good to me. Just a question on code structure given _import/views.py et al might get very long when the other import journeys get added. Should these evolve later into distinct dirs with their own views.py forms.py? And how can we share the common parts

kevinetienne:
@trbateman good questions. I think I might create a directory but I'm not sure how as there's no clear pattern given the current structure.
I'm waiting for the breaking point for:
 - import applications, eg I need more information about the other apps
 - managing applications, eg. what can be shared between import/export applications and access requests.

The project has already a lot of technical debts and I'm taking an approach where I'm refactoring the code when I need it.

PR #266, opened 2021-02-12
ICMSLST-479-import-user-sanctions Add sanctions and adhoc license & r…

trbateman:
…efactor of firearms

Also:
 * Add sanctions and adhoc license views
 * Refactor firearms to be more generic
 * Added models to admin.py


web/domains/case/_import/sanctions/urls.py
```diff
@@ -0,0 +1,26 @@
+from django.urls import re_path
+
+from . import views
+
+urlpatterns = [
+    re_path(
+        "applicant-details/(?P<pk>[0-9]+)/",
+        views.sanctions_applicant_details,
```
oskusalerma, line 8: function names in general should be verbs since functions should _do_ stuff.

so this one for example could be `sanctions_show_applicant_details`

web/domains/case/_import/sanctions/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+from .models import SanctionsAndAdhocApplication
+
+logger = logging.getLogger(__name__)
+
+
+@login_required
+@permission_required("web.importer_access", raise_exception=True)
+def sanctions_applicant_details(request, pk):
+    with transaction.atomic():
```
oskusalerma, line 16: given this is a read-only view why do we need a transaction?

web/domains/case/_import/sanctions/views.py
```diff
@@ -0,0 +1,92 @@
...
+        application = get_object_or_404(
+            SanctionsAndAdhocApplication.objects.select_for_update(), pk=pk
```
oskusalerma, line 18: no need for select_for_update in read-only-views

web/domains/case/_import/urls.py
```diff
@@ -1,80 +1,13 @@
...
-        "^firearms/oil/(?P<pk>[0-9]+)/import-contacts/(?P<entity>legal|natural)/create/$",
-        views.create_import_contact,
-        name="create-import-contact",
-    ),
-    re_path(
-        "^firearms/oil/(?P<application_pk>[0-9]+)/import-contacts/(?P<entity>legal|natural)/(?P<contact_pk>[0-9]+)/edit/$",
-        views.edit_import_contact,
-        name="edit-import-contact",
-    ),
+    path("", views.ImportApplicationChoiceView.as_view(), name="create"),
```
oskusalerma, line 8: this is not a create view, it's a choice view

web/domains/case/_import/urls.py
```diff
@@ -1,80 +1,13 @@
...
+    path("create/sanctions/", views.create_sanctions, name="create-sanctions"),
+    path("create/firearms/oil/", views.create_oil, name="create-oil"),
```
oskusalerma, line 10: URL should be create/firearms-oil and name shoudl be create-firearms-oil

web/flow/models.py
```diff
@@ -88,3 +88,9 @@ class Task(models.Model):
         on_delete=models.CASCADE,
         related_name="+",
     )
+
+    def __unicode__(self):
```
oskusalerma, line 92: this is not needed in python 3

web/domains/case/_import/views.py
```diff
@@ -37,657 +20,44 @@ class ImportApplicationChoiceView(TemplateView, PermissionRequiredMixin):
...
-                application.status = ImportApplication.STOPPED
+                application = model_class()
+                application.importer = form.cleaned_data["importer"]
+                application.importer_office = form.cleaned_data["importer_office"]
+                application.process_type = model_class.PROCESS_TYPE
+                application.created_by = request.user
+                application.last_updated_by = request.user
+                application.submitted_by = request.user
+                application.application_type = import_application_type
                 application.save()
```
oskusalerma, line 56: move the atomic block to be just these two save calls

oskusalerma:
tom please fix the commit message as we did on friday

oskusalerma (changes requested):

web/domains/case/_import/sanctions/urls.py
trbateman: Changed, others left as placeholders and not sure what they'll end up doing

web/domains/case/_import/sanctions/views.py
trbateman: Needed as get_task does a select for update, which must be done inside an atomic transaction

web/domains/case/_import/urls.py
trbateman: Done

PR #275, opened 2021-02-19
ICMSLST-536 Add country groups data migration.

trbateman:


web/domains/country/models.py
```diff
@@ -16,29 +16,24 @@ class Country(models.Model):
...
-            return "(new)"
+        return self.name
 
     @property
     def name_slug(self):
         return self.name.lower().replace(" ", "_")
 
     class Meta:
         ordering = ("name",)
+        unique_together = ["name", "is_active"]
```
oskusalerma, line 27: unique_together is deprecated: https://docs.djangoproject.com/en/3.1/ref/models/options/#unique-together

use UniqueConstraint instead. see web/domains/template/models.py for an example

web/domains/country/models.py
```diff
@@ -16,29 +16,24 @@ class Country(models.Model):
     hmrc_code = models.CharField(max_length=20, blank=False, null=False)
 
     def __str__(self):
-        if self.id:
```
oskusalerma, line 19: why are you changing this?

web/domains/country/models.py
```diff
@@ -16,29 +16,24 @@ class Country(models.Model):
...
     class Meta:
         ordering = ("name",)
+        unique_together = ["name", "is_active"]
 
 
 class CountryGroup(models.Model):
-    name = models.CharField(max_length=4000, blank=False, null=False)
-    comments = models.CharField(max_length=4000, blank=True, null=True)
+    name = models.CharField(max_length=4000, blank=False, null=False, unique=True)
+    comments = models.TextField(blank=True, null=True)
```
oskusalerma, line 32: why are you changing this from charfield to textfield, and in effect, removing the max length?

web/domains/country/models.py
```diff
@@ -16,29 +16,24 @@ class Country(models.Model):
...
     countries = models.ManyToManyField(Country, blank=True, related_name="country_groups")
 
     def __str__(self):
-        if self.id:
```
oskusalerma, line 38: why are you changing this?

web/migrations/0027_add_country_data.py
```diff
@@ -240,7 +240,7 @@ def load_country_data(apps, schema_editor):
                 hmrc_code="CG",
```
oskusalerma, line 240: umm, migration 26 has been merged to develop, me and kevin have run it, you can't just change history now.

oskusalerma (changes requested):

web/domains/country/models.py
trbateman: Returning name gives a more meaningful description of the object, especially in the admin.

web/domains/country/models.py
trbateman: Thanks for heads up. It's not deprecated at the moment but might be in the future, hence no deprecation warnings coming up.

web/domains/country/models.py
trbateman: Text field feels more appropriate given the string has line breaks in it. The default widget is therefore more suitable for django admin.

web/domains/country/models.py
trbateman: Same as above

web/migrations/0027_add_country_data.py
trbateman: Will give you a call about this.

oskusalerma (approved):

PR #285, opened 2021-02-26
ICMSLST-547 Add data migration for Template

kevinetienne:
I'm not sure why our data dump didn't include countries. It feels like
they have been missed.
Only one reference to CountryTranslationSet was found.


web/migrations/0037_add_template_data.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+def add_schedule_templates(apps, schema_editor):
+    Template = apps.get_model("web", "Template")
+    Template.objects.get_or_create(
+        start_datetime=pytz.timezone("UTC").localize(
+            datetime.strptime("13-FEB-2019 18:44:45", DATETIME_FORMAT), is_dst=None
+        ),
+        is_active=True,
+        template_name="CFS Schedule Template",
```
oskusalerma, line 72: doesn't this duplicate what's done in web/migrations/0011_add_cfs_schedule_data.py? see line 8

oskusalerma (changes requested):

trbateman (approved):

web/migrations/0037_add_template_data.py
kevinetienne: Fixed in the rebase https://github.com/uktrade/icms/pull/285/commits/38c096955bc45410c1627c3d3d14c21d3e8447ad. It doesn't make sense to keep it here as the other migration add reverse fks to it.

oskusalerma (approved):

PR #287, opened 2021-03-03
ICMSLST-541 Add add_dummy_data management command.

oskusalerma:
This will make it trivial for developers to start from a clean DB in a
few seconds.


kevinetienne (approved):
Looks good to me.

PR #289, opened 2021-03-08
ICMSLST-516 Add editing of wood quota applications (basic + exporter).

oskusalerma:


web/domains/case/_import/wood/forms.py
```diff
@@ -0,0 +1,57 @@
...
+            "contact",
+            "applicant_reference",
+            "shipping_year",
+            "exporter_name",
+            "exporter_address",
+            "exporter_vat_nr",
+        )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
kevinetienne, line 57: I suspect your next PR would need to use __init__? Otherwise you can delete it.

kevinetienne (approved):

PR #290, opened 2021-03-09
ICMSLST-525 Add sanction and adhoc licence application details goods

trbateman:


web/domains/case/_import/sanctions/views.py
```diff
@@ -24,18 +25,26 @@ def edit_sanctions_and_adhoc_licence_application(request, pk):
...
+        application_started = False
+        if application.origin_country is not None and application.consignment_country is not None:
+            application_started = True
+
+        if request.method == "POST" and "delete" in request.POST.get("action", []):
+            goods_pk = request.POST.get("item")
+            SanctionsAndAdhocApplicationGoods.objects.get(pk=str(goods_pk)).delete()
+            return redirect(
+                reverse("import:edit-sanctions-and-adhoc-licence-application", kwargs={"pk": pk})
+            )
```
kevinetienne, line 37: As you did for adding/editing goods, this should go in its own view.

web/domains/case/_import/sanctions/views.py
```diff
@@ -24,18 +25,26 @@ def edit_sanctions_and_adhoc_licence_application(request, pk):
         if not request.user.has_perm("web.is_contact_of_importer", application.importer):
             raise PermissionDenied
 
+        application_started = False
+        if application.origin_country is not None and application.consignment_country is not None:
+            application_started = True
+
+        if request.method == "POST" and "delete" in request.POST.get("action", []):
+            goods_pk = request.POST.get("item")
+            SanctionsAndAdhocApplicationGoods.objects.get(pk=str(goods_pk)).delete()
```
kevinetienne, line 34: I don't think you don't need to cast `goods_pk` to a string, but let me know if you need it.

web/domains/case/_import/sanctions/views.py
```diff
@@ -55,6 +68,93 @@ def edit_sanctions_and_adhoc_licence_application(request, pk):
...
+    with transaction.atomic():
+        application = get_object_or_404(
+            SanctionsAndAdhocApplication.objects.select_for_update(), pk=pk
+        )
+        task = application.get_task(ImportApplication.IN_PROGRESS, "prepare")
+
+        if not request.user.has_perm("web.is_contact_of_importer", application.importer):
+            raise PermissionDenied
+
+        application_started = False
```
kevinetienne, line 83: See previous comment, I think you can remove this flag.

web/domains/case/_import/sanctions/views.py
```diff
@@ -55,6 +68,93 @@ def edit_sanctions_and_adhoc_licence_application(request, pk):
...
+                obj = form.save(commit=False)
+                obj.import_application = application
+                obj.save()
+                return redirect(
+                    reverse(
+                        "import:edit-sanctions-and-adhoc-licence-application", kwargs={"pk": pk}
+                    )
+                )
+        else:
+            form = GoodsForm(instance=SanctionsAndAdhocApplicationGoods.objects.get(pk=goodspk))
```
kevinetienne, line 142: I would use the reverse relationship to make sure the person editing the goods doesn't temper with the `goods_pk`.

web/templates/web/domains/case/import/sanctions/edit_sanctions_and_adhoc_licence_application.html
```diff
@@ -6,21 +6,83 @@
...
+                            <a class="link-button button icon-pencil" data-input_action="edit"
+                               href="{{ url('import:edit-goods', kwargs={'pk': process.pk, 'goodspk': row.pk}) }}"></a>
+                            <button role="action-button"
+                                    class="link-button button no-prompt-action tooltipstered icon-bin"
+                                    aria-label="Delete"
+                                    data-input_action="delete"
+                                    data-input_csrfmiddlewaretoken="{{ csrf_token }}"
+                                    data-input_item="{{ row.pk }}"
+                                    data-confirm="Are you sure you want to delete this record?">
+                            </button>
```
kevinetienne, line 68: Delete should be a form which would post to the new `delete_goods` view.

web/domains/case/_import/sanctions/forms.py
```diff
@@ -55,3 +53,13 @@ def __init__(self, *args, **kwargs):
...
+
+
+class GoodsForm(forms.ModelForm):
+    class Meta:
+        model = SanctionsAndAdhocApplicationGoods
+        fields = ["commodity_code", "goods_description", "quantity_amount", "value"]
+        widgets = {
+            "goods_description": forms.Textarea(attrs={"cols": 80, "rows": 20}),
+        }
+        labels = {"value": "Value (GBP CIF)"}
```
kevinetienne, line 65: I think you need validation here, related to your one or two country fields.

This is how I see it working in ICMS1:
 - only a subset of countries makes the application valid;
 - if the correct countries are selected the application for goods display the relevant commodities code.

I think the application should display an error if the relevant countries are not selected and avoid saving.
And the goods form should only display a subset of commodities.

kevinetienne (commented):
Thanks, @trbateman looking good I've mad a few points which is worth addressing before doing a demo on Thursday:
 - countries (for main application) and commodities (for goods)
 - adding a delete goods view (as for adding and editing goods)
 - `application_started` flag which might be linked to the missing validation
 - using the reverse relationship for goods in the edit goods view (will avoid tampering with the url)
 - url for edit view.

Let me know if my comments are not clear.

web/domains/case/_import/sanctions/views.py
trbateman: Moving to it's own view, with form in template. Move subtasks to outside current form block.

PR #298, opened 2021-03-17
ICMSLST-511 Edit Licence start and end dates

kevinetienne:


web/domains/case/_import/forms.py
```diff
@@ -110,3 +112,28 @@ class Meta:
...
+        end_date = data.get("licence_end_date")
+        if not start_date or not end_date:
+            return
+        today = timezone.now().date()
+
+        error_message = "Date must be in the future."
+        if start_date < today:
+            self.add_error("licence_start_date", error_message)
+        if start_date > end_date:
+            self.add_error("licence_end_date", error_message)
```
oskusalerma, line 139: isn't this a separate error with a different message?

oskusalerma (changes requested):

web/domains/case/_import/forms.py
kevinetienne: Fixed in https://github.com/uktrade/icms/pull/298/commits/e1a52ddb7ce483571e84616176fde84d14f193c1

oskusalerma (approved):

PR #301, opened 2021-03-17
ICMSLST-511 Generate pdf preview for licence and cover letter

kevinetienne:


web/domains/case/_import/views.py
```diff
@@ -1,13 +1,17 @@
...
 from django.db import transaction
 from django.db.models import Q
+from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render, reverse
+from django.template.loader import render_to_string
 from django.utils import timezone
 from django.views.decorators.http import require_POST
 from django.views.generic import TemplateView
 from guardian.shortcuts import get_users_with_perms
+from weasyprint import HTML
```
oskusalerma, line 14: don't do that, extremely confusing name later in the file. just import weasyprint and then do `weasyprint.HTML(...)` later

web/domains/case/_import/views.py
```diff
@@ -694,3 +698,64 @@ def delete_endorsement(request, application_pk, endorsement_pk):
...
+        }
+
+        html_string = render_to_string(
+            request=request,
+            template_name="web/domains/case/import/preview-cover-letter.html",
+            context=context,
+        )
+
+        html = HTML(string=html_string, base_url=request.build_absolute_uri())
+        pdf_file = html.write_pdf()
```
oskusalerma, line 727: in an earlier job we hit problems with weasyprint on server machines that didn't have X server running
which they obviously won't
everything was fine on developer machines, but production didn't work
so i would like you to add a test using weasyprint to create a PDF
so we know whether it works without an X server

web/domains/case/_import/views.py
```diff
@@ -694,3 +698,64 @@ def delete_endorsement(request, application_pk, endorsement_pk):
...
+
+        response = HttpResponse(pdf_file, content_type="application/pdf;")
+        response["Content-Disposition"] = "filename=Cover Letter.pdf"
```
oskusalerma, line 730: do we really want a space in the filename?

web/domains/case/_import/models.py
```diff
@@ -265,6 +266,10 @@ def application_started(self):
     def application_approved(self):
         return self.decision == self.APPROVE
 
+    @property
+    def licence_issue_date(self):
+        return self.issue_date or timezone.now()
```
oskusalerma, line 271: issue_date is Date, timezone.now() is Datetime, so this seems wrong

oskusalerma (changes requested):

web/domains/case/_import/models.py
kevinetienne: Fixed in https://github.com/uktrade/icms/pull/301/commits/e4055daa0ab8f6c1b0dd2b9b741e4cb609aac359, calling now `timezone.now().date()`.

web/domains/case/_import/views.py
kevinetienne: You're probably right I've copied the filename from the initial system. I've removed it in https://github.com/uktrade/icms/pull/301/commits/e4055daa0ab8f6c1b0dd2b9b741e4cb609aac359

web/domains/case/_import/views.py
kevinetienne: Added two tests in https://github.com/uktrade/icms/pull/301/commits/e4055daa0ab8f6c1b0dd2b9b741e4cb609aac359. It looks like early versions of weasyprint were relying on `GTK/GDK` which in turn used to require some x libraries.

web/domains/case/_import/views.py
kevinetienne: You're totally right fixed in https://github.com/uktrade/icms/pull/301/commits/e4055daa0ab8f6c1b0dd2b9b741e4cb609aac359

web/tests/domains/case/_import/test_views.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    process_factories.TaskFactory.create(process=process, task_type="process")
+
+    client = Client()
+    client.login(username=ilb_admin.username, password="test")
+    response = client.get(f"/import/case/{process.pk}/cover-letter/preview/")
+
+    assert response.status_code == 200
+    assert response["Content-Type"] == "application/pdf"
+    assert response["Content-Disposition"] == "filename=CoverLetter.pdf"
```
oskusalerma, line 35: can you check the actual data a little bit, so there's some confidence the PDF generation has worked? PDF files start with `%PDF-` (see https://en.wikipedia.org/wiki/List_of_file_signatures), and I guess you could also check the file size is within 10% of what you're expecting?

web/tests/domains/case/_import/test_views.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    process_factories.TaskFactory.create(process=process, task_type="process")
+
+    client = Client()
+    client.login(username=ilb_admin.username, password="test")
+    response = client.get(f"/import/case/{process.pk}/licence/preview/")
+
+    assert response.status_code == 200
+    assert response["Content-Type"] == "application/pdf"
+    assert response["Content-Disposition"] == "filename=Licence.pdf"
```
oskusalerma, line 61: same here

oskusalerma (changes requested):

web/tests/domains/case/_import/test_views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/301/commits/db42dbf0c52935f595c45c6040a4e47d92cf5418

web/tests/domains/case/_import/test_views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/301/commits/815e639c822d1e96d7eb60cc7b1769ebdb09192c

oskusalerma (approved):

PR #305, opened 2021-03-19
ICMSLST-358 Add create derogations from sanctions ban.

trbateman:


web/domains/case/_import/derogations/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
+
+from . import views
+
```
kevinetienne, line 4: can you include the a`pp_name` please:

```
app_name = "derogations"
urlpatterns = [
]
```

web/domains/case/_import/derogations/views.py
```diff
@@ -0,0 +1,48 @@
...
+from django.core.exceptions import PermissionDenied
+from django.db import transaction
+from django.shortcuts import get_object_or_404, redirect, render, reverse
+
+from web.domains.case._import.models import ImportApplication
+
+from .forms import DerogationsForm
+from .models import DerogationsApplication
+
+logger = logging.getLogger(__name__)
```
kevinetienne, line 12: it looks like you're not using the logger maybe remove it. However maybe you'll need it in a subsequent pr?

oskusalerma:
The failing mypy check:

`web/admin.py:16: error: Module 'web.models' has no attribute 'DerogationsApplication'`

is probably because you need to add this to web/models.py:

`from web.domains.case._import.sanctions.models.py import *  # NOQA`

trbateman:
> web/models.py



> The failing mypy check:
> 
> `web/admin.py:16: error: Module 'web.models' has no attribute 'DerogationsApplication'`
> 
> is probably because you need to add this to web/models.py:
> 
> `from web.domains.case._import.sanctions.models.py import * # NOQA`

Thanks Osku. I can't see a file web/models.py, did you mean web/models/__init_.py? 

DerogationsApplication is actually in web.domains.case._import.derogations.models, not sanctions.modeks.py and both are already added to web/models/__init_.py.

web/domains/case/_import/derogations/urls.py
trbateman: Are you sure we want to do this? We've done it at the import level, but not any of the lower levels on wood, sanctions, firearms or derogations.

web/domains/case/_import/derogations/views.py
trbateman: Removed

web/domains/case/_import/derogations/urls.py
kevinetienne: Yes, we are going to a few import application types and it will avoid to have a naming clash, thanks.

oskusalerma (approved):

PR #309, opened 2021-03-24
ICMSLST-560 Add edit goods section of derogation from sanctions ban.

trbateman:


web/templates/web/domains/case/import/derogations/edit_derogations.html
```diff
@@ -20,7 +20,19 @@ <h4>Sanctions Derogation</h4>
...
+
+                        Please upload your contract in the Supporting Documents section below.
+
+    </div>
+
+    <h4>Goods</h4>
+    {{ fields.field(form.commodity_code) }}
+    {{ fields.field(form.goods_description) }}
+    {{ fields.field(form.quantity) }}
+    {{ fields.field(form.value) }}
```
kevinetienne, line 33: It looks like you need to include the unit in the form for the user?

web/domains/case/_import/derogations/forms.py
```diff
@@ -40,6 +41,32 @@ class DerogationsForm(forms.ModelForm):
         widget=forms.Textarea(attrs={"cols": 50, "rows": 3}),
     )
 
+    # TODO restrict commodities to group when it becomes known
```
kevinetienne, line 44: Another solution could be to hardcode the commodities in a choicefield instead as done in https://github.com/uktrade/icms/blob/6513be8bb47c417631b32737d4946f7838a5785c/web/domains/case/_import/wood/forms.py#L45-L53

kevinetienne (commented):
I think you need to add the unit to help the user filling the form.

web/templates/web/domains/case/import/derogations/edit_derogations.html
trbateman: Discuss to make these changes on the next ticket

kevinetienne (approved):
`commodity_code` and `units` for goods would be addressed in a separate ticket.

PR #315, opened 2021-03-30
ICMSLST-526 Add supporting documents for Sanctions and Adhoc Application

trbateman:


web/templates/web/domains/case/import/sanctions/add_document.html
```diff
@@ -0,0 +1,56 @@
...
+      </div>
+    </div>
+
+  {%- endcall %}
+
+{% endblock main_content %}
+
+{% block task_actions %}
+    <input type="submit" name="save" class="primary-button button" value="Save"/>
+{% endblock task_actions %}
```
kevinetienne, line 54: It looks like you have two submit buttons in this page.

kevinetienne (approved):
The view to display/download document can maybe made generic but as discussed I'll do it later as I needed the same view for my upcoming change in #317.

PR #316, opened 2021-03-30
ICMSLST-529 Add submit sanctions application.

trbateman:


web/domains/case/_import/sanctions/views.py
```diff
@@ -187,19 +189,41 @@ def sanctions_validation_summary(request, pk):
...
+                return redirect(reverse("home"))
+
+        else:
+            form = SubmitSanctionsForm()
+
     context = {
         "process_template": "web/domains/case/import/partials/process.html",
         "process": application,
         "task": task,
+        "form": form,
```
kevinetienne, line 226: Missing the declaration from the template see https://github.com/uktrade/icms/blob/b03b50851a9ea9ca118fb2399e0a4cd1d9de7454/web/domains/case/_import/firearms/views.py#L440

kevinetienne (approved):

PR #322, opened 2021-04-07
ICMSLST-557 Add editing of wood quota applications (contract documents)

oskusalerma:


web/domains/case/_import/views.py
```diff
@@ -849,22 +849,21 @@ def _view_file(request, application, related_file_model, file_pk):
     if not has_perm_importer and not has_perm_reference_data:
         raise PermissionDenied
 
-    with transaction.atomic():
```
oskusalerma, line 852: note to kevin: this is a read-only function, it's a remnant of how the code evolved over time that it had an atomic block in there, it's not needed so i'm removing it

web/domains/case/_import/wood/views.py
```diff
@@ -111,3 +118,110 @@ def delete_supporting_document(request, application_pk, document_pk):
...
+            raise PermissionDenied
+
+        document = application.contract_documents.get(pk=document_pk)
+        document.is_active = False
+        document.save()
+
+        return redirect(reverse("import:wood:edit-quota", kwargs={"pk": application_pk}))
+
+
+def edit_contract_document(request, application_pk, document_pk):
```
kevinetienne, line 195: Do you need `@login_required` and `@permission_required`?

oskusalerma: indeed i do

kevinetienne (approved):

PR #337, opened 2021-04-21
ICMSLST-619 Adapt firearms import application to new validations.

oskusalerma:


web/domains/case/_import/firearms/models.py
```diff
@@ -29,6 +31,8 @@ class ConstabularyEmail(models.Model):
     STATUSES = ((OPEN, "Open"), (CLOSED, "Closed"), (DRAFT, "Draft"))
 
     is_active = models.BooleanField(blank=False, null=False, default=True)
+
+    # TODO: link to OpenIndividualLicenceApplication instead, add a related_name
```
MattHolmes123, line 35: Do we add TODO's without a ticket in Jira?  A question about process rather

kevinetienne (approved):
:+1:

MattHolmes123 (approved):

PR #339, opened 2021-04-23
ICMSLST-624 Update firearms model links and add related_name values.

MattHolmes123:
All models should correctly link to OpenIndividualLicenceApplication.


web/domains/case/_import/firearms/models.py
```diff
@@ -61,7 +64,11 @@ class UserImportCertificate(models.Model):
         ("shotgun", "Shotgun Certificate"),
     )
 
-    import_application = models.ForeignKey(ImportApplication, on_delete=models.PROTECT)
+    import_application = models.ForeignKey(
+        OpenIndividualLicenceApplication,
+        on_delete=models.PROTECT,
+        related_name="user_certificates",
```
oskusalerma, line 70: maybe `user_imported_certificates`? `user_certificates` sounds like something ABOUT users

oskusalerma (changes requested):

web/domains/case/_import/firearms/models.py
MattHolmes123: Updateede

web/domains/case/_import/firearms/views.py
```diff
@@ -92,7 +92,7 @@ def list_user_import_certificates(request, pk):
             "process_template": "web/domains/case/import/partials/process.html",
             "process": application,
             "task": task,
-            "certificates": application.userimportcertificate_set.all(),
+            "certificates": application.user_certificates.all(),
```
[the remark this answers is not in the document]
MattHolmes123: There are a couple of places the partial template is used, so i think its best to leave it as "certificates"

oskusalerma (approved):

PR #340, opened 2021-04-23
ICMSLST-621 Add commodity data

kevinetienne:
* use CharField for commodity code in firearms and sanctions
 * remove CommodityType.allowed_code


web/migrations/0052_add_commodity_data.py
```diff
@@ -0,0 +1,12380 @@
...
+from django.db import migrations
+
+DATETIME_FORMAT = "%d-%b-%Y %H:%M:%S"
+DATE_FORMAT = "%d-%b-%Y"
+
+
+def add_units(apps, schema_editor):
+    Unit = apps.get_model("web", "Unit")
+
+    Unit.objects.get_or_create(
```
MattHolmes123, line 15: I think this will slow down the tests quite a bit.

We should maybe use this rather than get_or_create as we know its new data:
https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-create
It includes an example of inserting data in batches.

We could also record the time it takes to run these migrations both ways to see how much better it is

MattHolmes123 (changes requested):

web/migrations/0052_add_commodity_data.py
oskusalerma: I agree with Matt's comment, thousands of get_or_create calls will be quite slow, bulk_create would be much better.

oskusalerma (changes requested):
see above

oskusalerma (approved):

MattHolmes123 (approved):

PR #353, opened 2021-04-30
ICMSLST-640 Add view wood application view.

MattHolmes123:
Other changes:
 - Refactor edit wood application to use same partial templates.
 - Take ownership link now redirects to view application view.

### DIFF
--- a/web/domains/case/_import/views.py
+++ b/web/domains/case/_import/views.py
@@ -133,14 +133,16 @@ def _create_application(
 @login_required
 @permission_required("web.reference_data_access", raise_exception=True)
 @require_POST
-def take_ownership(request, pk):
+def take_ownership(request: HttpRequest, pk: int) -> HttpResponse:
     with transaction.atomic():
-        application = get_object_or_404(ImportApplication.objects.select_for_update(), pk=pk)
+        application: ImportApplication = get_object_or_404(
+            ImportApplication.objects.select_for_update(), pk=pk
+        )
         application.get_task([ImportApplication.SUBMITTED, ImportApplication.WITHDRAWN], "process")
         application.case_owner = request.user
         application.save()
 
-        return redirect(reverse("workbasket"))
+        return redirect(reverse("import:case-management", kwargs={"pk": application.pk}))
 
 
 @login_required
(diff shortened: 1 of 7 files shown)


web/domains/case/_import/views.py
```diff
@@ -135,11 +135,24 @@ def _create_application(
...
 
+        if task.process.process_type in (
+            DerogationsApplication.PROCESS_TYPE,
+            OpenIndividualLicenceApplication.PROCESS_TYPE,
+            SanctionsAndAdhocApplication.PROCESS_TYPE,
+            WoodQuotaApplication.PROCESS_TYPE,
+        ):
+            return redirect(reverse("import:view-case", kwargs={"pk": application.pk}))
+
+        # Default to workbasket if "view-case" not implemented for the application type.
```
kevinetienne, line 155: should we raise a `NotImplemented` instead? It will force developers to add application process type to the list?

web/domains/case/_import/views.py
```diff
@@ -135,11 +135,24 @@ def _create_application(
...
         application.case_owner = request.user
         application.save()
 
+        if task.process.process_type in (
+            DerogationsApplication.PROCESS_TYPE,
+            OpenIndividualLicenceApplication.PROCESS_TYPE,
+            SanctionsAndAdhocApplication.PROCESS_TYPE,
+            WoodQuotaApplication.PROCESS_TYPE,
+        ):
+            return redirect(reverse("import:view-case", kwargs={"pk": application.pk}))
```
kevinetienne, line 153: I think it should go to [`import:case-management`](https://github.com/uktrade/icms/blob/e60b403e214a42f98030ca010c849ba94d013ca8/web/domains/case/_import/urls.py#L46) instead of `view-case`.

I've tried with a view application and it went in the case management page. Maybe it goes somewhere else for other applications?

kevinetienne (commented):
Added two comments regarding:
 - raising an error when process type is not listed when taking ownership of the case
 - redirecting to the case to manage it when ownership has been granted.

oskusalerma (changes requested):

web/domains/case/_import/views.py
MattHolmes123: Agreed :+1:

oskusalerma (approved):

PR #355, opened 2021-05-04
ICMSLST-641 Add Checklist for Wood Quota applications.

MattHolmes123:


web/domains/case/_import/wood/models.py
```diff
@@ -34,3 +34,61 @@ class WoodQuotaApplication(ImportApplication):
...
+
+
+class WoodQuotaChecklist(models.Model):
+    class Response(models.TextChoices):
+        yes = ("yes", "Yes")
+        no = ("no", "No")
+        not_applicable = ("n/a", "N/A")
+
+    import_application = models.ForeignKey(
+        WoodQuotaApplication, on_delete=models.PROTECT, related_name="checklists"
```
oskusalerma, line 46: related_name is too generic, how about woodquota_checklists?

web/domains/case/_import/wood/models.py
```diff
@@ -34,3 +34,61 @@ class WoodQuotaApplication(ImportApplication):
...
+    )
+
+    sigl_wood_application_logged = models.BooleanField(
+        default=False,
+        verbose_name="Log on to SIGL Wood via this screen. Processing done on SIGL Wood.",
+    )
+
+    case_update = models.CharField(
```
oskusalerma, line 54: could these yes/no/na fields be nullable booleans?

oskusalerma (dismissed):

web/domains/case/_import/wood/models.py
MattHolmes123: No I thought that in the other checklist I did.
N/a is a valid required field

web/domains/case/_import/wood/models.py
MattHolmes123: The equivalent related names on the two other models are "checklists".

MattHolmes123: I don't feel to strongly either way, I can change it if you want.
@kevinetienne what do you think?

kevinetienne (approved):

PR #362, opened 2021-05-06
ICMSLST-642 Add function to print database queries.

MattHolmes123:


web/utils/db.py
```diff
@@ -0,0 +1,62 @@
...
+    queries_before_debug = connections["default"].queries
+
+    try:
+        yield
+
+    finally:
+        show_queries(queries_before_debug, show_time, fancy_print)
+
+
+def show_queries(queries_before_debug: List[Query], show_time: bool, fancy_print: bool) -> None:
```
oskusalerma, line 42: internal function, maybe add a `_` prefix?

web/utils/db.py
```diff
@@ -0,0 +1,62 @@
...
+    ignore = len(queries_before_debug)
+    queries_to_debug: List[Query] = connections["default"].queries[ignore:]
+
+    for query in queries_to_debug:
+        sql = str(query["sql"])
```
oskusalerma, line 47: str is not needed

web/utils/db.py
oskusalerma, line 42: maybe just pass in the number of queries to ignore

oskusalerma (changes requested):

oskusalerma (approved):

PR #366, opened 2021-05-06
ICMSLST-653 Bump max_length on reference fields from 50 to 100.

oskusalerma:
Also change "make migrations" to chown files from root.


kevinetienne (approved):
:100:

MattHolmes123 (approved):

PR #368, opened 2021-05-07
ICMSLST-381 Add Import Contacts to DFLApplication import type.

MattHolmes123:
Other changes:

- Refactor of the import Import Contact urls, views & templates.


web/domains/case/_import/urls.py
```diff
@@ -13,11 +13,34 @@
...
+                path("dfl/", include("web.domains.case._import.fa_dfl.urls")),
+                path("oil/", include("web.domains.case._import.fa_oil.urls")),
+                # Firearms and Ammunition - Import Contact urls
+                path(
+                    "<int:pk>/import-contacts/",
+                    views.list_import_contacts,
+                    name="fa-list-import-contacts",
+                ),
+                re_path(
+                    "^(?P<pk>[0-9]+)/import-contacts/(?P<entity>legal|natural)/create/$",
```
oskusalerma, line 31: plase stick to path, use normal int/str captures, and add an "else: raise Error("invalid entity") in the view

oskusalerma (changes requested):

web/templates/web/domains/case/import/fa-import-contacts/list.html
```diff
@@ -0,0 +1,38 @@
+{% extends "flow/task-base.html" %}
```
kevinetienne, line 1: I wonder if we should just get rid of the sidebar for `know brought from` and just add the list and the buttons on the main page as a table.
The advantage of having it on the same page would be to avoid to do type checking and the disadvantage is that it would make the form longer.

In any case if we decide to have it as a table in the main form we should do it in another ticket maybe

MattHolmes123: Yer lets revisit this when we have done all three firearm importer types.

oskusalerma (approved):

PR #373, opened 2021-05-10
ICMSLST-656 Edit Firearms and Ammunition - Specific Import Licence (SIL)

kevinetienne:
Edit application for importer.

### DIFF
--- a/web/domains/case/_import/fa_sil/forms.py
+++ b/web/domains/case/_import/fa_sil/forms.py
@@ -1,0 +1,116 @@
+from django import forms
+from guardian.shortcuts import get_users_with_perms
+
+from web.domains.country.models import Country
+from web.domains.user.models import User
+
+from . import models
+
+
+class PrepareSILForm(forms.ModelForm):
+    applicant_reference = forms.CharField(
+        label="Applicant's Reference",
+        help_text="Enter your own reference for this application.",
+        required=False,
+    )
+
+    contact = forms.ModelChoiceField(
+        queryset=User.objects.none(),
+        help_text="Select the main point of contact for the case. This will usually be the person who created the application.",
+    )
+
+    section1 = forms.BooleanField(required=False, label="Firearms Licence for")
+    section2 = forms.BooleanField(required=False, label="")
+    section5 = forms.BooleanField(required=False, label="")
+    section58_obsolete = forms.BooleanField(required=False, label="")
+    section58_other = forms.BooleanField(required=False, label="")
+    other_description = forms.CharField(
+        required=False,
+        label="Other Section Description",
+        help_text="""
+            If you have selected Other in Firearms Act Sections. Please explain why you are making
+            this request under this 'Other' section.
+        """,
+    )
+
+    commodity_code = forms.ChoiceField(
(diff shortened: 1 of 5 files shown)


web/domains/case/_import/fa_sil/forms.py
```diff
@@ -0,0 +1,116 @@
...
+    section5 = forms.BooleanField(required=False, label="")
+    section58_obsolete = forms.BooleanField(required=False, label="")
+    section58_other = forms.BooleanField(required=False, label="")
+    other_description = forms.CharField(
+        required=False,
+        label="Other Section Description",
+        help_text="""
+            If you have selected Other in Firearms Act Sections. Please explain why you are making
+            this request under this 'Other' section.
+        """,
```
MattHolmes123, line 33: this should be written like this:
```
help_text=(
    "fooooooooooooo"
    " bar"
)
```

web/domains/case/_import/fa_sil/forms.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+    commodity_code = forms.ChoiceField(
+        label="Commodity Code",
+        help_text="""
+            You must pick the commodity code group that applies to the items that you wish to
+            import. Please note that "ex Chapter 97" is only relevant to collectors pieces and
+            items over 100 years old. Please contact HMRC classification advisory service,
+            01702 366077, if you are unsure of the correct code.
+        """,
```
MattHolmes123, line 43: as above.

web/domains/case/_import/fa_sil/forms.py
```diff
@@ -0,0 +1,116 @@
...
+                ("true", "Yes"),
+                ("false", "No"),
+            ]
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        # At least one section should be selected
+        licence_for = ["section1", "section2", "section5", "section58_obsolete", "section58_other"]
+        sections = [cleaned_data.get(section) for section in licence_for]
```
MattHolmes123, line 105: If you change this to a generator expression it will stop evaluating as soon as it needs to rather than create a full list.  
It won't make any noticeable difference but still :) 
an example:  
```
>> test = (x for x in range(10))
>> not any(test)
False
>> print(next(test))
2
```

web/domains/case/_import/fa_sil/forms.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/373/commits/56708e9a1774afda1d84377173a4879ffa329f51

web/domains/case/_import/fa_sil/forms.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/373/commits/56708e9a1774afda1d84377173a4879ffa329f51

web/domains/case/_import/fa_sil/forms.py
kevinetienne: Fixed in https://github.com/uktrade/icms/pull/373/commits/56708e9a1774afda1d84377173a4879ffa329f51

PR #375, opened 2021-05-11
ICMSLST-556 Add DLF Application Goods Certificates.

MattHolmes123:
Features:
  - Add goods section to main form
  - List goods certificates
  - Add good certificate
  - Edit goods certificate
  - Delete goods certificate
  - View goods certificate


web/domains/case/_import/fa_dfl/models.py
```diff
@@ -1,18 +1,63 @@
...
+class DFLGoodsCertificate(File):
+    """Deactivated Firearms Licence Goods certificate"""
+
+    goods_description = models.CharField(
+        max_length=4096, verbose_name="Goods Description", blank=False, null=False
+    )
+
+    deactivated_certificate_reference = models.CharField(
+        max_length=50, verbose_name="Deactivated Certificate Reference", blank=False, null=False
+    )
```
kevinetienne, line 18: Are `blank=False` and `null=False` the default for django model fields? If yes maybe we might not need them.

web/domains/case/_import/fa_dfl/models.py
```diff
@@ -1,18 +1,63 @@
...
+        dcf = self.deactivated_certificate_reference
+        return f"DFLGoodsCertificate(id={self.pk}, deactivated_certificate_reference={dcf!r})"
 
 
 class DFLApplication(ImportApplication):
     """Firearms & Ammunition Deactivated Firearms Licence application"""
 
+    class CommodityCodes(models.TextChoices):
+        EX_CHAPTER_93 = "ex Chapter 93", "ex Chapter 93"
+        EX_CHAPTER_97 = "ex Chapter 97", "ex Chapter 97"
```
kevinetienne, line 39: Can you add the parenthesis for the tuples please?

web/domains/case/_import/models.py
```diff
@@ -234,13 +238,15 @@ class ImportApplication(WorkbasketBase, Process):
...
         related_name="import_applications_from",
+        verbose_name="Country Of Origin",
     )
     consignment_country = models.ForeignKey(
         Country,
         on_delete=models.PROTECT,
         blank=True,
         null=True,
         related_name="import_applications_to",
+        verbose_name="Country Of Consignment",
```
kevinetienne, line 249: Does adding `verbose_name` require to update the migrations?

MattHolmes123: Ahh yes it does... are we ok with this?
I just noticed those labels were defined in 3 or so forms when in reality it should be the verbose_name

web/domains/case/_import/fa_dfl/forms.py
```diff
@@ -79,5 +64,58 @@ def __init__(self, *args, **kwargs):
...
+            country_groups__name="Firearms and Ammunition (Deactivated) Issuing Countries"
+        ),
+    )
+
+
+class EditDLFGoodsCertificateForm(forms.ModelForm):
+    class Meta:
+        model = models.DFLGoodsCertificate
+        fields = ("goods_description", "deactivated_certificate_reference", "issuing_country")
+        help_texts = {
```
oskusalerma, line 107: empty line above please

web/domains/case/_import/fa_dfl/models.py
```diff
@@ -1,18 +1,63 @@
...
+from web.domains.constabulary.models import Constabulary
+from web.domains.country.models import Country
+from web.domains.file.models import File
+
+
+class DFLGoodsCertificate(File):
+    """Deactivated Firearms Licence Goods certificate"""
+
+    goods_description = models.CharField(
+        max_length=4096, verbose_name="Goods Description", blank=False, null=False
```
oskusalerma, line 13: how about it verbose name is exactly what django would create anyway we don't manually specify it

oskusalerma (approved):

PR #380, opened 2021-05-13
ICMSLST-659 Refactor case management code (notes).

oskusalerma:
This adds the framework of being able to have just one URL/view for shared
code in case management for all of import/export/accessrequest, and also
uses it to move all note functions over.


MattHolmes123 (approved):
Changes look good.

kevinetienne (approved):

PR #385, opened 2021-05-17
ICMSLST-666 Refactor applicant case management urls.

MattHolmes123:


web/domains/case/urls.py
```diff
@@ -44,6 +44,13 @@
     ),
 ]
 
+applicant_case_management_urls = [
```
oskusalerma, line 47: this file is case/urls.py, everything here is about case management. i think this can be called just applicant_urls. same for the actual URL, can be just "applicant/"

web/domains/case/views.py
```diff
@@ -47,6 +49,30 @@ def _get_class_imp_or_exp_or_access(case_type: str) -> ImpOrExpOrAccessT:
...
 
+def _has_importer_exporter_access(user: User, case_type: str) -> bool:
+    has_permission = False
+
+    if case_type == "import":
+        has_permission = user.has_perm("web.importer_access")
+
+    elif case_type == "export":
+        has_permission = user.has_perm("web.exporter_access")
+
```
oskusalerma, line 60: throw exception is case_type is anything else. same in the other func

web/domains/case/views.py
```diff
@@ -296,6 +322,99 @@ def delete_note_document(
...
+        is_contact = _is_importer_exporter_contact(request.user, application, case_type)
+        if not is_contact:
+            raise PermissionDenied
+
+        if request.POST:
+            form = forms.WithdrawForm(request.POST)
+
+            if form.is_valid():
+                withdrawal = form.save(commit=False)
+                if case_type == "import":
```
oskusalerma, line 348: empty line before please

oskusalerma (changes requested):

oskusalerma (approved):

PR #392, opened 2021-05-21
ICMSLST-648 Fix update request for ILB Admin

kevinetienne:
- Remove non ILB Admin views for Update Requests (To be implemented in ICMLST-679)
 - Add missing template to create Update Request
 - Fix closing request
 - Move update requests urls from import to case
 - Make update request available to export case management (ICMSLST-432)

### DIFF
--- a/web/domains/case/urls.py
+++ b/web/domains/case/urls.py
@@ -81,6 +81,11 @@ further_information_requests_urls = [
     ),
 ]
 
+update_requests_urls = [
+    path("", views.manage_update_requests, name="manage-update-requests"),
+    path("<int:update_request_pk>/close/", views.close_update_request, name="close-update-request"),
+]
+
 urlpatterns = [
     path(
         "<casetype:case_type>/<int:application_pk>/",
@@ -98,15 +103,9 @@ urlpatterns = [
                 # ILB Admin Case management (import/export)
                 path("admin/", include(admin_urls)),
                 #
-                # TODO: (ICMSLST-648)
-                # path(
-                #     "update-requests/", views.manage_update_requests, name="manage-update-requests"
-                # ),
-                # path(
-                #     "update-requests/<int:update_request_pk>/close/",
-                #     views.close_update_requests,
-                #     name="close-update-requests",
-                # ),
+                # update requests
+                # TODO: implement importer/exporter response: (ICMSLST-679)
+                path("update-requests/", include(update_requests_urls)),
                 #
                 # notes (import/export)
                 path("notes/", include(note_urls)),
--- a/web/domains/case/urls.py
+++ b/web/domains/case/urls.py
@@ -82,8 +82,15 @@ further_information_requests_urls = [
 ]
 
 update_requests_urls = [
-    path("", views.manage_update_requests, name="manage-update-requests"),
-    path("<int:update_request_pk>/close/", views.close_update_request, name="close-update-request"),
+    path("manage/", views.manage_update_requests, name="manage-update-requests"),
+    path(
+        "<int:update_request_pk>/",
+        include(
+            [
+                path("close/", views.close_update_request, name="close-update-request"),
+            ]
+        ),
+    ),
 ]
 
 urlpatterns = [
--- a/web/domains/case/views.py
+++ b/web/domains/case/views.py
@@ -420,15 +420,51 @@ def archive_withdrawal(
         return redirect(reverse("workbasket"))
 
 
-def _manage_update_requests(
-    request, application, model_class, email_subject, email_content, contacts, url_namespace
-):
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def manage_update_requests(
+    request: HttpRequest, *, application_pk: int, case_type: str
+) -> HttpResponse:
+    model_class = _get_class_imp_or_exp(case_type)
+
     with transaction.atomic():
-        task = application.get_task([model_class.SUBMITTED, model_class.WITHDRAWN], "process")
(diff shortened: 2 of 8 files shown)


web/domains/case/urls.py
```diff
@@ -81,6 +81,11 @@
     ),
 ]
 
+update_requests_urls = [
+    path("", views.manage_update_requests, name="manage-update-requests"),
```
oskusalerma, line 85: can this url be "manage/", we don't have any other "empty" urls like this here

web/domains/case/urls.py
```diff
@@ -81,6 +81,11 @@
...
+    path("<int:update_request_pk>/close/", views.close_update_request, name="close-update-request"),
```
oskusalerma, line 86: if we're going to have respond-update-request etc in the future, could you add this under its own path already that captures the pk?

web/domains/case/views.py
```diff
@@ -420,16 +420,52 @@ def archive_withdrawal(
...
+        elif case_type == "export":
+            template_code = "CA_APPLICATION_UPDATE_EMAIL"
+
+            # TODO: replace with case reference ICMSLST-348
+            placeholder_content = {
+                "CASE_REFERENCE": application_pk,
+                "EXPORTER_NAME": application.exporter.name,
+                "CASE_OFFICER_NAME": request.user,
+            }
+        else:
```
oskusalerma, line 461: line 428 will already have errored out if case_type is not in import/export, this is not needed

web/domains/case/views.py
```diff
@@ -439,7 +475,7 @@ def _manage_update_requests(
                 update_request.status = models.UpdateRequest.OPEN
                 update_request.save()
 
-                application.status = model_class.UPDATE_REQUESTED
+                application.status = ImportApplication.UPDATE_REQUESTED
```
oskusalerma, line 478: this seems like a backwards change...? why would export application use an importapplication constant?

web/domains/case/views.py
```diff
@@ -449,8 +485,22 @@ def _manage_update_requests(
...
 
+                if case_type == "import":
+                    contacts = get_users_with_perms(
+                        application.importer, only_with_perms_in=["is_contact_of_importer"]
+                    ).filter(user_permissions__codename="importer_access")
+                elif case_type == "export":
+                    contacts = get_users_with_perms(
+                        application.exporter, only_with_perms_in=["is_contact_of_exporter"]
+                    ).filter(user_permissions__codename="exporter_access")
+                else:
```
oskusalerma, line 496: same, case_type has already been checked to be import/export

web/domains/case/views.py
```diff
@@ -463,14 +513,14 @@ def _manage_update_requests(
...
-            "process_template": f"web/domains/case/{url_namespace}/partials/process.html",
+            "process_template": f"web/domains/case/{case_type}/partials/process.html",
             "process": application,
             "task": task,
             "page_title": f"{application.application_type.get_type_description()} - Update Requests",
             "form": form,
             "update_requests": update_requests,
             "current_update_request": current_update_request,
-            "url_namespace": url_namespace,
+            "case_type": case_type,  # TODO: ICMSLST-670 remove
```
oskusalerma, line 523: umm...no. why would we ever remove case_type? this comment needs to go on url_namespace=case_type line. or if you've tested and this all works without url_namespace then you don't need any comment at all, and neither do you need url_namespace

web/domains/case/views.py
```diff
@@ -463,14 +513,14 @@ def _manage_update_requests(
             )
 
         context = {
-            "process_template": f"web/domains/case/{url_namespace}/partials/process.html",
+            "process_template": f"web/domains/case/{case_type}/partials/process.html",
```
oskusalerma, line 516: have you checked you need this anymore? a 1-minute browse through your templates seems to indicate to me you don't, but please try it out

oskusalerma (changes requested):

web/domains/case/views.py
kevinetienne: I confirm it's defined in the base template for case management https://github.com/uktrade/icms/blob/43577b3d64357e768652552f6bc685bc583fae67/web/templates/web/domains/case/manage/base.html#L6

web/domains/case/urls.py
kevinetienne: good idea fixed in https://github.com/uktrade/icms/pull/392/commits/c808497089ee578135b4f2fbca308193e2bfb673

web/domains/case/urls.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/392/commits/c808497089ee578135b4f2fbca308193e2bfb673 let me know if this wasn't what you meant.

web/domains/case/views.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/392/commits/c808497089ee578135b4f2fbca308193e2bfb673, there was another ImportApplication laying a few lines above too, thanks

web/domains/case/views.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/392/commits/c808497089ee578135b4f2fbca308193e2bfb673

web/domains/case/views.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/392/commits/c808497089ee578135b4f2fbca308193e2bfb673

oskusalerma (approved):

PR #400, opened 2021-05-25
ICMSLST-477 Add create screen for import/outward processing trade.

oskusalerma:
Also:

    ICMSLST-477 Rename pk to application_pk on import edit views.
    
    This makes it possible to group all application views under one include
    (this is not done on existing application types).


web/domains/case/_import/fa_dfl/views.py
```diff
@@ -36,9 +36,11 @@ def _get_page_title(page: str) -> str:
 
 @login_required
 @permission_required("web.importer_access", raise_exception=True)
-def edit_dlf(request: HttpRequest, pk: int) -> HttpResponse:
+def edit_dfl(request: HttpRequest, application_pk: int) -> HttpResponse:
```
MattHolmes123, line 39: kwargs only views

web/domains/case/_import/fa_oil/views.py
```diff
@@ -41,10 +41,10 @@
 
 @login_required
 @permission_required("web.importer_access", raise_exception=True)
-def edit_oil(request: HttpRequest, pk: int) -> HttpResponse:
+def edit_oil(request: HttpRequest, application_pk: int) -> HttpResponse:
```
MattHolmes123, line 44: same (and for any others that are missing)

web/management/commands/add_dummy_data.py
```diff
@@ -16,6 +17,12 @@ def handle(self, *args, **kwargs):
         if not settings.DEBUG:
             raise CommandError("DEBUG flag is not set, refusing to run!")
 
+        ImportApplicationType.objects.filter(
```
MattHolmes123, line 20: Do we want a comment to say we are adding this to test?

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #411, opened 2021-06-02
ICMSLST-647 Fix non-bugs found by mypy.

MattHolmes123:
Other changes:
  - enable check_untyped_defs flag
  - exclude migrations from mypy checks for speed

### DIFF
--- a/web/domains/country/views.py
+++ b/web/domains/country/views.py
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING, TypedDict
+
 import structlog as logging
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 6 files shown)


web/models/mixins.py
```diff
@@ -4,23 +4,23 @@
...
     # e.g. the system shouldn't allow you to have more than one active
     # translation for language X for some template (see what was done in
     # ICMSLST-483)
     def unarchive(self):
         self.is_active = True
-        self.save()
+        self.save()  # type: ignore[attr-defined]
 
 
 class Sortable(object):
```
kevinetienne, line 18: I think we can get rid of this class. We are inheriting from `Sortable` in https://github.com/uktrade/icms/blob/901a456733a734127d206af5473e6d62b740c52a/web/domains/firearms/models.py#L89

But `swap_order` is not used.

kevinetienne: Happy to create a ticket for it let me know.

kevinetienne: Created ICMSLST-700

kevinetienne (approved):

web/domains/country/views.py
```diff
@@ -221,6 +227,11 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class MissingTranslationsDict(TypedDict):
```
oskusalerma, line 230: all types at top of files

oskusalerma (changes requested):

web/domains/country/views.py
MattHolmes123: Done

oskusalerma (approved):

PR #419, opened 2021-06-09
ICMSLST-701 Refactor Importer/Exporter Access Request workbasket code.

MattHolmes123:
Move logic from templates to python model.


web/templates/web/domains/workbasket/workbasket.html
```diff
@@ -141,15 +141,15 @@ <h4>
 
                 <td>{{ row.subject }}</td>
 
-                <td>{{ row.company }}</td>
+                <td>{{ row.company|safe }}</td>
```
kevinetienne, line 144: should safe be [`nl2br`](https://github.com/uktrade/icms/blob/develop/web/jinja2.py#L17-L25)?  Same comment apply for other change in this file.

kevinetienne: Do we capture html for the `company` and `info`

MattHolmes123: I will have a look and more than likely change it to use nl2br

kevinetienne (approved):

web/domains/case/access/models.py
```diff
@@ -63,25 +62,99 @@ class AccessRequest(WorkbasketBase, Process):
...
+                    name="View",
+                    url=reverse(
+                        "case:view", kwargs={"application_pk": self.pk, "case_type": "access"}
+                    ),
+                )
+            ]
+
+            for fir in self.further_information_requests.open():
+                kwargs = {"application_pk": self.pk, "fir_pk": fir.pk, "case_type": "access"}
+                owner_actions.append(
```
oskusalerma, line 145: empty line before

web/domains/case/access/models.py
```diff
@@ -63,25 +62,99 @@ class AccessRequest(WorkbasketBase, Process):
...
+            info_rows.append("Approval Requested")
+
+        r.information = "\n".join(info_rows)
+
+        task = self.get_active_task()
+
+        if user.has_perm("web.reference_data_access"):
+            admin_actions: list[WorkbasketAction] = []
+
+            if self.process_type == "ExporterAccessRequest":
```
oskusalerma, line 108: how about doing this:

```
if self.process_type == "exporter...":
  entity = "exporter"
else:
  entity = "importer"

```  
and then this collapses down to just one admin_actions.append

oskusalerma (changes requested):

oskusalerma (approved):

PR #431, opened 2021-06-14
ICMSLST-554 Add verified certificates to sil applications

kevinetienne:
Code change was done in 3 parts:
 - remove `VerifiedCertificate` model which is replaced by a `m2m`
 - move verified certificate from oil application to fa
 - add verified certificate to sil


oskusalerma (approved):

web/domains/case/_import/fa/views.py
```diff
@@ -542,6 +550,102 @@ def archive_certificate(
...
+        application.importer.firearms_authorities.active(), pk=authority_pk
+    )
+
+    return view_application_file(
+        request.user, application, firearms_authority.files, document_pk, "import"
+    )
+
+
+@login_required
+@permission_required("web.importer_access", raise_exception=True)
```
MattHolmes123, line 620: This isn't needed as it's checked in `check_application_permission`

MattHolmes123 (changes requested):
Added one trivial comment :+1:

MattHolmes123 (approved):

PR #435, opened 2021-06-15
ICMSLST-717 Add response preparation view for FA-SIL applications

MattHolmes123:
Changes:
      - Add FA-SIL response preparation screen.
      - Save endorcement when saving FA-SIL form.
      - Add view to edit FA-SIL Goods.
      - Add ability to set the cover letter.

### DIFF
--- a/web/domains/case/_import/fa_sil/forms.py
+++ b/web/domains/case/_import/fa_sil/forms.py
@@ -6,6 +6,7 @@ from guardian.shortcuts import get_users_with_perms
 from web.domains.case._import.forms import ChecklistBaseForm
 from web.domains.country.models import Country
 from web.domains.firearms.models import ObsoleteCalibre
+from web.domains.template.models import Template
 from web.domains.user.models import User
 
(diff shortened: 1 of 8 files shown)


web/domains/case/_import/fa_sil/forms.py
```diff
@@ -536,3 +537,65 @@ def __init__(self, *args, **kwargs):
 
         for f in self.fields:
             self.fields[f].required = False
+
+
+class ResponsePrepBaseForm(forms.ModelForm):
+    """Base class form for editing description and quantity in the response preparation screen"""
+
+    class Meta:
+        abstract = True
```
kevinetienne, line 546: I might be wrong but I think `abstract` is for django `models`.

MattHolmes123: Yep my bad :)

kevinetienne (approved):
LGTM

oskusalerma (approved):

PR #443, opened 2021-06-18
ICMSLST-739 Filter OPT Compensating products commodity codes correctly.

MattHolmes123:
Changes:
  - Added widget to filter commodities by correct group & category.
  - Added some JS to clear the form when a category changes.
  - Updated form clean method to ensure no invalid commodities are saved.

### DIFF
--- a/web/domains/case/_import/opt/forms.py
+++ b/web/domains/case/_import/opt/forms.py
@@ -92,12 +92,10 @@ class CompensatingProductsOPTForm(forms.ModelForm):
             commoditygroup__group_code=category
         ).values_list("commodity_code", flat=True)
 
-        for commodity in user_commodities:
-            if commodity not in valid_commodities:
-                self.add_error(
-                    "cp_commodities", f"Invalid commodity code selected for category {category}"
-                )
-                break
+        if any(commodity not in valid_commodities for commodity in user_commodities):
+            self.add_error(
+                "cp_commodities", f"Invalid commodity code selected for category {category}"
+            )
 
         return cleaned_data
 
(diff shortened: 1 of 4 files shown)


web/domains/case/_import/opt/forms.py
```diff
@@ -67,18 +68,39 @@ class Meta:
...
+        valid_commodities = OptCompensatingProductsCommodityWidget.queryset.filter(
+            commoditygroup__group_code=category
+        ).values_list("commodity_code", flat=True)
+
+        for commodity in user_commodities:
+            if commodity not in valid_commodities:
+                self.add_error(
+                    "cp_commodities", f"Invalid commodity code selected for category {category}"
+                )
+                break
```
oskusalerma, line 100: empty line before control flow statements please

web/domains/case/_import/opt/forms.py
```diff
@@ -67,18 +68,39 @@ class Meta:
...
+        if not commodities or not category:
+            return cleaned_data
+
+        user_commodities = commodities.values_list("commodity_code", flat=True)
+
+        valid_commodities = OptCompensatingProductsCommodityWidget.queryset.filter(
+            commoditygroup__group_code=category
+        ).values_list("commodity_code", flat=True)
+
+        for commodity in user_commodities:
```
oskusalerma, line 95: would this work:

```
if any(commodity not in valid_commodities for commodity in user_commodities):
  self.add_error(blaa)

```

no need for break/loop

oskusalerma (changes requested):

web/domains/case/_import/opt/forms.py
MattHolmes123: Agreed

oskusalerma (approved):

PR #448, opened 2021-06-23
   ICMSLST-721 Creater proper case references.

oskusalerma:
ICMSLST-721 Creater proper case references.
    
    In more detail:
    
     * Add ICMSMiddleware that adds request.icms.
     * ...which contains LockManager.
     * ...which allows one to obtain deadlock-free sets of locks on DB tables.
     * ...which our new function, allocate_case_reference, uses to allocate
       new case references in a safe manner.

    ICMSLST-721 Add ApplicationBase.submit_application.
    
    Switch everyone over to it.

    ICMSLST-721 Change import/export/access status to be TextChoices.
    
    Also for import/export, move these up one level to ApplicationBase.
    
     * status / submit_datetime / reference fields
     * Statuses

    ICMSLST-721 Add "add_dummy_application" command.
    
    Useful for development.

    ICMSLST-721 Fix access request code not setting last_updated_by.
    
    It was typoing it as last_update_by.


web/models/models.py
```diff
@@ -0,0 +1,17 @@
...
+        ]
+
+    # e.g. "IMA" for import applications
+    prefix = models.CharField(max_length=8)
+
+    # this is null for importer/exporter access requests
+    year = models.IntegerField(null=True)
+
+    # incrementing number for each prefix/year combination, starting from 1
+    reference = models.IntegerField()
```
kevinetienne, line 17: Should `year` and `reference` be a `PositiveIntegerField`?

kevinetienne (approved):
Feel free to update the model fields if you think it's necessary. It feels the year and the count would be set in the utility function anyway.

MattHolmes123 (approved):

PR #456, opened 2021-06-29
ICMSLST-634 Fix/work around mypy issues found by django-stubs.

oskusalerma:
This is phase one, still 150+ to go.

In more detail:

 * Upgrade mypy, add types-pytz and types-requests.

 * Fix is_active mypy issues.

 * Ignore attr-defined error for status field.

 * Ignore several attributes that aren't defined in base class.

 * Import reverse from django.urls, not django.shortcuts.

 * Type optional forms.

 * Fix type complaints in mailshot/views.py.

 * Make all File fields non-nullable.

   They do not need to be nullable anymore, we only add valid files to the
   database nowadays.

   Marking them non-nullable removes mypy complaints about Optional[x].

 * Add user_is_contact_of_org method.

 * Type applications with a checklist.

 * Disable warnings for reverse one_to_one relationships.

 * Ignore mypy warnings regarding class type.

 * Fix related_file_manager type.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -7,3 +7,10 @@ check_untyped_defs = True
 
 ; [mypy-web.*]
 ; disallow_untyped_defs = True
+
+# ICMSLST-634 Note: Even if you comment out these sections you will see error if
+# you have the package installed in your venv.
+# plugins = mypy_django_plugin.main
+
+# [mypy.plugins.django-stubs]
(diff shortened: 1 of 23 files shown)


MattHolmes123 (approved):
Remove comment about is_active being fixed

kevinetienne (approved):

PR #458, opened 2021-06-30
ICMSLST-654 Add missing validation to forms.

oskusalerma:
Also:

ICMSLST-654 Comment out selenium/chrome in docker-compose.yml.

We are not running the BDD tests at the moment, downloading/running these
is just wasting time and resources.

### DIFF
--- a/web/domains/case/_import/wood/forms.py
+++ b/web/domains/case/_import/wood/forms.py
@@ -81,39 +81,23 @@ class SupportingDocumentForm(forms.Form):
     document = ICMSFileField(required=True)
 
 
-class AddContractDocumentForm(forms.Form):
+class AddContractDocumentForm(forms.ModelForm):
     document = ICMSFileField(required=True)
-
-    reference = forms.CharField(
-        help_text="Enter the reference number of the contract/pre-contract between the importer and exporter.",
-        required=True,
-    )
-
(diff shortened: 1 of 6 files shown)


web/domains/case/_import/wood/forms.py
```diff
@@ -81,39 +81,23 @@ class SupportingDocumentForm(forms.Form):
...
+    class Meta:
+        model = models.WoodContractFile
+        fields = ("reference", "contract_date")
 
-    contract_date = forms.DateField(
-        help_text="Enter the date of the contract/pre-contract between the importer and exporter.",
-        required=True,
-        widget=DateInput(),
-    )
+        widgets = {"contract_date": DateInput()}
```
MattHolmes123, line 91: https://docs.djangoproject.com/en/3.2/ref/models/fields/#datefield
I Don't think you need this (same for all the forms you've done this in)

MattHolmes123 (changes requested):

MattHolmes123 (approved):

kevinetienne (approved):

web/domains/case/_import/wood/forms.py
oskusalerma: incorrect sir!

PR #463, opened 2021-07-02
ICMSLST-762 Add "supporting documents" to Textiles application.

oskusalerma:


web/domains/case/_import/textiles/views.py
```diff
@@ -36,12 +39,85 @@ def edit_textiles(request: AuthenticatedHttpRequest, *, application_pk: int) ->
...
+            TextilesApplication.objects.select_for_update(), pk=application_pk
+        )
+
+        check_application_permission(application, request.user, "import")
+
+        task = application.get_task(ImportApplication.Statuses.IN_PROGRESS, "prepare")
+
+        if request.POST:
+            form = DocumentForm(data=request.POST, files=request.FILES)
+            document = request.FILES.get("document")
```
kevinetienne, line 69: can we get the document from the form cleaned data as we're passing it to the form, eg:
```python
            form = DocumentForm(data=request.POST, files=request.FILES)

            if form.is_valid():
                document = form.cleaned_data.get("document")
                create_file_model(document, request.user, application.supporting_documents)
```

MattHolmes123 (approved):
Approved (assuming you change kevins comment)

PR #468, opened 2021-07-05
ICMSLST-761 Add Textiles (Quota) view

mikeylockwood:
View the details of a submitted Textile application

### DIFF
--- a/web/templates/web/domains/case/import/textiles/view.html
+++ b/web/templates/web/domains/case/import/textiles/view.html
@@ -1,0 +1,26 @@
+{% extends "web/domains/case/import/view_case.html" %}
+
+{% block main_content %}
+  {% call application_section("Textiles (Quota)") %}
+    {{ application_field(process.applicant_reference, "Applicant's Reference", "N/a") }}
+    {{ application_field("Yes" if process.goods_cleared else "No", "Will the goods be cleared in another Member State of the European Union?") }}
+    {{ application_field(process.shipping_year, "Shipping Year") }}
+    {{ application_field(process.origin_country, "Country Of Origin") }}
+    {{ application_field(process.consignment_country, "Country Of Consignment") }}
+  {% endcall %}
+
+  {% call application_section("Goods") %}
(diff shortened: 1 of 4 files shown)


web/templates/web/domains/case/import/textiles/view.html
```diff
@@ -0,0 +1,26 @@
+{% extends "web/domains/case/import/view_case.html" %}
+
+{% block main_content %}
+  {% call application_section("Textiles (Quota)") %}
+    {{ application_field(process.applicant_reference, "Application Reference", "N/a") }}
```
mikeylockwood, line 5: Just noticed this is "Applicant's Reference" in the current system.

web/templates/web/domains/case/import/textiles/view.html
```diff
@@ -0,0 +1,26 @@
+{% extends "web/domains/case/import/view_case.html" %}
+
+{% block main_content %}
+  {% call application_section("Textiles (Quota)") %}
+    {{ application_field(process.applicant_reference, "Applicant's Reference", "N/a") }}
+    {{ application_field("Yes" if process.goods_cleared else "No", "Will the goods be cleared in another Member State of the European Union?") }}
```
mikeylockwood, line 6: Will we always want to display boolean fields as Yes / No? Would it be better to handle this in the macro?

MattHolmes123: The macro was only added to avoid adding this for every field:  

```
{% macro application_field(field, label, default="") -%}
  <div class="row">
    <div class="three columns">
      <label class="prompt west">{{ label }}</label>
    </div>
    <div class="six columns">
      <span class="text-widget">{{ field or default }}</span>
    </div>
    <div class="three columns"></div>
  </div>
{%- endmacro %}
```

I would say its fine to leave it how you have it.  
(I didn't know you could do inline if's in jinja so `+1` for that :)

oskusalerma (approved):

PR #470, opened 2021-07-06
ICMSLST-782 Allow agent to create and edit applications

kevinetienne:
Update `check_application_permission` to verify if the agent has access to the application (import/export).

This PR adds an agent in `add_dummy_data` management command. The agent can create/edit/submit applications.

Next PRs will:
  - allow case worker to add agents with the correct permissions
  - restrict the contact list on the application


web/domains/case/_import/sanctions/views.py
```diff
@@ -206,32 +217,30 @@ def _edit_goods(
 
 
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
 @require_POST
-def delete_goods(request, application_pk, goods_pk):
+def delete_goods(request: AuthenticatedHttpRequest, *, application_pk: int, goods_pk: int):
```
MattHolmes123, line 221: return type missing

web/domains/case/_import/sanctions/views.py
```diff
@@ -206,32 +217,30 @@ def _edit_goods(
...
 
         get_object_or_404(application.sanctionsandadhocapplicationgoods_set, pk=goods_pk).delete()
 
     return redirect(reverse("import:sanctions:edit", kwargs={"application_pk": application_pk}))
 
 
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
-def add_supporting_document(request, pk):
+def add_supporting_document(request: AuthenticatedHttpRequest, *, pk: int):
```
MattHolmes123, line 235: return type missing

web/domains/case/_import/sanctions/views.py
```diff
@@ -256,7 +265,9 @@ def add_supporting_document(request, pk):
 
 @require_GET
 @login_required
-def view_supporting_document(request, application_pk, document_pk):
+def view_supporting_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 270: return type missing

web/domains/case/_import/sanctions/views.py
```diff
@@ -265,17 +276,17 @@ def view_supporting_document(request, application_pk, document_pk):
 
 @require_POST
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
-def delete_supporting_document(request, application_pk, document_pk):
+def delete_supporting_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 281: return type missing

web/domains/case/_import/wood/views.py
```diff
@@ -70,15 +68,13 @@ def edit_wood_quota(request: AuthenticatedHttpRequest, *, application_pk: int) -
 
 
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
-def add_supporting_document(request, pk):
+def add_supporting_document(request: AuthenticatedHttpRequest, *, pk: int):
```
MattHolmes123, line 71: return type missing

web/domains/case/_import/wood/views.py
```diff
@@ -104,7 +100,9 @@ def add_supporting_document(request, pk):
 
 @require_GET
 @login_required
-def view_supporting_document(request, application_pk, document_pk):
+def view_supporting_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 105: return type missing

web/domains/case/_import/wood/views.py
```diff
@@ -114,17 +112,17 @@ def view_supporting_document(request, application_pk, document_pk):
 
 @require_POST
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
-def delete_supporting_document(request, application_pk, document_pk):
+def delete_supporting_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 117: return type missing

web/domains/case/_import/wood/views.py
```diff
@@ -182,25 +178,27 @@ def add_contract_document(request: AuthenticatedHttpRequest, pk: int) -> HttpRes
 
 @require_GET
 @login_required
-def view_contract_document(request, application_pk, document_pk):
+def view_contract_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 183: return type missing

web/domains/case/_import/wood/views.py
```diff
@@ -182,25 +178,27 @@ def add_contract_document(request: AuthenticatedHttpRequest, pk: int) -> HttpRes
...
     return import_views.view_file(request, application, application.contract_documents, document_pk)
 
 
 @require_POST
 @login_required
-@permission_required("web.importer_access", raise_exception=True)
-def delete_contract_document(request, application_pk, document_pk):
+def delete_contract_document(
+    request: AuthenticatedHttpRequest, *, application_pk: int, document_pk: int
+):
```
MattHolmes123, line 193: return type missing

web/management/commands/add_dummy_data.py
```diff
@@ -27,6 +27,15 @@ def handle(self, *args, **options):
         if settings.APP_ENV not in ("local", "dev"):
             raise CommandError("Can only add dummy data in 'dev' / 'local' environments!")
 
+        self.activate_import_application_types()
+
+        user = self.create_superuser(options["password"])
```
MattHolmes123, line 32: The "admin" user we create day to day should be created with User.objects.create_user.

web/management/commands/add_dummy_data.py
```diff
@@ -60,7 +75,40 @@ def handle(self, *args, **options):
...
-        # exporter
+        return user
+
+    def create_agent(self, password: str) -> User:
+        try:
+            agent = User.objects.get(username="agent")
+
+            self.stdout.write("Existing user: agent")
+        except User.DoesNotExist:
+            agent = User.objects.create(
```
MattHolmes123, line 86: If you see here I create a user using User.objects.create_user:  
`web/management/commands/add_test_data.py`  
It sets the password correctly etc

MattHolmes123 (changes requested):

web/management/commands/add_dummy_data.py
kevinetienne: Thanks, I'll copy the same logic you did for adding user when rebasing this branch

web/management/commands/add_dummy_data.py
```diff
@@ -60,7 +75,40 @@ def handle(self, *args, **options):
...
+            agent.set_password(password)
+
+            self.stdout.write("Created following user: agent")
+
+        # permissions
+        for perm in [
+            "importer_access",
+            "exporter_access",
+        ]:
+            agent.user_permissions.add(Permission.objects.get(codename=perm))
```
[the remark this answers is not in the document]
kevinetienne: `importer_access` and `exporter_access` give access in general to create/edit/submit applications. They are 'general' permissions.
`is_agent_of_importer`, `is_agent_of_exporter` are object level permissions.

If we were going to not use `importer_access` and `exporter_access` for `agent` it would mean by extension we could do the same with `contacts` on the `importer` and `exporter`. It would simplify checking creating/editing/submitting applications but it wouldn't allow us to revoke access in general.

If we were going to not use `importer_access` and `exporter_access` for `agent` and use them with `contacts` on the `importer` and `exporter it would means we would have a slightly more complicated check on who can create/edit/submit applications, so maybe it's not worth it.

Hope that helps!

web/tests/domains/case/export/test_views.py
```diff
@@ -119,6 +119,7 @@ def test_edit_no_auth(self):
     def test_no_task(self):
```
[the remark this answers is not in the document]
kevinetienne: I'll add a few thanks

web/management/commands/add_dummy_data.py
kevinetienne: I'll check what we have with the rebase and I will use `create_user` good point, thanks

web/tests/domains/case/test_views.py
```diff
@@ -28,11 +28,21 @@ def imp_application_user():
     return ActiveUserFactory.create(permission_codenames=["importer_access"])
 
 
+@pytest.fixture
+def imp_agent_application_user():
```
MattHolmes123, line 32: I think all these fixtures should be replaced by the newer ones here:  
https://github.com/uktrade/icms/blob/develop/web/tests/conftest.py#L41-L73.
Also your new fixtures should do the data setup here:  
https://github.com/uktrade/icms/blob/develop/web/management/commands/add_test_data.py#L17

The ones you added are copying my bad example where I'm creating users / importers & offices records every test.
If we create the data in add_test_data.py and then create fixtures that only fetch data then it should be quicker.

MattHolmes123: These tests use the better fixtures:  
https://github.com/uktrade/icms/blob/develop/web/tests/domains/case/_import/fa_dfl/test_views.py

MattHolmes123 (changes requested):

web/tests/domains/case/test_views.py
kevinetienne: fixed in https://github.com/uktrade/icms/pull/470/commits/e277edead7a20a354684b86f3b1e7d3370b7e061 and in https://github.com/uktrade/icms/pull/470/commits/14b5aae757243f245d51bb272d2e20bf569087bf. I've followed your recommendation to not stuff everything (the applications for importer/exporter/agent) in test data.

MattHolmes123 (approved):

PR #471, opened 2021-07-06
ICMSLST-758 and ICMSLST-759 Add Manage Checklist and Authorisation views to textiles application.

MattHolmes123:


web/domains/case/_import/textiles/views.py
```diff
@@ -201,3 +205,47 @@ def delete_document(
         document.save()
 
         return redirect(reverse("import:textiles:edit", kwargs={"application_pk": application_pk}))
+
+
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def manage_checklist(request: AuthenticatedHttpRequest, *, application_pk):
```
oskusalerma, line 212: type application_pk, add return type

oskusalerma (changes requested):

oskusalerma (approved):

PR #472, opened 2021-07-06
ICMSLST-757 Add Textiles (Quota) response preparation view

mikeylockwood:
* Add view for response preparation for textiles
* Include link to edit goods using `GoodsTextilesLicenceForm`

### DIFF
--- a/web/domains/case/_import/textiles/views.py
+++ b/web/domains/case/_import/textiles/views.py
@@ -17,6 +17,7 @@ from web.utils.validation import ApplicationErrors, PageErrors, create_page_erro
 
 from .forms import (
     EditTextilesForm,
+    GoodsTextilesLicenceForm,
     TextilesChecklistForm,
     TextilesChecklistOptionalForm,
 )
@@ -110,6 +111,15 @@ def submit_textiles(request: AuthenticatedHttpRequest, *, application_pk: int) -
             form = SubmitForm(data=request.POST)
 
             if form.is_valid() and not errors.has_errors():
+                endorsement = Template.objects.get(
+                    template_name="Endorsement 1 (must be updated each year)",
+                    template_type=Template.ENDORSEMENT,
+                    application_domain="IMA",
+                )
+                application.endorsements.create(content=endorsement.template_content)
+                application.category_licence_description = (
+                    application.category_commodity_group.group_description
+                )
                 application.submit_application(request, task)
 
                 return redirect(reverse("workbasket"))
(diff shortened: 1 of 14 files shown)


web/templates/web/domains/case/import/manage/prepare-response.html
```diff
@@ -3,6 +3,21 @@
...
+    <div class="three columns">
+      <label class="prompt west">{{ label }}</label>
+    </div>
+    <div class="six columns">
+      <span class="text-widget">{{ field or default }}</span>
+    </div>
+    <div class="three columns"></div>
+  </div>
+{%- endmacro %}
+{# END MACROS #}
```
kevinetienne, line 18: Should we make the other [`application_field`](https://github.com/uktrade/icms/blob/bbb2ffb1cff786fbbc17fb200d65525dff670b5a/web/templates/web/domains/case/import/view_case.html#L17-L28) available more globally?

web/domains/case/_import/textiles/views.py
```diff
@@ -249,3 +262,44 @@ def manage_checklist(request: AuthenticatedHttpRequest, *, application_pk: int)
             template_name="web/domains/case/import/management/checklist.html",
             context=context,
         )
+
+
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def edit_goods_licence(request: AuthenticatedHttpRequest, application_pk: int) -> HttpResponse:
```
MattHolmes123, line 269: We have an unwritten rule about view arguments being keyword only (after the request) argument.

web/templates/web/domains/case/import/manage/prepare-response.html
MattHolmes123: We have form macros here:  
`icms/web/templates/forms/fields.html`  
What about a another file for displaying fields when viewing the application / screens like prepare-response? e.g.  
`icms/web/templates/display/fields.html`  
And then add application_section and application_field to there.

web/tests/domains/case/_import/textiles/test_admin_views.py
```diff
@@ -0,0 +1,83 @@
+import re
+
+import pytest
+from django.urls import reverse
+from pytest_django.asserts import assertRedirects
+
+from web.domains.case._import.textiles.models import TextilesApplication
+
```
MattHolmes123, line 8: Your tests are failing, I will review them when they are passing :+1:

MattHolmes123 (changes requested):

web/templates/web/domains/case/import/manage/prepare-response.html
mikeylockwood: Yes that sounds sensible, will add

web/domains/case/_import/textiles/views.py
mikeylockwood: I haven't seen this in any of the other import application views. Happy to change it, but we'd want all the views to be consistent

MattHolmes123: image

MattHolmes123: @mikeylockwood All of these are keyword only arguments after the request parameter.

oskusalerma (approved):

MattHolmes123 (approved):

PR #473, opened 2021-07-07
ICMSLST-786 Redirect all application types to workbasket after submit.

MattHolmes123:


web/domains/case/models.py
```diff
@@ -402,3 +404,12 @@ def submit_application(self, request: AuthenticatedHttpRequest, task: Task) -> N
         task.save()
 
         Task.objects.create(process=self, task_type="process", previous=task)
+
+    def redirect_after_submit(self, request: AuthenticatedHttpRequest):
```
oskusalerma, line 408: return tpe missing

oskusalerma (changes requested):

oskusalerma (approved):

PR #476, opened 2021-07-07
ICMSLST-789 Add querysets to show export applications in workbasket

kevinetienne:


web/domains/case/export/models.py
```diff
@@ -143,7 +143,7 @@ def user_is_contact_of_org(self, user: User) -> bool:
     def get_workbasket_subject(self) -> str:
         return "\n".join(
             [
-                "Certificate Application",
+                "Export Application",
```
oskusalerma, line 146: don't change that, i took great care looking at what the legacy system calls these things

oskusalerma (changes requested):

web/domains/case/export/models.py
kevinetienne: updated in https://github.com/uktrade/icms/pull/476/commits/19c94bffdf68d9075a70f5bd47d26a3cdd2af79b

oskusalerma (approved):

kevinetienne:
Tests pass locally, merging as circleci is having problem

image

PR #477, opened 2021-07-07
ICMSLST-785 Add config for S3 bucket

mikeylockwood:


config/settings/base.py
```diff
@@ -159,11 +159,19 @@
 }
 
 # for https://github.com/uktrade/django-chunk-s3-av-upload-handlers
-AWS_REGION = "eu-west-2"
-AWS_ACCESS_KEY_ID = "dev"
-AWS_SECRET_ACCESS_KEY = "bar"
-AWS_STORAGE_BUCKET_NAME = "icms.local"
-AWS_S3_ENDPOINT_URL = "http://localstack:4572/"
+if "aws-s3-bucket" in VCAP_SERVICES:
```
oskusalerma, line 162: can you instead move the hard-coded localstack version to non_prod_base.py and the vcap-reading one to production.py, and remove these completely from base.py

oskusalerma (changes requested):

config/settings/production.py
```diff
@@ -9,6 +9,8 @@
 
 env = environ.Env()
 
+VCAP_SERVICES = env.json("VCAP_SERVICES", {})
```
oskusalerma, line 12: this shouldn't have a default, it should fall over if it doesn't exist, the application won't work otherwise

config/settings/production.py
```diff
@@ -34,6 +36,16 @@
 ILB_CONTACT_PHONE = env.str("ICMS_ILB_CONTACT_PHONE")
 ICMS_FIREARMS_HOMEOFFICE_EMAIL = env.str("ICMS_FIREARMS_HOMEOFFICE_EMAIL")
 
+# for https://github.com/uktrade/django-chunk-s3-av-upload-handlers
+S3_ROOT_DIRECTORY = "documents"
```
oskusalerma, line 40: this is common, can't this be in base.py?

config/settings/production.py
```diff
@@ -34,6 +36,16 @@
...
+
+if "aws-s3-bucket" in VCAP_SERVICES:
```
oskusalerma, line 42: again, this is not optional, the application won't work if this is missing, so it should fall over so we're immediately aware. take out the if.

oskusalerma (changes requested):

config/settings/base.py
```diff
@@ -158,12 +158,7 @@
     "required": "You must enter this item",
 }
 
-# for https://github.com/uktrade/django-chunk-s3-av-upload-handlers
-AWS_REGION = "eu-west-2"
-AWS_ACCESS_KEY_ID = "dev"
-AWS_SECRET_ACCESS_KEY = "bar"
-AWS_STORAGE_BUCKET_NAME = "icms.local"
-AWS_S3_ENDPOINT_URL = "http://localstack:4572/"
+# Root directory for s3 bucekt
```
oskusalerma, line 161: typo

oskusalerma (changes requested):

oskusalerma (approved):

PR #480, opened 2021-07-09
ICMSLST-606 Add Prior Surveillance goods section.

MattHolmes123:


web/domains/case/_import/sps/utils.py
```diff
@@ -0,0 +1,10 @@
+# TODO: Revisit this when we know how to calculate it.
```
oskusalerma, line 1: these functions have nothing to do with sps...how about moving them to web/utils/currency.py?

can write some simple tests for them as well

web/domains/case/_import/sps/models.py
```diff
@@ -15,7 +16,41 @@ class Type(models.TextChoices):
 class PriorSurveillanceApplication(ImportApplication):
     PROCESS_TYPE = ImportApplicationType.ProcessTypes.SPS
 
-    # TODO: add other fields
```
oskusalerma, line 18: why are you deleting this comment, my ticket 605 has more fields to add?

web/domains/case/_import/sps/models.py
```diff
@@ -15,7 +16,41 @@ class Type(models.TextChoices):
...
+            " commodity code in this box is correct. If you are unsure of the"
+            " correct commodity code, consult the HM Revenue & Customs at"
+            " classification.enquiries@hmrc.gsi.gov.uk or use the online trade"
+            " tariff https://www.gov.uk/trade-tariff/sections."
+        ),
+    )
+
+    quantity = models.PositiveIntegerField(
+        null=True,
+        verbose_name="Quantity",
```
oskusalerma, line 36: why do you need that, django will create exactly that?

web/domains/case/_import/sps/views.py
```diff
@@ -43,7 +44,10 @@ def edit_sps(request: AuthenticatedHttpRequest, *, application_pk: int) -> HttpR
             form = EditSPSForm(data=request.POST, instance=application)
 
             if form.is_valid():
-                form.save()
+                instance: PriorSurveillanceApplication = form.save()
```
oskusalerma, line 47: don't you want to say `form.save(commit=False)` since you're modifying the object after?

oskusalerma (changes requested):

web/domains/case/_import/sps/views.py
MattHolmes123: You do when you don't forget about commit=False :)

oskusalerma (approved):

PR #489, opened 2021-07-14
ICMSLST-677 Fix note UI annoyances.

oskusalerma:
In more detail:

  * Disallow editing/archiving of archived notes
  * Fix note.status checking for DELETED, not COMPLETED
    * This achieves the effect of only allowing archiving notes with
      status=DRAFT.
  * Remove archive/unarchive buttons from note editing screen.
  * Fix unarchive button having wrong icon.


MattHolmes123 (approved):

kevinetienne (approved):
We can use the following but I think it doesn't matter too much:

```
  {{ csrf_input }}
```
instead of

```
<input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}" />
```

PR #494, opened 2021-07-16
ICMSLST-849 Update related commodity search and edit functionality

MattHolmes123:


web/domains/commodity/models.py
```diff
@@ -83,8 +83,6 @@ def __str__(self):
 
 
 class Commodity(Archivable, models.Model):
-    LABEL = "Commodity"
```
MattHolmes123, line 86: Note to self: 
module level COMMODITY_TYPES dictionary can be deleted from this file.

kevinetienne (approved):

PR #499, opened 2021-07-20
ICMSLST-783 Contacts for importers, exporters and agents

kevinetienne:
Remove regex in url for entity (individual/organisation) and replace with converter.
Refactor the code to add and delete contacts in a new module called contacts.
Fix who can be displayed in the contact form (all active users as opposed to user with a set of global permissions).
Add two TODOs to remove Imporer.user fk and to add agent on form to create applications.


web/domains/contacts/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.urls import path
+
+from . import views
+
+app_name = "contacts"
+
+urlpatterns = [
+    path("<actortype:actor_type>/<int:actor_pk>/add/", views.add, name="add"),
+    path(
+        "<actortype:actor_type>/<int:actor_pk>/<int:contact_pk>/delete/",
```
MattHolmes123, line 10: Should we group this bit `<actortype:actor_type>/<int:actor_pk>/` like we do elsewhere?

web/domains/contacts/views.py
```diff
@@ -0,0 +1,120 @@
...
+from django.views.decorators.http import require_POST
+from guardian.shortcuts import assign_perm, get_users_with_perms, remove_perm
+
+from web.domains.user.models import User
+from web.models import Exporter, Importer
+from web.types import AuthenticatedHttpRequest
+
+from .forms import ContactForm
+
+ImpOrExp = Union[Importer, Exporter]
```
MattHolmes123, line 16: Elsewhere we use ImpOrExp to mean an import or export application.
I think its confusing to have a type with the same name mean something different.
I think we should refer to Importer & Exporter as Organisation like we have done elsewhere.

The same applies to `ImpOrExpT` below

web/converters.py
```diff
@@ -52,3 +52,31 @@ def to_url(self, value):
...
+
+    def to_url(self, value):
+
+        if value not in ["individual", "organisation"]:
+            raise ValueError
+
+        return value
+
+
+class ActorTypeConverter:
```
MattHolmes123, line 71: I think Actor should be replaced with Org or Organisation.  
in `web/domains/case/models.py` we have several `def user_is_XXX_of_org()` methods.
I was pretty confused by what actor meant initially.

web/domains/contacts/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.urls import path
+
+from . import views
+
+app_name = "contacts"
+
+urlpatterns = [
+    path("<actortype:actor_type>/<int:actor_pk>/add/", views.add, name="add"),
```
MattHolmes123, line 8: I think `org_type` and `org_pk` are more accurate variable names to use throughout.

web/domains/contacts/views.py
```diff
@@ -0,0 +1,120 @@
...
+        raise NotImplementedError(f"Unknown case_type {actor_type}")
+
+
+# TODO: ICMSLST-859 ICMSLST-860 actor's contacts can view/edit actor details
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+@require_POST
+def add(request: AuthenticatedHttpRequest, *, actor_type: str, actor_pk: int):
+    model_class = _get_class_imp_or_exp(actor_type)
+    actor: ImpOrExp = get_object_or_404(model_class, pk=actor_pk)
```
MattHolmes123, line 35: The last time I will mention `actor` -> `org` :smile: ... You get the idea

web/domains/contacts/views.py
```diff
@@ -0,0 +1,120 @@
...
+
+    contacts: ImpOrExp = current_contacts(actor)
```
MattHolmes123, line 37: This isn't the correct type.
I would do the following and type it the django-stubs way.

```
# At the top of the file:
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from django.db.models import QuerySet
    
# create a type for the returned users.
contacts: "QuerySet[User]" = current_contacts(actor)
```

See point `2` here: https://github.com/typeddjango/django-stubs#i-cannot-use-queryset-or-manager-with-type-annotations

web/domains/contacts/views.py
```diff
@@ -0,0 +1,120 @@
...
+    delete_contact_perm(actor, contact)
+
+    if actor.is_agent():
+        return redirect(reverse(f"{actor_type}-agent-edit", kwargs={"pk": actor.pk}))
+
+    else:
+        return redirect(reverse(f"{actor_type}-edit", kwargs={"pk": actor.pk}))
+
+
+def current_contacts(actor: ImpOrExp) -> User:
```
MattHolmes123, line 71: the return type here should be a QuerySet as discussed above.

web/domains/contacts/views.py
```diff
@@ -0,0 +1,120 @@
...
+    """Current active contacts associated with importer/exporter/agent."""
+    if isinstance(actor, Importer):
+        return get_users_with_perms(actor, only_with_perms_in=["is_contact_of_importer"]).filter(
+            user_permissions__codename="importer_access"
+        )
+
+    elif isinstance(actor, Exporter):
+        return get_users_with_perms(actor, only_with_perms_in=["is_contact_of_exporter"]).filter(
+            user_permissions__codename="exporter_access"
+        )
```
MattHolmes123, line 81: Shall we have an else that throws an exception?  
Same question to all the assign / remove permission functions.

web/domains/importer/views.py
```diff
@@ -481,43 +449,33 @@ def edit_agent(request, pk):
 @login_required
 @permission_required("web.reference_data_access", raise_exception=True)
 @require_POST
-def archive_agent(self, pk):
+def archive_agent(self, *, pk: int):
```
MattHolmes123, line 452: Why don't `archive_agent` and `unarchive_agent` have a request?  
Also neither of them define a return type

MattHolmes123 (changes requested):

web/domains/importer/views.py
kevinetienne: wow this is a bit strange and it's working! I'll add the request though

web/domains/contacts/views.py
kevinetienne: I was thinking as we're checking if we have an importer or an exporter type we might not need it. But happy to add it.

web/domains/contacts/views.py
kevinetienne: Not sure why I didn't get an error here.

web/domains/importer/views.py
kevinetienne: we have a `self` there...

web/converters.py
kevinetienne: good call I struggled a bit with the naming and found the term `actor` in legacy docs

web/domains/contacts/views.py
kevinetienne: I've used `Org` and `OrgT` let me know if that was what you mean?

web/domains/contacts/views.py
MattHolmes123: I don't think the underlying functions being called are typed either so it probably assumes Any

web/domains/importer/views.py
MattHolmes123: WOW :D How did I not see that... I thought it looked strange.

MattHolmes123 (approved):
Reviewed and left a few comments for Kevin to sort.

PR #502, opened 2021-07-22
ICMSLST-850 & ICMSLST-863 Changes

MattHolmes123:


web/domains/commodity/models.py
```diff
@@ -24,16 +24,19 @@ def __str__(self):
 
 class Commodity(Archivable, models.Model):
     is_active = models.BooleanField(blank=False, null=False, default=True)
-    start_datetime = models.DateTimeField(auto_now_add=True, blank=False, null=False)
-    end_datetime = models.DateTimeField(blank=True, null=True)
     commodity_code = models.CharField(max_length=10, blank=False, null=False)
-    validity_start_date = models.DateField(blank=False, null=True)
+    validity_start_date = models.DateField(blank=False, null=False)
```
kevinetienne, line 28: I think you can remove `blank=False, null=False` as they might be the default?

web/domains/commodity/models.py
```diff
@@ -24,16 +24,19 @@ def __str__(self):
...
     validity_end_date = models.DateField(blank=True, null=True)
     quantity_threshold = models.IntegerField(blank=True, null=True)
     sigl_product_type = models.CharField(max_length=3, blank=True, null=True)
 
     commodity_type = models.ForeignKey(CommodityType, on_delete=models.PROTECT)
 
+    # These are to be ignored, the start_datetime is simply a timestamp.
+    start_datetime = models.DateTimeField(auto_now_add=True, blank=False, null=False)
```
kevinetienne, line 36: I think you can remove blank=False, null=False as they might be the default?

web/utils/commodity.py
```diff
@@ -0,0 +1,54 @@
...
+def get_usage_countries(application_usage: "QuerySet[Usage]") -> "QuerySet[Country]":
+    """Return countries linked to the usage records."""
+    return Country.objects.filter(usage__in=application_usage).distinct()
+
+
+def get_usage_commodities(application_usage: "QuerySet[Usage]") -> "QuerySet[Commodity]":
+    """Return commodities linked to the usage records."""
+    return Commodity.objects.filter(
+        commoditygroup__in=CommodityGroup.objects.filter(usages__in=application_usage)
+    ).distinct()
```
kevinetienne, line 54: Should we add `.filter(is_active=True)`?

web/utils/commodity.py
```diff
@@ -0,0 +1,54 @@
...
+        # Filtering the manytomany commodities causes duplicate records
+        .distinct()
+    )
+
+    return application_usage
+
+
+def get_usage_countries(application_usage: "QuerySet[Usage]") -> "QuerySet[Country]":
+    """Return countries linked to the usage records."""
+    return Country.objects.filter(usage__in=application_usage).distinct()
```
kevinetienne, line 47: Should we add `.filter(is_active=True)`?

web/domains/case/_import/derogations/widgets.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def get_queryset(self) -> "QuerySet[Commodity]":
+        return get_usage_commodities(
+            get_usage_records(ImportApplicationType.Types.DEROGATION)  # type: ignore[arg-type]
+        )
+
+    def filter_queryset(
+        self, request: HttpRequest, term: str, queryset: QuerySet = None, **dependent_fields
+    ) -> "QuerySet[Commodity]":
+        """Filter the available countries depending on the application type selected by the user."""
```
kevinetienne, line 66: Is it dependent on `Derogation`?

kevinetienne: Should it be:

```
Filter commodities with the available countries related to Derogation type.
```
```

File: web/domains/case/_import/derogations/widgets.py [hunk 6]
```diff
@@ -0,0 +1,82 @@
...
+
+        return attrs
+
+    def get_queryset(self) -> "QuerySet[Commodity]":
+        return get_usage_commodities(
+            get_usage_records(ImportApplicationType.Types.DEROGATION)  # type: ignore[arg-type]
+        )
+
+    def filter_queryset(
+        self, request: HttpRequest, term: str, queryset: QuerySet = None, **dependent_fields
```
kevinetienne, line 64: I guess you've checked it was not an authenticated request? Does `django-select2` re-use the cookies?

web/domains/commodity/widgets.py
```diff
@@ -20,3 +27,70 @@ class CommodityGroupCommodityWidget(CommodityWidget):
     # Key is a name of a field in a form.
     # Value is a name of a field in a model (used in `queryset`).
     dependent_fields = {"commodity_type": "commodity_type"}
+
+
+class UsageCountryWidget(s2forms.ModelSelect2Widget):
+    """Used when creating usage records"""
+
+    queryset = Country.objects.all()
```
kevinetienne, line 35: Should we filter out the non active countries? It's probably going to be used by the ilb admin so I don't know?

web/domains/case/_import/derogations/widgets.py
MattHolmes123: As discussed its not authenticated and we have a ticket to sort all of them :+1:

kevinetienne: We should create our own view to serve data for `django-select2`: https://django-select2.readthedocs.io/en/latest/django_select2.html#security-authentication

kevinetienne (approved):

PR #505, opened 2021-07-26
ICMSLST-874 Add CFS schedule endpoints to add, edit and delete schedules.

MattHolmes123:


web/domains/case/export/views.py
```diff
@@ -198,16 +204,73 @@ def submit_com(request: AuthenticatedHttpRequest, *, application_pk: int) -> Htt
...
+            "process": application,
+            "task": task,
+            "form": form,
+            "schedules": schedules,
+        }
+
+        return render(request, "web/domains/case/export/edit-cfs.html", context)
+
+
+@login_required
```
kevinetienne, line 238: it looks like we need `@require_POST`?

web/templates/web/domains/case/export/edit-cfs.html
```diff
@@ -30,11 +30,54 @@ <h4>Application details</h4>
...
+        </td>
+      </tr>
+    {% endfor %}
+    </tbody>
+  </table>
+  <hr>
+  <a href="{{ icms_url('export:cfs-schedule-add', kwargs={'application_pk': process.pk}) }}"
+     class="button small-button icon-plus">
+    Add Schedule
+  </a>
```
kevinetienne, line 82: Should be wrapped in a post.

MattHolmes123: yep agreed

web/domains/case/export/views.py
MattHolmes123: yep

kevinetienne (approved):

PR #508, opened 2021-07-26
Add migration for populating product legislation data

mikeylockwood:


web/migrations/0014_add_product_legislation_data.py
```diff
@@ -0,0 +1,197 @@
...
+        False,
+    ),
+]
+
+
+def add_import_data(apps, schema_editor):
+    ProductLegislation = apps.get_model("web", "ProductLegislation")
+    ProductLegislation.objects.bulk_create(
+        [
+            ProductLegislation(**dict(zip(PRODUCT_LEGISLATION_FIELDS, pl_data)))
```
MattHolmes123, line 183: Do we need to set the `id` column?

mikeylockwood: It's not necessary, I was just matching the ids with the id in the source data. Could just remove it and have it auto increment

MattHolmes123 (approved):
Approved, I've mentioned the docstring for Michael to change.

PR #514, opened 2021-07-29
ICMSLST-881 Add edit screen for Iron and Steel.

oskusalerma:


web/domains/case/_import/ironsteel/forms.py
```diff
@@ -0,0 +1,63 @@
...
+            ("false", "No"),
+        ]
+
+        self.fields["origin_country"].queryset = Country.objects.filter(
+            country_groups__name="Iron and Steel (Quota) COOs"
+        )
+
+        self.fields["consignment_country"].queryset = Country.objects.filter(
+            country_groups__name="Iron and Steel (Quota) COCs"
+        )
```
kevinetienne, line 63: I think we should filter out the non active one

```
      self.fields["origin_country"].queryset = Country.objects.filter(
            country_groups__name="Iron and Steel (Quota) COOs"
        ).filter(is_active=True)

        self.fields["consignment_country"].queryset = Country.objects.filter(
            country_groups__name="Iron and Steel (Quota) COCs"
        ).filter(is_active=True)
```

kevinetienne (approved):

web/domains/case/_import/ironsteel/forms.py
oskusalerma: good point

mikeylockwood (approved):

PR #518, opened 2021-08-02
ICMSLST-851 Update Sanctions app to filter commodities using usage records.

MattHolmes123:
Other changes:
  - Only allow goods to be added when main application is complete.
  - Update submit view to check goods commodities.
  - Update Response prop to display commodity & disable commodity in form.


web/domains/case/_import/sanctions/forms.py
```diff
@@ -32,41 +34,47 @@ def __init__(self, *args, **kwargs):
...
-        self.fields["origin_country"].queryset = countries
-        self.fields["consignment_country"].queryset = countries
+        # Didn't use `get_usage_countries` for speed.
+        self.fields["origin_country"].queryset = Country.objects.filter(
+            country_groups__name="Sanctions and Adhoc License"
+        )
+
+        self.fields["consignment_country"].queryset = Country.objects.filter(
+            country_groups__name="Sanctions and Adhoc License Countries of shipping (consignment)"
+        )
```
kevinetienne, line 44: do we need `is_active` for both queries?

web/domains/case/_import/sanctions/views.py
```diff
@@ -309,6 +311,39 @@ def submit_sanctions(request: AuthenticatedHttpRequest, *, application_pk: int)
...
+            if goods_commodity not in sanction_and_adhoc_commodities:
+                page_errors.add(
+                    FieldError(
+                        field_name="Goods - Commodity Code",
+                        messages=[
+                            f"Commodity '{goods_commodity}' is invalid for the selected country of origin."
+                        ],
+                    )
+                )
+
```
kevinetienne, line 346: Do you need to check for the quantity based on usage data?

MattHolmes123: Probably but I wasn't going to bother for now.  
This is the message you see in legacy when editing usage records:  
`Optionally enter the maximum allocation of commodities in this group an applicant may request (textile and outward processing trade applications only). The applicant will be asked to provide evidence of past trade if they exceed this limit. Leave blank to apply no maximum.`

Wasn't sure if it was needed tbh

web/domains/case/_import/sanctions/forms.py
MattHolmes123: Sure do :) 
I don't know why I keep forgetting this.

web/domains/case/_import/sanctions/views.py
kevinetienne: Thanks, yes it makes sense

kevinetienne (approved):

oskusalerma (approved):

PR #525, opened 2021-08-03
ICMS-583: Add products to CFS Schedule

mikeylockwood:
* New model `CFSProduct` to hold products on `CFSSchedule`
* New model `CFSProductActiveIngredient` to hold active ingredients for `CFSProduct`
* New model `CFSProductType` to hold product type numbers for `CFSProduct`
* Add interface to add products on CFS schedule edit page
** Products get added / edited on a new screen
** Active ingredients / product type numbers each get added and edited on a new separate screen


web/domains/case/export/forms.py
```diff
@@ -262,6 +266,25 @@ def clean(self):
             self.add_error("goods_placed_on_uk_market", "Both of these fields cannot be yes")
             self.add_error("goods_export_only", "Both of these fields cannot be yes")
 
+        # check if legislation is not biocidal but products contain ingredients and product type numbers
+        legislations = self.cleaned_data["legislations"]
```
kevinetienne, line 270: I've spotted the other variables coming from `self.cleaned_data` have a type maybe you could add one for `legilsations`?

web/domains/case/export/forms.py
```diff
@@ -282,3 +305,115 @@ def __init__(self, *args, **kwargs):
...
+    def __init__(self, *args, schedule: CFSSchedule, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.schedule = schedule
+
+    def clean_product_name(self):
+        product_name = self.cleaned_data["product_name"]
+
+        if (
+            self.schedule.products.filter(product_name=product_name)
+            .exclude(pk=self.instance.pk)
```
kevinetienne, line 324: Do you always have an instance? (The `save` method below check if we have a pk on the instance)
What happen if instance is `None`?

web/domains/case/export/forms.py
```diff
@@ -282,3 +305,115 @@ def __init__(self, *args, **kwargs):
...
+            .exists()
+        ):
+            raise forms.ValidationError("Product name must be unique to the schedule.")
+
+        return product_name
+
+    def save(self, commit=True):
+        if not self.instance.pk:
```
kevinetienne, line 332: Is it possible to have an instance without a `pk`?

web/domains/case/export/forms.py
```diff
@@ -282,3 +305,115 @@ def __init__(self, *args, **kwargs):
...
+            product_name: str = self.cleaned_data["product_name"]
+            self.schedule.products.add(CFSProduct(product_name=product_name), bulk=False)
```
kevinetienne, line 334: Is it easier to get the product from the form and to add it in the view?

kevinetienne: Otherwise you can do something like this (we save the form first and add the product to the schedule):

```python
product = super().save(commit=commit)
self.schedule.products.add(product)
```

web/domains/case/export/forms.py
```diff
@@ -282,3 +305,115 @@ def __init__(self, *args, **kwargs):
...
+    def __init__(self, *args, product: CFSProduct, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.product = product
+
+    def clean_product_type_number(self):
+        product_type_number = self.cleaned_data["product_type_number"]
+
+        if (
+            self.product.product_type_numbers.filter(product_type_number=product_type_number)
+            .exclude(pk=self.instance.pk)
```
kevinetienne, line 355: Same comment about the instance as the previous form.

web/domains/case/export/forms.py
```diff
@@ -282,3 +305,115 @@ def __init__(self, *args, **kwargs):
...
+            .exists()
+        ):
+            raise forms.ValidationError("Product type number must be unique to the product.")
+
+        return product_type_number
+
+    def save(self, commit=True):
+        product_type_number = self.cleaned_data["product_type_number"]
+
+        if not self.instance.pk:
```
kevinetienne, line 365: Same comment about the instance as the previous form.

web/domains/case/export/urls.py
```diff
@@ -53,6 +53,58 @@
...
+                                        path(
+                                            "product-type/<int:product_type_pk>/edit/",
+                                            views.cfs_edit_product_type,
+                                            name="cfs-schedule-edit-product-type",
+                                        ),
+                                        path(
+                                            "product-type/<int:product_type_pk>/delete/",
+                                            views.cfs_delete_product_type,
+                                            name="cfs-schedule-delete-product-type",
+                                        ),
```
kevinetienne, line 104: Maybe you can group the ingredients and the products (and the following pks) together?

web/domains/case/export/views.py
```diff
@@ -392,6 +404,482 @@ def cfs_delete_manufacturer(
...
+                if not is_biocidal:
+                    return redirect(
+                        reverse(
+                            "export:cfs-schedule-edit",
+                            kwargs={"application_pk": application_pk, "schedule_pk": schedule.pk},
+                        )
+                    )
+
+                product_name = form.cleaned_data["product_name"]
+                product = schedule.products.filter(product_name=product_name).last()
```
kevinetienne, line 442: should you get the product from the form instead?
```python
            product = form.save()
```

web/domains/case/export/views.py
```diff
@@ -392,6 +404,482 @@ def cfs_delete_manufacturer(
...
+                return redirect(
+                    reverse(
+                        "export:cfs-schedule-edit",
+                        kwargs={"application_pk": application_pk, "schedule_pk": schedule.pk},
+                    )
+                )
+
+        else:
+            form = CFSProductForm(
+                data={"product_name": product.product_name}, instance=product, schedule=schedule
```
kevinetienne, line 510: do you need to pass `product_name` if you pass the `product` instance?

web/domains/case/export/views.py
```diff
@@ -392,6 +404,482 @@ def cfs_delete_manufacturer(
...
+                            "application_pk": application_pk,
+                            "schedule_pk": schedule.pk,
+                            "product_pk": product.pk,
+                        },
+                    )
+                )
+
+        else:
+            form = CFSProductTypeForm(
+                data={"product_type_number": product_type.product_type_number},
```
kevinetienne, line 824: do you need the data if you have an instance?

web/domains/case/export/views.py
```diff
@@ -461,7 +947,54 @@ def _get_cfs_errors(application: CertificateOfFreeSaleApplication) -> Applicatio
...
                 schedule_page_errors,
             )
+
+            # Check that the schedule has products
+            if not schedule.products.exists():
+                schedule_page_errors.add(
+                    FieldError(
+                        field_name="Products", messages=["At least one product must be added"]
+                    )
+                )
```
kevinetienne, line 957: To make it easier for the user and if it makes sense maybe you can create a new `PageErrors` associated with the `Add Product` url for the schedule?

web/templates/web/domains/case/export/cfs-edit-product-child.html
```diff
@@ -0,0 +1,32 @@
...
+    </ul>
+  </div>
+{% endblock %}
+
+{% block main_content %}
+  <h4>{{ page_title }}</h4>
+  {% call forms.form(action='', method='post', csrf_input=csrf_input, enctype='multipart/form-data') -%}
+    {{ application_field(product.product_name, "Product Name")}}
+    {% for field in form.fields %}
+      {{ fields.field(form[field]) }}
```
kevinetienne, line 28: `{{ fields.field(field) }}`

web/domains/case/export/forms.py
mikeylockwood: The same for is used for adding and editing  `CFSProduct`  objects. When adding a new object the instance is a `CFSProduct` object with a pk of `None`

web/domains/case/export/forms.py
mikeylockwood: Yes it's a queryset of legislations. I didn't realise subscripting the `QuerySet` model worked but it looks like it does now

web/domains/case/export/forms.py
mikeylockwood: An instance can never be `None` in a `ModelForm`

```
# ModelForm __init__
        if instance is None:
            # if we didn't get an instance, instantiate a new one
            self.instance = opts.model()
            object_data = {}
```

web/domains/case/export/forms.py
mikeylockwood: So the current structure doesn't allow a `CFSProduct` object to be created without it's parent `CFSSchedule` (related through a not null FK to the parent)

I could change this to a M2M, but thought that would create a pointless extra table as a product will always have to be associated with one schedule.

web/domains/case/export/views.py
```diff
@@ -392,6 +404,482 @@ def cfs_delete_manufacturer(
...
+            CertificateOfFreeSaleApplication.objects.select_for_update(), pk=application_pk
+        )
+        check_application_permission(application, request.user, "export")
+        application.get_task(ExportApplication.Statuses.IN_PROGRESS, "prepare")
+
+        schedule: CFSSchedule = get_object_or_404(
+            CFSSchedule.objects.select_for_update(), pk=schedule_pk
+        )
+
+        schedule.products.filter(pk=product_pk).delete()
```
mikeylockwood, line 549: In these delete views, I perform hard deletes on the objects once they are removed from the parent. Should I add a deleted column and change these to soft deletes?

web/templates/web/domains/case/export/cfs-edit-product-child.html
mikeylockwood: `field` here is the field key (a string) not the field itself. I will update the code to make that clearer

web/domains/case/export/forms.py
mikeylockwood: The way I was doing it was stupid before, I've now changed this to just add the parent to the instance if it is a new object

```
if not self.instance.pk:
    self.instance.schedule = self.schedule
```

web/domains/case/export/views.py
mikeylockwood: Yes good point, you don't

web/domains/case/export/views.py
oskusalerma: i don't think so

web/domains/case/export/urls.py
```diff
@@ -53,6 +53,72 @@
                                 views.cfs_delete_manufacturer,
                                 name="cfs-schedule-delete-manufacturer",
                             ),
+                            path(
+                                "product/add/",
+                                views.cfs_add_product,
+                                name="cfs-schedule-add-product",
+                            ),
+                            path(
```
oskusalerma, line 61: this nesting is out of control, can you move the product  stuff to a variable as other urls.py files do

oskusalerma (changes requested):

oskusalerma (approved):

PR #527, opened 2021-08-04
ICMSLST-840 Add Iron and Steel management checklist

kevinetienne:


web/domains/case/_import/ironsteel/views.py
```diff
@@ -363,3 +369,47 @@ def edit_certificate(
         }
 
         return render(request, "web/domains/case/import/ironsteel/edit_certificate.html", context)
+
+
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def manage_checklist(request: AuthenticatedHttpRequest, *, application_pk: int):
```
oskusalerma, line 376: missing return type

oskusalerma (approved):

web/domains/case/_import/ironsteel/views.py
kevinetienne: Added in https://github.com/uktrade/icms/pull/527/commits/2feb542251aa756238f9a7936f6eaf3f9f880f5f

PR #528, opened 2021-08-04
ICMSLST-886 Implement "view case" for iron and steel.

oskusalerma:


mikeylockwood (approved):

web/templates/web/domains/case/import/textiles/view.html
```diff
@@ -14,7 +14,7 @@
   {% endcall %}
 
   {% call application_section("Goods") %}
-    {{ application_field(process.category_commodity_group.pk, "Category") }}
+    {{ application_field(process.category_commodity_group.group_code, "Category") }}
```
kevinetienne, line 17: well caught :fishing_pole_and_fish:

kevinetienne (approved):

PR #535, opened 2021-08-09
ICMSLST-858 Add copy CFS application schedule endpoint.

MattHolmes123:


web/domains/case/export/views.py
```diff
@@ -866,6 +866,66 @@ def cfs_delete_product_type(
...
+        application.get_task(ExportApplication.Statuses.IN_PROGRESS, "prepare")
+
+        schedule_to_copy: CFSSchedule = get_object_or_404(application.schedules, pk=schedule_pk)
+
+        # ManyToMany you can just use `.all()` to get the records
+        # ForeignKeys you have to fetch from the db before the save.
+        legislations_to_copy = schedule_to_copy.legislations.all()
+        products_to_copy = [p for p in schedule_to_copy.products.all().order_by("pk")]
+
+        schedule_to_copy.pk = None
```
kevinetienne, line 892: As we have inheritance do we need to set `id` and `pk to `None` see the following sentence in the docs:
> Due to how inheritance works, you have to set both pk and id to None, and _state.adding to True:

```
class ThemeBlog(Blog):
    theme = models.CharField(max_length=200)

django_blog = ThemeBlog(name='Django', tagline='Django is easy', theme='python')
django_blog.save() # django_blog.pk == 3
...
django_blog.pk = None
django_blog.id = None
django_blog._state.adding = True
django_blog.save() # django_blog.pk == 4
```

web/domains/case/export/views.py
```diff
@@ -866,6 +866,66 @@ def cfs_delete_product_type(
...
 
 
+@login_required
+def cfs_copy_schedule(
+    request: AuthenticatedHttpRequest, *, application_pk: int, schedule_pk: int
+) -> HttpResponse:
+    """Copy an application schedule.
+
+    # How to copy a model instance
+    # https://docs.djangoproject.com/en/2.2/topics/db/queries/#copying-model-instances
```
kevinetienne, line 876: 3.2 maybe?

MattHolmes123: Yer good catch.
Thanks

web/domains/case/export/views.py
MattHolmes123: Discussed offline.

kevinetienne (approved):

PR #537, opened 2021-08-10
ICMSLST-582 Add CSF view application view.

MattHolmes123:
Other changes:
  - Refactor CFS templates to have common prefix.


web/templates/web/domains/case/export/cfs-view.html
```diff
@@ -0,0 +1,84 @@
...
+  {% block application_contact %}
+    {{ super() }}
+  {% endblock %}
+
+  {% call application_section("Certificate of Free Sale") %}
+    {{ application_field(application_countries|nl2br, "Countries") }}
+  {% endcall %}
+
+  {% for schedule in schedules %}
+    {% set schedule_legislations = schedule.legislations.all().values_list("name", flat=True) %}
```
mikeylockwood, line 32: Shouldn't need `.all()` here

MattHolmes123: Changed.

kevinetienne (approved):

web/domains/case/views.py
```diff
@@ -1467,6 +1470,25 @@ def _view_com(
...
+    cfs_fields = CFSSchedule._meta.get_fields()
+    labels = {f.name: getattr(f, "verbose_name", "") for f in cfs_fields}
+    app_countries = "\n".join(application.countries.all().values_list("name", flat=True))
+
+    context = {
+        "process_template": "web/domains/case/export/partials/process.html",
+        "process": application,
+        "labels": labels,
+        "application_countries": app_countries,
+        "schedules": application.schedules.filter(is_active=True),
```
mikeylockwood, line 1485: Maybe an `order_by` to guarantee the ordering when looping through the schedules in the template

mikeylockwood (approved):

web/domains/case/views.py
MattHolmes123: good shout.
Updated

PR #539, opened 2021-08-11
ICMSLST-736 & ICMSLST-801 Fix several Sanctions and Adhoc application bugs

MattHolmes123:


web/domains/case/_import/sanctions/forms.py
```diff
@@ -46,14 +46,10 @@ def __init__(self, *args, **kwargs):
 
 
 class GoodsForm(forms.ModelForm):
-    quantity_amount = forms.DecimalField(label="Quantity Amount")
-    value = forms.DecimalField(label="Value")
-
     class Meta:
         model = SanctionsAndAdhocApplicationGoods
         fields = ["commodity", "goods_description", "quantity_amount", "value"]
         widgets = {"goods_description": forms.Textarea(attrs={"cols": 80, "rows": 20})}
```
kevinetienne, line 52: I found adding `steps` to makes the input to increment by one every time the user click helpful if you see what I'm talking about:
```
        widgets = {
            "quantity_amount": forms.NumberInput(attrs={"step": 1}),
            "value": forms.NumberInput(attrs={"step": 1}),
        }
```
helps

kevinetienne (approved):

web/domains/case/_import/sanctions/forms.py
MattHolmes123: yep I can add this :+1:

PR #541, opened 2021-08-12
ICMSLST=955 Use one sidebar to create/edit/submit applications

kevinetienne:


web/templates/web/domains/case/export/cfs-edit-schedule.html
```diff
@@ -3,7 +3,7 @@
 {% from "display/fields.html" import application_section, application_field %}
 
 {% block sidebar %}
-  {% include "partial/case/export/cfs/sidebar.html" %}
+  {% include "partial/case/sidebar.html" %}
```
kevinetienne, line 6: delete please

web/templates/partial/case/link-update-requests.html
```diff
@@ -0,0 +1,8 @@
+{% if process.status == "UPDATE_REQUESTED" %}
+  <li>
+    <a
+      href="{{ icms_url('case:list-update-requests', kwargs={'application_pk': process.pk, 'case_type': 'import'}) }}">
```
MattHolmes123: Should this be `'case_type': case_type`

MattHolmes123 (changes requested):

web/templates/partial/case/link-update-requests.html
kevinetienne: Yes it should be but I've removed this file from this PR as it should be coming with the update request code.

MattHolmes123: No worries.
I thought it looked like new code :)

MattHolmes123 (approved):

mikeylockwood (approved):

PR #543, opened 2021-08-13
ICMSLST-679 Respond to Update Request

kevinetienne:


web/domains/case/app_checks.py
```diff
@@ -499,3 +521,26 @@ def _get_case_notes_errors(application: ImpOrExp, case_type: str) -> list[PageEr
...
+        page_name="Application Updates",
+        url=reverse(
+            "case:respond-update-request",
+            kwargs={"application_pk": application.pk, "case_type": case_type},
+        ),
+    )
+
+    update_requests = application.update_requests.filter(is_active=True)
+    pending_update = update_requests.filter(status=models.UpdateRequest.Status.UPDATE_IN_PROGRESS)
+    incomplete_update = update_requests.filter(status=models.UpdateRequest.Status.RESPONDED).filter(
```
mikeylockwood, line 537: Is there a reason you use two filters here instead of having both fields in the same filter query?
```
incomplete_update = update_requests.filter(status=models.UpdateRequest.Status.RESPONDED, response_detail__isnull=True)
```

kevinetienne: It follows previous comments on PRs where we decided to split it in multiple filters. Although it was just one comment and I'm not sure if we're following it.

web/domains/case/views.py
```diff
@@ -635,6 +653,113 @@ def close_update_request(
...
+
+        check_application_state(application, "prepare", case_type)
+
+        update_requests = application.update_requests.filter(is_active=True)
+        update_request = update_requests.filter(
+            status__in=[
+                models.UpdateRequest.Status.UPDATE_IN_PROGRESS,
+                models.UpdateRequest.Status.RESPONDED,
+            ]
+        ).get()
```
mikeylockwood, line 722: Why `.filter(status_in=[...]).get()` instead of just `.get(status_in=[...])`?

mikeylockwood (approved):

web/domains/case/models.py
```diff
@@ -423,9 +482,11 @@ def submit_application(self, request: AuthenticatedHttpRequest, task: Task) -> N
             # implement GMP application type.
             prefix = "CA"
 
-        self.reference = allocate_case_reference(
-            lock_manager=request.icms.lock_manager, prefix=prefix, use_year=True, min_digits=5
-        )
+        if not self.reference:
```
MattHolmes123, line 485: Do we need the extra if?
Other places that call `allocate_case_reference` don't do it.

web/domains/case/views.py
```diff
@@ -125,6 +125,28 @@ def check_application_permission(application: ImpOrExpOrAccess, user: User, case
         raise PermissionDenied
 
 
+def check_application_state(application: ImpOrExpOrAccess, task_type: str, case_type: str) -> Task:
```
MattHolmes123, line 128: As discussed I think the following would be a little clearer (name & order of function parameters):  
```
def get_application_current_task(application: ImpOrExpOrAccess, case_type: str, task_type: str):
    """Gets the current valid task for all application types.

    Also ensure the application is in the correct status for the supplied task
    """
    ...
```

web/domains/case/views.py
```diff
@@ -567,11 +575,15 @@ def manage_update_requests(
...
+                    raise NotImplementedError(
+                        f"case type {case_type} is not implemented for update requests"
+                    )
 
-                notify.update_request(
+                email.send_email.delay(
                     update_request.request_subject,
                     update_request.request_detail,
-                    contacts,
+                    list(contacts.values_list("email", flat=True)),
```
MattHolmes123, line 586: isn't it a list already?
or if it isn't it acts like one I mean.
I guess whereever it gets passed to doesn't like the values_list?

web/domains/case/views.py
```diff
@@ -635,6 +653,113 @@ def close_update_request(
...
 
 
+@login_required
+def start_update_request(
+    request: AuthenticatedHttpRequest, *, application_pk: int, update_request_pk=int, case_type: str
+) -> HttpResponse:
+    model_class = _get_class_imp_or_exp(case_type)
+
+    with transaction.atomic():
+        application: ImpOrExp = get_object_or_404(  # type: ignore[assignment]
```
MattHolmes123, line 663: What mypy error is this silencing?

web/domains/case/views.py
MattHolmes123, line 722: Can there only be one update request at a time then?

web/domains/case/views.py
```diff
@@ -635,6 +653,113 @@ def close_update_request(
...
+            ]
+        ).get()
+        previous_update_requests = update_requests.filter(status=models.UpdateRequest.Status.CLOSED)
+
+        if request.POST:
+            form = forms.UpdateRequestResponseForm(request.POST, instance=update_request)
+            if form.is_valid():
+                update_request = form.save(commit=False)
+
+                if update_request.response_detail:
```
MattHolmes123, line 730: is response_detail required in the form?  
if so wouldn't this block just be:
```
if form.is_valid():
    update_request = form.save(commit=False)
    update_request.status = models.UpdateRequest.Status.RESPONDED    
    
    update_request.response_by = request.user
    update_request.response_datetime = timezone.now()
    update_request.save()
```

MattHolmes123 (changes requested):

web/domains/case/forms.py
```diff
@@ -65,6 +65,17 @@ class Meta:
...
 
+class UpdateRequestResponseForm(forms.ModelForm):
+    class Meta:
+        model = UpdateRequest
+        fields = ("response_detail",)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["response_detail"].required = True
```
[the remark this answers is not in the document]
kevinetienne: you're right

web/domains/case/views.py
kevinetienne: Thanks for the doctring too

web/domains/case/views.py
kevinetienne: One open request yeah

web/domains/case/views.py
kevinetienne: yes, I was thinking somehow the field can be empty but no you're right.

web/domains/case/models.py
kevinetienne: If I don't safeguard it, the case get a new reference number. In v1 the case reference stays the same. What do you think?

web/domains/case/views.py
kevinetienne: Fixed in 940684a9eff2856f9b95f4ae80593ae7c53176c4.

web/domains/case/views.py
kevinetienne: Fixed in 940684a9eff2856f9b95f4ae80593ae7c53176c4.

web/domains/case/app_checks.py
```diff
@@ -499,3 +521,26 @@ def _get_case_notes_errors(application: ImpOrExp, case_type: str) -> list[PageEr
...
+        response_detail__isnull=True
+    )
+
+    if pending_update or incomplete_update:
```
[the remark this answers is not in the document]
kevinetienne: Updated in 940684a9eff2856f9b95f4ae80593ae7c53176c4.

web/domains/case/forms.py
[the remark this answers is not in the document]
kevinetienne: Fixed in fc092dc34c3f903d0123fc5dde889a1769e92716.

web/domains/case/views.py
kevinetienne: Removed in fc092dc34c3f903d0123fc5dde889a1769e92716.

web/domains/case/views.py
kevinetienne: Updated in fc092dc34c3f903d0123fc5dde889a1769e92716.

web/domains/case/views.py
kevinetienne: I thought of being clever and use a `set` as the `queryset` raise the following error when trying to delay sending the email:

>Object of type QuerySet is not JSON serializable

but `set` produces the same error!

kevinetienne: fixed in b96838e63a9ae3e3dc8a0cc8e09dc268415bcc2e.

web/domains/case/models.py
MattHolmes123: I think you're correct then.
I assume whatever you've added can call this endpoint multiple times then?

web/domains/case/app_checks.py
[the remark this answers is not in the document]
MattHolmes123: I don't see the change for this @kevinetienne?

web/domains/case/models.py
```diff
@@ -412,9 +425,57 @@ def _get_applicant_actions(
...
+                                kwargs=kwargs,
+                            ),
+                        )
+                    )
+
         return applicant_actions
 
-    def submit_application(self, request: AuthenticatedHttpRequest, task: Task) -> None:
+    def submit_application(
+        self, request: AuthenticatedHttpRequest, task: Task, override_reference: bool = True
```
MattHolmes123, line 477: How come you've added `override_reference`?

web/domains/case/models.py
kevinetienne: Yes I believe so.

web/domains/case/models.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/543/commits/1ff74a56592cca36bec9de0ffdce7a9fafb82bd6

web/domains/case/app_checks.py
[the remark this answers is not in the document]
kevinetienne: Thanks, updated in https://github.com/uktrade/icms/pull/543/commits/9cc551449ac87ae0e12e0d1a3edb86ce988b5186

MattHolmes123 (approved):

PR #547, opened 2021-08-17
ICMSLST-866 Add validation to commodity usage form.

MattHolmes123:


web/domains/commodity/forms.py
```diff
@@ -243,3 +244,46 @@ def __init__(self, *args, **kwargs):
...
+            if start_date < latest.start_date:
+                self.add_error("start_date", "Start date is before previous start date")
+
+            if latest.end_date and start_date < latest.end_date:
+                self.add_error("start_date", "Start date is before previous end date")
+
+            no_end_dates = not end_date and not latest.end_date
+            previous_no_end_date = end_date and not latest.end_date
+            if no_end_dates or previous_no_end_date:
+                self.add_error("end_date", "Previous end date has not been set.")
```
kevinetienne, line 284: should this be:
```
if not latest.end_date:
    self.add_error("end_date", "Previous end date has not been set.")
```

as I think it's saying at the moment if we have or not an `end_date` or not and not `latest.end_date` add the error.

MattHolmes123: Silly me!

kevinetienne (approved):

PR #548, opened 2021-08-18
ICMSLST-926 Add gmp_certificate_issued field to GMP application.

MattHolmes123:
Other Changes:
  - Add an icms specific RadioSelect widget that matches legacy.


web/templates/forms/widgets/radio_option.html
```diff
@@ -0,0 +1,9 @@
+{# This template is the same as django/forms/widgets/radio_option.html #}
+{# It adds the extra radio-label class that fox.css has the styling for #}
+
+{% if widget.wrap_label %}
+  <label{% if widget.attrs.id %} for="{{ widget.attrs.id }}"{% endif %} class="radio-label">
+{% endif %}
+{% include "django/forms/widgets/input.html" %}
+  {% if widget.wrap_label %} {{ widget.label }}</label>
+{% endif %}
```
mikeylockwood, line 9: This confused me at first, as the include is only one line, I think it would be neater as
```
{% if widget.wrap_label %}
  <label{% if widget.attrs.id %} for="{{ widget.attrs.id }}"{% endif %} class="radio-label">
  {% include "django/forms/widgets/input.html" %}
  {{ widget.label }}
</label>
{% else %}
  {% include "django/forms/widgets/input.html" %}
{% endif %}
```

mikeylockwood (approved):

web/templates/forms/widgets/radio_option.html
MattHolmes123: Agreed, changing now.

MattHolmes123: Hmm actually im going to leave as it is.
Only reason being its supposed to be as close to the template I copied it from (as In I only added the extra class)

MattHolmes123: The original template:
https://github.com/django/django/blob/main/django/forms/templates/django/forms/widgets/input_option.html

PR #549, opened 2021-08-18
ICMSLST-924: Edit GMP application

mikeylockwood:


web/domains/case/export/forms.py
```diff
@@ -430,20 +433,53 @@ class Meta:
...
+            "responsible_person_country": icms_widgets.RadioSelect,
         }
 
         help_texts = {"countries": ""}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        for field in self.fields:
+            self.fields[field].required = True
```
MattHolmes123, line 467: How come you've had to do this?
I don't see `blank=True` in any of the model fields so this shouldn't be needed.

web/domains/case/export/models.py
```diff
@@ -350,7 +350,89 @@ class CertificateTypes(models.TextChoices):
         ISO_22716 = ("ISO_22716", "ISO 22716")
         BRC_GLOBAL_STANDARD = ("BRC_GSOCP", "BRC Global Standard for Consumer Products")
 
-    # TODO: Add remaining model fields
+    class AddressEntryType(models.TextChoices):
```
MattHolmes123, line 353: This is in two models now, we could move it to `web/models/shared.py` like the others.
The above usage of it would be simplified if you did.
e.g. from this:
```
search = CertificateOfGoodManufacturingPracticeApplication.AddressEntryType.SEARCH
if self.instance.manufacturer_address_entry_type == search:
```
To this
```
if self.instance.manufacturer_address_entry_type == AddressEntryType.SEARCH
```

web/domains/case/export/forms.py
```diff
@@ -430,20 +433,53 @@ class Meta:
...
+
+        if self.instance.pk:
+            search = CertificateOfGoodManufacturingPracticeApplication.AddressEntryType.SEARCH
+            if self.instance.manufacturer_address_entry_type == search:
+                self.fields["manufacturer_address"].widget.attrs["readonly"] = True
+
+            if self.instance.responsible_person_address_entry_type == search:
+                self.fields["manufacturer_address"].widget.attrs["readonly"] = True
```
MattHolmes123, line 475: I think `manufacturer_address` is supposed to be `responsible_person_address`

web/domains/case/export/models.py
```diff
@@ -350,7 +350,89 @@ class CertificateTypes(models.TextChoices):
...
+    )
+
+    responsible_person_address = models.CharField(
+        max_length=4000,
+        verbose_name="Address",
+        null=True,
+    )
+
+    responsible_person_country = models.CharField(
+        max_length=30,
```
MattHolmes123, line 392: I think this only needs to be `3` looking at the `CountryType` values

web/templates/forms/fields.html
```diff
@@ -114,3 +114,31 @@
...
+      <div class="six columns input-group ">
+        <select id="id_available_addresses{{'_' + id if id else ''}}" name="available_addresses"></select>
+      </div>
+      <div class="three columns"></div>
+    </div>
+    <div name="address_field">
+      {{ field(form[address_field]) }}
+    </div>
+  </div>
+  <script src="{{ static('web/js/api/postcode-lookup.js') }}" type="text/javascript"></script>
```
MattHolmes123, line 142: By including the scripts in the macro don't both scripts get added twice on your form?
I think the scripts should be imported in the `page_js` block unless there is a good reason to include them inline like this.

web/templates/web/domains/case/export/cfs-manufacturer-address.html
```diff
@@ -52,6 +44,4 @@ <h4>{{ page_title }}</h4>
 
 {% block page_js %}
   {{ form.media.js }}
-  <script src="{{ static('web/js/api/postcode-lookup.js') }}" type="text/javascript"></script>
```
MattHolmes123, line 55: Like I said, I think we should include the scripts once here.

MattHolmes123 (changes requested):

web/domains/case/export/forms.py
mikeylockwood: I did this before removing `blank=True` and forgot to remove it

MattHolmes123: Ahh :)

web/domains/case/export/models.py
```diff
@@ -350,7 +346,85 @@ class CertificateTypes(models.TextChoices):
         ISO_22716 = ("ISO_22716", "ISO 22716")
         BRC_GLOBAL_STANDARD = ("BRC_GSOCP", "BRC Global Standard for Consumer Products")
 
-    # TODO: Add remaining model fields
+    class CountryType(models.TextChoices):
+        GB = ("GB", "Great Britain")
+        NIR = ("NIR", "Northern Ireland")
```
kevinetienne, line 351: Not sure if we follow any convention but Northern Ireland is abbreviated as `NI`.

mikeylockwood: NI is the country code for Nicaragua. Northern Ireland does not have it's own country code but it does have a region code in GB as GB-NIR. I just wanted to keep as close to the country codes as possible

MattHolmes123 (approved):

kevinetienne (approved):

PR #552, opened 2021-08-19
ICMLST-878 Fix warnings and max allocation

kevinetienne:


web/templates/partial/know-bought-from.html
```diff
@@ -0,0 +1,4 @@
+<div id="know-bought-from-info-box" class="info-box info-box-info" style="display: {{ 'block' if display_info else 'none' }}">
```
MattHolmes123: if this is the same every time:
`display_info = process.know_bought_from == False`
Why not set it at the beginning of the partial. e.g.
`{% set display_info = process.know_bought_from == False %}`

web/utils/commodity.py
```diff
@@ -158,3 +160,12 @@ def get_active_commodities(commodities: "QuerySet[Commodity]") -> "QuerySet[Comm
         commodity_end_date__gte=today,
         is_active=True,
     )
+
+
+def get_usage_data(app_type: str, app_sub_type: str = None) -> dict[str, dict[str, float]]:
+    usages = get_usage_records(ImportApplicationType.Types.TEXTILES).exclude(maximum_allocation__isnull=True)  # type: ignore[arg-type]
```
MattHolmes123, line 166: should this be hardcoded to textiles if it lives in `web/utils/commodity`

MattHolmes123 (changes requested):

web/utils/commodity.py
kevinetienne: good point! I should have passed the args back

MattHolmes123, line 166: Maybe change to a values_list rather than fetching the full model instances

web/domains/case/views.py
```diff
@@ -2574,3 +2574,27 @@ def get_page_title(case_type: str, application: ImpOrExpOrAccess, page: str) ->
         return "Access Request - {page}"
     else:
         raise NotImplementedError(f"Unknown case_type {case_type}")
+
+
+def _get_selected_product_data(application: ApplicationsWithCaseEmail) -> str:
+    assert isinstance(application, CertificateOfFreeSaleApplication)
```
MattHolmes123, line 2580: This will fall over if its any of the other application types

kevinetienne: yeah that was the idea, I didn't know if I wanted to cast the application or do something else.

web/domains/case/views.py
```diff
@@ -2574,3 +2574,27 @@ def get_page_title(case_type: str, application: ImpOrExpOrAccess, page: str) ->
...
+
+    for p in products:
+        p_types = (str(pk) for pk in p.product_type_numbers.values_list("pk", flat=True))
+        ingredient_list = p.active_ingredients.values_list("name", "cas_number")
+        ingredients = (f"{name} ({cas})" for name, cas in ingredient_list)
+
+        product = "\n".join(
+            [
+                f"Product: {p.product_name}",
+                f"Product type numbers: {', '.join(p_types)!r}",
```
MattHolmes123: You using the `!r` so you get the quotes in the string?

kevinetienne: This was fixed in the last commit https://github.com/uktrade/icms/pull/552/commits/37d0bd206ec3bb61a0e8ec0528c03fc93f33c9ee and maybe your tab didn't get refreshed. Marking it as fixed.

web/domains/case/views.py
MattHolmes123: ahh I've just seen its in the `elif application.process_type == CertificateOfFreeSaleApplication.PROCESS_TYPE:` block anyway.

mikeylockwood (approved):

web/domains/case/views.py
kevinetienne: you're correct I can pass `application.certificateoffreesaleapplication` instead. I think I prefer this solution as it will error before pushing new code if the function is reused. Thanks.

kevinetienne: updated in https://github.com/uktrade/icms/pull/552/commits/cf3275284bd1d5d55538a8682ac542886e61e530

web/utils/commodity.py
kevinetienne: I've used `values` instead as I can access the values by their attribute https://github.com/uktrade/icms/pull/552/commits/cf3275284bd1d5d55538a8682ac542886e61e530

kevinetienne: fixed in https://github.com/uktrade/icms/pull/552/commits/cf3275284bd1d5d55538a8682ac542886e61e530

web/templates/partial/know-bought-from.html
kevinetienne: I didn't know about `set`, updated in https://github.com/uktrade/icms/pull/552/commits/cf3275284bd1d5d55538a8682ac542886e61e530

web/utils/commodity.py
```diff
@@ -158,3 +159,16 @@ def get_active_commodities(commodities: "QuerySet[Commodity]") -> "QuerySet[Comm
...
+
+def get_usage_data(app_type: str, app_sub_type: str = None) -> dict[str, dict[str, float]]:
+    usages = (
+        get_usage_records(app_type, app_sub_type)
+        .exclude(maximum_allocation__isnull=True)
+        .values_list("commodity_group_id", "country_id", "maximum_allocation")
+    )
+    usage_data: dict[str, dict[str, float]] = defaultdict(dict)
+    for u in usages:
+        usage_data[u.commodity_group_id][u.country_id] = u.maximum_allocation
```
MattHolmes123, line 172: Does this work?  Don't you need `named=True` for the `values_list` call

MattHolmes123 (approved):
Approved but please fix the buggo :bug:

web/utils/commodity.py
kevinetienne: Sorry I've ignored your comment several times! Fixed now, thanks.

PR #553, opened 2021-08-19
ICMSLST-926 Add Manufacturing Certificate section to GMP Application.

MattHolmes123:
Other changes:
  - Move submit_gmp to below all other GMP views.


web/domains/case/export/views.py
```diff
@@ -1191,54 +1195,79 @@ def _get_cfs_errors(application: CertificateOfFreeSaleApplication) -> Applicatio
...
+
+                application.save()
+
+                return redirect(
+                    reverse("export:gmp-edit", kwargs={"application_pk": application_pk})
+                )
 
         else:
-            form = SubmitForm()
+            form = EditGMPForm(instance=application, initial={"contact": request.user})
```
kevinetienne, line 1238: Maybe this could be done when we create the application. I think if we agree on this we could do it in another ticket.

kevinetienne (approved):

mikeylockwood (approved):

web/domains/case/export/views.py
MattHolmes123: If we did that wouldn't we be saving a contact for them?
If im not mistaken the `initial` dict just acts like default data rather than saving anything.

PR #555, opened 2021-08-20
ICMSLST-925: Add brands to GMP application

mikeylockwood:


web/domains/case/export/forms.py
```diff
@@ -425,6 +426,13 @@ class ProductsFileUploadForm(forms.Form):
     file = ICMSFileField()
 
 
+class GMPBrandForm(forms.ModelForm):
+    class Meta:
+        model = GMPBrand
+
```
MattHolmes123, line 432: No need for new line here.

web/domains/case/export/urls.py
```diff
@@ -96,6 +96,11 @@
             [
                 path("edit/", views.edit_gmp, name="gmp-edit"),
                 path("submit/", views.submit_gmp, name="gmp-submit"),
+                path("brand/add/", views.gmp_add_brand, name="gmp-add-brand"),
+                path("brand/edit/<int:brand_pk>/", views.gmp_edit_brand, name="gmp-edit-brand"),
+                path(
+                    "brand/delete/<int:brand_pk>/", views.gmp_delete_brand, name="gmp-delete-brand"
+                ),
```
MattHolmes123, line 103: We could group the brands like the document section below

web/domains/case/export/views.py
```diff
@@ -1436,3 +1438,123 @@ def _add_cert_error(
...
+                )
+        else:
+            form = GMPBrandForm()
+
+        context = {
+            "process_template": "web/domains/case/export/partials/process.html",
+            "process": application,
+            "task": task,
+            "form": form,
+            "page_title": "Add brand",
```
kevinetienne, line 1473: can you add `"case_type": "export",` please, it would break the update requests otherwise?

web/domains/case/export/views.py
```diff
@@ -1436,3 +1438,123 @@ def _add_cert_error(
...
+        else:
+            form = GMPBrandForm(instance=instance)
+
+        context = {
+            "process_template": "web/domains/case/export/partials/process.html",
+            "process": application,
+            "task": task,
+            "form": form,
+            "instance": instance,
+            "page_title": "Edit brand",
```
kevinetienne, line 1516: can you add `"case_type": "export",` please, it would break the update requests otherwise?

web/domains/case/export/urls.py
kevinetienne, line 103: you might want to use includes for the brand.

web/templates/web/domains/case/export/gmp-edit.html
```diff
@@ -7,19 +7,26 @@
...
   </div>
 
-  <h4>Contact</h4>
-  {{ fields.field(form.contact) }}
-
   <h4>Countries</h4>
   {{ application_field(country, "Country") }}
 
+  <h4>Brands</h4>
+  {% with read_only = false, brands = process.brands.all() %}
```
MattHolmes123, line 21: Lets go with `False` (We have a mix I know)
But it is python rendering this after all :)

MattHolmes123 (approved):

PR #561, opened 2021-08-23
ICMSLST-929: GMP application view

mikeylockwood:


web/jinja2.py
```diff
@@ -71,6 +71,10 @@ def menu(request):
     return Menu().as_html(request)
 
 
+def verbose_name(obj, field_name):
+    return obj._meta.get_field(field_name).verbose_name
```
MattHolmes123, line 75: When I did this for OPT I did this:
```
    opt_fields = OutwardProcessingTradeApplication._meta.get_fields()
    labels = {f.name: getattr(f, "verbose_name", "") for f in opt_fields}
```
Not every field has a verbose name attribute and can cause an exception

MattHolmes123: I like the filter though, just extend it with an optional default

web/jinja2.py
```diff
@@ -71,6 +71,10 @@ def menu(request):
     return Menu().as_html(request)
 
 
+def verbose_name(obj, field_name):
```
MattHolmes123, line 74: Add types too.
obj being  a model etc

MattHolmes123 (approved):

PR #563, opened 2021-08-23
ICMSLST-880: Order category list on textile application

mikeylockwood:


oskusalerma:
we don't put a : after the ticket number in the commit message

PR #565, opened 2021-08-24
Remove prefix from str method from product legislations

mikeylockwood:


MattHolmes123 (approved):

web/domains/case/export/forms.py
```diff
@@ -249,9 +257,6 @@ class Meta:
             "goods_placed_on_uk_market": icms_widgets.RadioSelect,
             "goods_export_only": icms_widgets.RadioSelect,
             "any_raw_materials": icms_widgets.RadioSelect,
-            "legislations": Select2MultipleWidget(
-                attrs={"data-minimum-input-length": 0, "data-placeholder": "Select Legislation"},
-            ),
```
MattHolmes123, line 254: Can we do this in the `__init__` instead:
`self.fields["legislations"].queryset = ProductLegislation.objects.filter(is_active=True).order_by("name")`

MattHolmes123: Means we can stick to having as much information the model as we can (the label in this instance)

oskusalerma (approved):

PR #573, opened 2021-08-31
ICMSLST-943 Authorise documents view

mikeylockwood:


web/domains/case/forms.py
```diff
@@ -101,6 +103,22 @@ def clean(self):
...
 
 
+class AuthoriseForm(forms.Form):
+    password = forms.CharField(strip=False, widget=forms.widgets.PasswordInput)
+
+    def __init__(self, *args, request: AuthenticatedHttpRequest, **kwargs) -> None:
+        self.request = request
+        super().__init__(*args, **kwargs)
+
+    def clean_password(self) -> None:
```
MattHolmes123, line 113: Is everything clean at this point?

as in why not do this in the main clean method?
e.g.

```
    def clean(self) -> Dict[str, Any]:
        cleaned_data = super().clean()
        
        # Do extra password cleaning
```

web/domains/case/forms.py
```diff
@@ -101,6 +103,22 @@ def clean(self):
...
+        password = self.cleaned_data["password"]
+        user = self.request.user
+        user_cache = authenticate(self.request, username=user.username, password=password)
+
+        if user_cache is None:
+            self.add_error("password", "Please enter your password")
+
```
oskusalerma, line 120: clean_ methods need to return the value, so `return password` here?

web/domains/case/forms.py
MattHolmes123: I can never remember where the `def clean_field()` methods are called in the django form logic

web/domains/case/views.py
```diff
@@ -2124,6 +2133,44 @@ def start_authorisation(
...
+            model_class.objects.select_for_update(), pk=application_pk
+        )
+
+        task = get_application_current_task(application, case_type, "authorise")
+
+        if request.POST:
+            form = forms.AuthoriseForm(data=request.POST, request=request)
+
+            if form.is_valid():
+                return redirect(reverse("workbasket"))
```
oskusalerma, line 2154: should we use django.messages to show some message here? being dumped to workbasket is a big jump. "authorized case XXX..." kind of message

web/domains/case/views.py
```diff
@@ -143,12 +143,15 @@ def get_application_current_task(
     if case_type in ["import", "export"]:
         if task_type == "process":
             return application.get_task(
-                [application.Statuses.SUBMITTED, application.Statuses.WITHDRAWN], task_type
+                [application.Statuses.SUBMITTED, application.Statuses.WITHDRAWN],
+                task_type,
```
MattHolmes123, line 147: Did you mean to leave this `,` in?

oskusalerma (changes requested):

web/domains/case/views.py
```diff
@@ -2124,6 +2133,44 @@ def start_authorisation(
             )
 
 
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def authorise_documents(
```
kevinetienne, line 2138: do we need to list the password as sensitive see https://docs.djangoproject.com/en/3.2/howto/error-reporting/#filtering-sensitive-information

web/domains/case/forms.py
```diff
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING, Any
 
 from django import forms
+from django.contrib.auth import authenticate
```
kevinetienne, line 4: do we need to use `authenticate` from https://github.com/uktrade/icms/blob/3a1e0c9f6fde79345c1d08573bd982eeda1c26bb/web/auth/models.py#L7?

web/domains/case/views.py
mikeylockwood: I think there will be more to do here regarding validaiton and digital signing. I will find a relevent ticket number and add a TODO

web/domains/case/forms.py
mikeylockwood: Yeah I can move this to the clean method instead

web/domains/case/forms.py
mikeylockwood: I would think so, yes

mikeylockwood: Actually looking at this, it looks like this is a custom backend added for login. It seems to authenticate with Django's authenticate and then adds login attempts on failure. Do we want something similar for a user that is already logged in?

web/domains/case/views.py
mikeylockwood: I wasn't sure as the authenticate call marks it as sensitive but I guess it could error before hitting that

web/domains/case/forms.py
kevinetienne: You're right, the user is already logged in, we don't need those checks. Thanks!

MattHolmes123: The password check is a measure to prevent other people authorising it from your machine I guess?  It doesn't really matter that the user is logged in I suppose.

mikeylockwood: So looking into this, calling django authenticate does use the custom backend as it gets it from the settings. It also does suspend the account if the user is already logged in. However, if the user is logged in, the suspension currently doesn't affect the user until they log out

mikeylockwood: Added a check to see if the user is authenticated

web/domains/case/views.py
```diff
@@ -2106,6 +2108,12 @@ def start_authorisation(
             application.status = model_class.Statuses.PROCESSING
             application.save()
 
+            task.is_active = False
+            task.finished = timezone.now()
+            task.save()
+
+            Task.objects.create(process=application, task_type="authorise", previous=task)
```
kevinetienne, line 2115: Looks like you need to update cancel authorisation too now. As it would except a new state.

See https://github.com/uktrade/icms/blob/3a1e0c9f6fde79345c1d08573bd982eeda1c26bb/web/domains/case/views.py#L2127-L2145

web/domains/case/views.py
kevinetienne: You are right and I would still add it to the view to prevent to leak the ilb admin password or any password on a 3rd party platform for example (sentry, aws...). This is how you can add it:

```python
@login_required
@permission_required("web.reference_data_access", raise_exception=True)
@sensitive_post_parameters("password")
def authorise_documents(
```

I'm not sure if you need to have `sensitive_variables` in the form after adding the filter to view. This is when looking at what django does:

see
https://github.com/django/django/blob/af10e97531a59e4af09b5ec0c1a3ea476f2b6015/django/contrib/auth/views.py#L52
https://github.com/django/django/blob/af10e97531a59e4af09b5ec0c1a3ea476f2b6015/django/contrib/auth/forms.py#L163

web/domains/case/views.py
```diff
@@ -2124,6 +2132,49 @@ def start_authorisation(
             )
 
 
+@login_required
+@permission_required("web.reference_data_access", raise_exception=True)
+def authorise_documents(
+    request: AuthenticatedHttpRequest, *, application_pk: int, case_type: str
```
kevinetienne, line 2138: You are right and I would still add it to the view to prevent to leak the ilb admin password or any password on a 3rd party platform for example (sentry, aws...). This is how you can add it:

```python
@login_required
@permission_required("web.reference_data_access", raise_exception=True)
@sensitive_post_parameters("password")
def authorise_documents(
```

I'm not sure if you need to have `sensitive_variables` in the form after adding the filter to view. This is when looking at what django does:

see
https://github.com/django/django/blob/af10e97531a59e4af09b5ec0c1a3ea476f2b6015/django/contrib/auth/views.py#L52
https://github.com/django/django/blob/af10e97531a59e4af09b5ec0c1a3ea476f2b6015/django/contrib/auth/forms.py#L163

web/domains/case/views.py
```diff
@@ -2124,6 +2132,49 @@ def start_authorisation(
...
+            model_class.objects.select_for_update(), pk=application_pk
+        )
+
+        task = get_application_current_task(application, case_type, "authorise")
+
+        if request.POST:
+            form = forms.AuthoriseForm(data=request.POST, request=request)
+
+            if form.is_valid():
+                # TODO: ICMSLST-809 Check validation that is needed when generating license file
```
oskusalerma, line 2153: finish task here, set owner etc

oskusalerma (changes requested):

oskusalerma (approved):

PR #577, opened 2021-09-02
ICMSLST-940 Add basic search form for import/export applications.

oskusalerma:
ICMSLST-940 Add basic search form for import/export applications.
    
    Unfinished, committing now so we can hook up the search code to the form.


web/templates/web/domains/case/search/search.html
```diff
@@ -0,0 +1,50 @@
+{% extends "layout/no-sidebar.html" %}
+{% import "forms/forms.html" as forms %}
+{% import "forms/fields.html" as fields %}
+
+{% block main_content %}
+<div class="no-sidebar" id="content">
+    <div id="content-ie8-float">
+        <div id="form-container">
+            {% call forms.form(method='post', csrf_input=csrf_input) %}
```
kevinetienne, line 9: do we want to use `GET` instead? It will allow to bookmark a search with the query params in the url.

kevinetienne (approved):

PR #580, opened 2021-09-06
ICMSLST-952 Document packs for authorisation

mikeylockwood:


web/templates/web/domains/case/manage/document-list.html
```diff
@@ -0,0 +1,85 @@
...
+  </tr>
+  </thead>
+  <tbody>
+      {% if cover_letter_flag %}
+        <tr>
+            <td>{{ short_name }} Cover Letter</td>
+            <td>COMPLETE</td>
+            <td>
+                <a
+                    href="{{ icms_url('import:preview-cover-letter', kwargs={'application_pk': process.pk}) }}"
```
kevinetienne, line 17: I'm not sure if we should be using this link for the preview. `preview-cover-letter` is an overview of the cover letter. It's used before the case was authorised. I believe once the case has been authorised the button to view the PDF will link to the fully generated PDF.

web/templates/web/domains/case/manage/document-list.html
```diff
@@ -0,0 +1,85 @@
...
+            </td>
+        </tr>
+      {% endif %}
+      <tr>
+          <td>{{ short_name }} Licence{{" (Holders Copy)" if customs_copy else ""}}</td>
+          <td>COMPLETE</td>
+          <td>
+              {# TODO: Replace this link with generated licence #}
+              <a
+                  href="{{ icms_url('import:preview-licence', kwargs={'application_pk': process.pk}) }}"
```
kevinetienne, line 31: I'm not sure if we should be using this link for the preview. `preview-licence` is an overview of the cover letter. It's used before the case was approved. I believe once the case has been authorised the button to view the PDF will link to the fully generated PDF.

web/templates/web/domains/case/document-packs.html
```diff
@@ -0,0 +1,32 @@
...
+        </a>
+      </li>
+    </ul>
+  </div>
+{% endblock %}
+
+{% block main_content %}
+  <h3>Document Pack</h3>
+  {{ application_field("Response", "Document Set") }}
+  {{ application_field(process.reference | string + " Response Documents", "Document Pack") }}
```
kevinetienne, line 26: Can reference be something different than a string?

web/domains/case/_import/models.py
```diff
@@ -29,6 +31,16 @@ class Types(models.TextChoices):
         TEXTILES = ("TEX", "Textiles (Quota)")
         WOOD_QUOTA = ("WD", "Wood (Quota)")
 
+    class TypesShortNames(Enum):
```
oskusalerma, line 34: as discussed lets just use the Types long name, not invent another list of type names

oskusalerma (changes requested):

web/domains/case/_import/models.py
mikeylockwood: Just remembered the issue I had with this. Taking Iron and Steel as the example:

We store `IS` in the type column. How can we then get back to the long name  in the text choices from the type saved in the column? In the `Types` class this would need to be referenced by `IRON_STEEL` not `IS`

web/templates/web/domains/case/manage/document-list.html
mikeylockwood: Added a TODO for the ticket to generate the PDF

web/templates/web/domains/case/manage/document-list.html
mikeylockwood: Added a TODO for the ticket to generate the PDF

oskusalerma (approved):

PR #602, opened 2021-09-16
ICMSLST-1036 Add search by case status for Import Applications

kevinetienne:


mikeylockwood (approved):

web/tests/utils/test_search.py
```diff
@@ -2,6 +2,7 @@
 import io
 from typing import TYPE_CHECKING
 
+import pytest
 from django.utils.timezone import make_aware
 from openpyxl import load_workbook
 from pytest import fixture
```
MattHolmes123, line 8: Do you want to remove this and replace with `pytest.fixture` please

MattHolmes123 (approved):

web/tests/utils/test_search.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/602/commits/2fbdb1bf9d42fcf336e33ea066448b46acd542fd

PR #605, opened 2021-09-17
ICMSLST-1116 Add Process.get_specific_model to downcast models.

oskusalerma:
ICMSLST-1116 Add Process.get_specific_model to downcast models.
    
    Also add typing.final and IS_FINAL to each final application model
    subclass. The first helps mypy enforce nobody subclasses them further,
    and the latter is needed for the downcast to work.

    ICMSLST-1116 Move ProcessTypes to web/flow/models, combine it into one.
    
    This is a global unique mapping, no point pretending otherwise, and it
    complicates things greatly trying to do anything with it when you have
    to import subclasses from the base classes.

    ICMSLST-1116 Enforce Process.process_type is always set.


kevinetienne (approved):

mikeylockwood (approved):

web/tests/domains/case/_import/test_models.py
```diff
@@ -58,3 +74,48 @@ def create_import_application(self):
...
+        TextilesApplication,
+        WoodQuotaApplication,
+    ],
+)
+@pytest.mark.django_db
+def test_import_downcast(application_model):
+    importer = ImporterFactory(
+        offices=(OfficeFactory(is_active=True), OfficeFactory(is_active=True))
+    )
+    user = UserFactory(is_active=True)
```
MattHolmes123, line 99: There are fixtures already for `importer` and `test_import_user` in `icms/web/tests/conftest.py`.  
Those fixtures fetch the preexisting record rather than creating a new one for each test like these tests will.

web/tests/flow/test_models.py
```diff
@@ -39,3 +39,11 @@ def test_get_task_error():
     with pytest.raises(errors.TaskError) as excinfo:
         process.get_task("submitted", "process")
     assert f"Expected one active task, got {len(tasks)}"
+
+
+@pytest.mark.django_db
+def test_downcast_unknown():
+    p = Process(process_type="blaa")
+
+    with pytest.raises(RuntimeError, match="Unknown process_type blaa"):
```
MattHolmes123, line 48: Any reason to not check for the error it raises?  e.g.  
`with pytest.raises(NotImplementedError, match="Unknown process_type blaa"):`

oskusalerma: oh i always forget the syntax for checking exceptions in pytest, so i search for it, and i copypasted it, and forgot to change the type..

web/tests/domains/case/_import/test_models.py
oskusalerma: ok will use those

web/tests/flow/test_models.py
MattHolmes123: :)

MattHolmes123 (approved):

PR #606, opened 2021-09-17
ICMSLST-954: Add supplementary reports to firearm import applications

mikeylockwood:


web/domains/case/_import/fa/views.py
```diff
@@ -536,6 +537,130 @@ def provide_report(request: AuthenticatedHttpRequest, *, application_pk: int) ->
...
+
+        supplementary_info: SupplementaryInfo = application.supplementary_info
+
+        if request.POST:
+            form = SupplementaryReportForm(
+                data=request.POST, application=application, supplementary_info=supplementary_info
+            )
+
+            if form.is_valid():
+                report: SupplementaryReport = form.save()
```
kevinetienne, line 561: If this is working we can set the supplementary info in the view?

```python
report: SupplementaryReport = form.save(commit=False)
report.supplementary_info = report.supplementary_info
report.save()
```

kevinetienne: I think adding `supplementary_info` as above would avoid to pass it in the form for the edit view.

web/domains/case/_import/fa/models.py
```diff
@@ -95,13 +79,57 @@ class SupplementaryInfo(models.Model):
...
+        SupplementaryInfo, related_name="reports", on_delete=models.CASCADE
+    )
+    transport = models.CharField(choices=TransportType.choices, max_length=4, blank=False)
+    date_received = models.DateField(verbose_name="Date Received")
+
+    bought_from = models.ForeignKey(
+        ImportContact,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="+",
```
kevinetienne, line 100: I was going to suggest `limit_choice_to` but we don't know the reverse relationship back to application from `supplementary_info`... It might work if the base class was not an abstract class but it might bring other problems. Maybe something to evaluate.


```python
        limit_choices_to=models.Q(application=supplementary_info.application),
````


> Sets a limit to the available choices for this field when this field is rendered using a ModelForm or the admin (by default, all objects in the queryset are available to choose). Either a dictionary, a Q object, or a callable returning a dictionary or Q object can be used.

File: web/domains/case/_import/fa/views.py (hunk 1 above)
Comment (mikeylockwood) [reply web/domains/case/_import/fa/views.py:561]: Yes, good point

Comment (MattHolmes123) [reply web/domains/case/_import/fa/views.py:561]: I guess you mean this @kevinetienne ?
`report.supplementary_info = application.supplementary_info`

File: web/domains/case/_import/fa/forms.py [hunk 3]
```diff
@@ -71,3 +78,28 @@ def __init__(self, *args, application, **kwargs):
...
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.supplementary_info = supplementary_info
+        self.fields["bought_from"].queryset = application.importcontact_set.all()
+
+    def save(self, commit=True):
+        if not self.instance.pk:
+            self.instance.supplementary_info = self.supplementary_info
```
MattHolmes123, line 103: I prefer what kevin suggests later :+1:

web/domains/case/_import/fa/models.py
```diff
@@ -95,13 +79,57 @@ class SupplementaryInfo(models.Model):
...
+
+    # TODO: ICMSLST-960: Add FK for uploaded document
+    # TODO: ICMSLST-961: Goods reference for firearm needs to be linked to this info
+
+    report = models.ForeignKey(
+        SupplementaryReport, related_name="firearms", on_delete=models.CASCADE
+    )
+    serial_number = models.CharField(max_length=20, null=True)
+    calibre = models.CharField(max_length=20, null=True)
+    model = models.CharField(max_length=20, verbose_name="Make and Model", null=True)
```
MattHolmes123, line 122: I think the max_length isn't big enough for serial_number, calibre and model.
I can exceed your limits in legacy
image

web/domains/case/_import/fa/forms.py
```diff
@@ -71,3 +73,18 @@ def __init__(self, *args, application, **kwargs):
...
+        fields = ("transport", "date_received", "bought_from")
+        widgets = {"date_received": DateInput}
+
+    def __init__(self, *args, application: FaImportApplication, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.fields["bought_from"].queryset = application.importcontact_set.all()
+
+    def save(self, commit=True):
+        return super().save(commit=commit)
```
MattHolmes123, line 90: Do we need to define `def save` anymore?

web/templates/web/domains/case/import/fa/partials/report-list.html
```diff
@@ -0,0 +1,85 @@
+{% if not reports %}
+  <div class="info-box info-box-info">
+    There are no supplementary reports provided.
+  </div>
+{% else %}
+  {% for report in reports %}
+    <h3>
+      {{ report }}
```
MattHolmes123, line 8: I don't really like using a `__str__` method to a template UI but thats just my preference.  
Its just not obvious when looking at the template what it will do.

web/templates/web/domains/case/import/fa/partials/report-list.html
```diff
@@ -0,0 +1,85 @@
...
+      {% for firemarm in report.firearms.all() %}
+        {# TODO: ICMSLST-960: Include link to uploaded document if not manually entered #}
+        {# TODO: ICMSLST-961: Include all relevant firearm fields if not uploaded document #}
+        <tr>
+          <td></td>
+          <td>{{ firearm.serial_number }}</td>
+          <td>{{ firearm.calibre }}</td>
+          <td>{{ firearm.model }}</td>
+          <td></td>
+          {% if not upload %}
```
MattHolmes123, line 58: Do you have conflicting logic intentionally between the header and data rows?
`{% if upload %}` and `{% if not upload %}`

web/domains/case/_import/fa/models.py
mikeylockwood: I like the idea of using `limit_choice_to`, but yes it won't work with the current model structure. I still think the best way would be to have a fk back from `SupplementaryInfo` to `ImportApplication`

MattHolmes123 (approved):

PR #611, opened 2021-09-23
Refactor provide report section

mikeylockwood:


web/domains/case/_import/fa_oil/views.py
```diff
@@ -219,3 +229,200 @@ def manage_checklist(request: AuthenticatedHttpRequest, *, application_pk: int)
...
+        report: OILSupplementaryReport = supplementary_info.reports.get(pk=report_pk)
+
+        if request.POST:
+            form = OILSupplementaryReportForm(
+                data=request.POST, instance=report, application=application
+            )
+
+            if form.is_valid():
+                report: OILSupplementaryReport = form.save(commit=False)  # type: ignore[no-redef]
+                report.supplementary_info = supplementary_info
```
MattHolmes123, line 342: Sane comment about this one (and I assume the other edit_report view)

web/domains/case/_import/fa_oil/views.py
```diff
@@ -219,3 +229,200 @@ def manage_checklist(request: AuthenticatedHttpRequest, *, application_pk: int)
...
+        application: OpenIndividualLicenceApplication = get_object_or_404(
+            OpenIndividualLicenceApplication.objects.select_for_update(), pk=application_pk
+        )
+
+        check_application_permission(application, request.user, "import")
+
+        task = get_application_current_task(application, "import", Task.TaskType.ACK)
+
+        supplementary_info: OILSupplementaryInfo = application.supplementary_info
+        report: OILSupplementaryReport = supplementary_info.reports.get(pk=report_pk)
```
MattHolmes123, line 399: Can these be combined (in the three `add_report_firearm_manual` views)
`application.supplementary_info.reports.get(pk=report_pk)`

web/domains/case/_import/fa/types.py
```diff
@@ -1,5 +1,51 @@
...
 
 FaImportApplication = Union[OpenIndividualLicenceApplication, DFLApplication, SILApplication]
+FaSupplementaryInfo = Union[DFLSupplementaryInfo, OILSupplementaryInfo, SILSupplementaryInfo]
+FaSupplementaryReport = Union[
+    DFLSupplementaryReport, OILSupplementaryReport, SILSupplementaryReport
+]
+FaSupplementaryReportFirearm = Union[
+    DFLSupplementaryReportFirearm, OILSupplementaryReportFirearm, SILSupplementaryReportFirearm
+]
+FaSupplementaryReportForm = Type[
```
kevinetienne, line 42: I think we've tried to append `T` at the end of the variable when making a new new type see https://github.com/uktrade/icms/blob/6ad30400b292ff5dfa64452e8ca787bd10e66a35/web/domains/case/types.py#L34 for example.

web/domains/case/_import/fa/types.py
```diff
@@ -1,5 +1,51 @@
...
+    Union[DFLSupplementaryReportForm, OILSupplementaryReportForm, SILSupplementaryReportForm]
+]
+FaSupplementaryReportFirearmForm = Type[
```
kevinetienne, line 45: same comment about the type (T) here

web/domains/case/_import/fa/views.py
```diff
@@ -694,3 +755,29 @@ def _get_fa_application(application: ImportApplication) -> FaImportApplication:
     firearms_application: FaImportApplication = getattr(application, link)
 
     return firearms_application
+
+
+def _get_supplementary_report_form(application: FaImportApplication) -> FaSupplementaryReportForm:
+    if application.process_type == OpenIndividualLicenceApplication.PROCESS_TYPE:
```
kevinetienne, line 761: As an alternative and following what Osku has done you can do `ProcessTypes.FA_OIL`.

web/domains/case/_import/fa/types.py
MattHolmes123, line 45: When defining types that are a `Type` lets have a `T` on the end like we do elsewhere

web/domains/case/_import/fa/views.py
```diff
@@ -656,6 +670,53 @@ def delete_report(
...
+            "case_type": "import",
+            "contacts": application.importcontact_set.all(),
+            "page_title": "Firearms Supplementary Information Overview",
+            "form": form,
+            "report": report,
+        }
+
+        return render(
+            request=request,
+            template_name="web/domains/case/import/fa/provide-report/edit-report.html",
```
MattHolmes123, line 715: A view to add a report returns a template called `edit-report.html`
I guess it all just uses the same form?

MattHolmes123 (approved):

web/domains/case/_import/fa/views.py
mikeylockwood: Ah yeah, it's just rendering the same form for now. I'm changing it in ICMSLT-961. Not gonna add a todo now as I will pick it up straight after merging

PR #616, opened 2021-09-27
ICMSLST-1125 Change CFS schedules to use FK, not ManyToMany.

oskusalerma:
ICMSLST-1125 Change CFS schedules to use FK, not ManyToMany.
    
    Also remove is_active, delete schedules when they're deleted.


web/domains/case/export/views.py
```diff
@@ -1035,12 +1035,15 @@ def cfs_copy_schedule(
 
         schedule_to_copy: CFSSchedule = get_object_or_404(application.schedules, pk=schedule_pk)
 
+        # FIXME: adapt to schedules using FK
```
MattHolmes123, line 1038: Noo

oskusalerma: fixed

mikeylockwood (approved):

MattHolmes123 (approved):

PR #618, opened 2021-09-29
ICMSLST-1120 Update BEIS recipient and attachments for case email

kevinetienne:
- Make recipient for BEIS and HSE editable
- Add all the attachments by default for BEIS email

### DIFF
--- a/web/domains/case/views/views_email.py
+++ b/web/domains/case/views/views_email.py
@@ -437,7 +437,7 @@ def _create_email(application: ApplicationsWithCaseEmail) -> models.CaseEmail:
         raise Exception(f"CaseEmail for application not supported {application.process_type}")
 
 
-def _create_fa_case_email(application: FaImportApplication):
+def _create_fa_case_email(application: FaImportApplication) -> models.CaseEmail:
     template = Template.objects.get(is_active=True, template_code="IMA_CONSTAB_EMAIL")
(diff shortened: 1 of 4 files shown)


web/domains/case/views/views_email.py
```diff
@@ -417,99 +417,128 @@ def _get_case_email_application(application: ImpOrExp) -> ApplicationsWithCaseEm
...
-        )
-        cc_address_list = []
+    elif application.process_type == pt.GMP:
+        return create_gmp_case_email(application)
 
     else:
         raise Exception(f"CaseEmail for application not supported {application.process_type}")
 
+
+def _create_fa_case_email(application: FaImportApplication):
```
MattHolmes123, line 440: return types for these new functions

kevinetienne: Oh yes!

MattHolmes123 (approved):

mikeylockwood (approved):

web/domains/case/views/views_email.py
kevinetienne: Updated the return types in https://github.com/uktrade/icms/pull/618/commits/daf50e9fe2de22ab16ff79bee4fec8ac2fe3cb36 for the 4 new functions, thanks.

PR #619, opened 2021-09-29
ICMSLST-1035 Add wildcard filter function for searching applications. 

MattHolmes123:
Other changes:

Add several indexes to core application tables.


web/utils/search.py
```diff
@@ -792,6 +791,37 @@ def _apply_export_application_filter(
     return model
 
 
+def _apply_wildcard_filter(
+    model: "QuerySet[Model]", field: str, search_pattern: str
+) -> "QuerySet[Model]":
+    """Strings with `%` are converted into django orm code using one of several methods.
```
oskusalerma, line 797: s/orm/ORM/

web/utils/search.py
```diff
@@ -792,6 +791,37 @@ def _apply_export_application_filter(
...
+    :param field: The name of the field to search on
+    :param search_pattern: the user supplied search pattern
+    """
+
+    if search_pattern.strip() == "%":
+        return model
+
+    # The default search unless changed below
+    search_regex = search_pattern.replace("%", ".*")
+    search = {f"{field}__iregex": f"^{search_regex}"}
```
oskusalerma, line 809: if we're doing regex search, do we need to escape regex characters in user supplied string?

oskusalerma (changes requested):

web/utils/search.py
MattHolmes123: Probably, I haven't really thought about it yet

MattHolmes123: Added a subtask for this.

web/utils/search.py
```diff
@@ -270,6 +270,36 @@ def get_search_results_spreadsheet(case_type: str, results: SearchResults) -> by
...
+    :param field: The name of the field to search on
+    :param search_pattern: the user supplied search pattern
+    """
+
+    if search_pattern.strip() == "%":
+        return {}
+
+    # The default search unless changed below
+    search_regex = search_pattern.replace("%", ".*")
+    search = {f"{field}__iregex": f"^{search_regex}"}
```
mikeylockwood, line 287: Does the regex need to be escaped before creating a regex search?

```
search_regex = re.escape(search_pattern).replace("//%", ".*")
"""

mikeylockwood: Maybe that won't work with  `istartswith` and `iendswith` below

MattHolmes123: I have a story to escape the regex later.
re.escape isn't necessarily the correct thing to do.

mikeylockwood (approved):

kevinetienne (approved):

PR #623, opened 2021-10-01
ICMSLST-68 Add documents to Mailshot

kevinetienne:


web/domains/mailshot/actions.py
```diff
@@ -9,20 +7,25 @@
...
 
 
 class View(actions.View):
     def display(self, mailshot):
-        return mailshot.status != Mailshot.DRAFT
+        return mailshot.status != Mailshot.Statuses.DRAFT
+
+
+class ViewReceived(View):
+    def href(self, object):
```
MattHolmes123, line 19: Are we bothered about typing these methods?

MattHolmes123 (approved):

web/domains/mailshot/actions.py
kevinetienne: Good point I think we should, as they are far from going anywhere.

mikeylockwood (approved):

web/domains/mailshot/actions.py
kevinetienne: The list view are far from going anywhere I should have added.

kevinetienne: Updated in https://github.com/uktrade/icms/pull/623/commits/846c342a3b13c10a5fadf5f600055609ba2959e3

PR #624, opened 2021-10-04
ICMSLST-941 Add advanced search import application search fields.

MattHolmes123:


web/domains/case/forms_search.py
```diff
@@ -88,6 +89,86 @@ def clean(self):
...
+
+    chief_usage_status = forms.ChoiceField(
+        label="CHIEF Usage Status",
+        required=False,
+        choices=[
+            (None, "Any"),
+            ("cancelled", "Cancelled"),
+            ("exhausted", "Exhausted"),
+            ("expired", "Expired"),
+            ("surrendered", "Surrendered"),
```
kevinetienne, line 109: You can get the Chief types if you want from https://github.com/uktrade/icms/blob/0c84977838dc613ffd3f7ee1394fe5c899ce565b/web/domains/case/_import/models.py#L147-L151

However you need to update the display value for `Surrendered`

kevinetienne (approved):

web/domains/case/forms_search.py
MattHolmes123: Cool

PR #625, opened 2021-10-04
ICMSLST-1140 Update permissions for mailshot

kevinetienne:
Replace `mailshot_access` with `reference_data_access` and `importer_access`/`exporter_access`. Only case workers can manage mailshots and only importers/exporters/agents can see them.

Fix permissions for `MailshotDetailView` (case worker's view) and `MailshotReceivedDetailView` (importer/exporter's view).


web/domains/mailshot/views.py
```diff
@@ -336,3 +328,18 @@ def delete_document(
...
+    if user.has_perm("web.reference_data_access"):
+        return True
+
+    importer_access = user.has_perm("web.importer_access")
+    exporter_acesss = user.has_perm("web.exporter_access")
+
+    if importer_access or exporter_acesss:
+        return True
+
+    return False
```
mikeylockwood, line 345: `return importer_access or exporter_access`

mikeylockwood (approved):

MattHolmes123 (approved):

web/domains/mailshot/views.py
kevinetienne: Updated in https://github.com/uktrade/icms/pull/625/commits/66d83752c7fd3a89e9197052e3af5cdf5b092ebf

PR #629, opened 2021-10-05
ICMSLST-1086 Add search logic for several fields

MattHolmes123:


web/domains/case/_import/models.py
```diff
@@ -158,7 +158,9 @@ class Meta:
                 name="IA_search_case_reference_idx",
                 opclasses=["text_pattern_ops"],
             ),
-            models.Index(fields=["-submit_datetime"], name="IA_submit_datetime_idx"),
+            models.Index(
+                *[models.Q(submit_datetime__isnull=False)], name="IA_submit_datetime_notnull_idx"
```
mikeylockwood, line 162: Can this first arg just be `models.Q(submit_datetime__isnull=False)`

web/migrations/0001_initial.py
```diff
@@ -6139,7 +6139,9 @@ class Migration(migrations.Migration):
         ),
         migrations.AddIndex(
             model_name="importapplication",
-            index=models.Index(fields=["-submit_datetime"], name="IA_submit_datetime_idx"),
+            index=models.Index(
+                models.Q(("submit_datetime__isnull", False)), name="IA_submit_datetime_notnull_idx"
```
mikeylockwood, line 6143: Double parenthesis on the `Q` call

mikeylockwood (approved):

web/migrations/0001_initial.py
mikeylockwood: Obviously the first arg is a tuple

MattHolmes123: Django is doing this not me :)

web/domains/case/_import/models.py
MattHolmes123: Yer it totally can, I will change it tomorrow thanks

PR #630, opened 2021-10-06
ICMSLST-1152 Fix mailshot templates for 'edit' and 'detail' views

kevinetienne:


web/templates/web/domains/mailshot/edit.html
```diff
@@ -1,7 +1,48 @@
 {% extends "layout/sidebar.html" %}
 {% import "forms/forms.html" as forms %}
 {% import "forms/fields.html" as fields %}
+{% from "display/fields.html" import application_field %}
 
+{% block css %}
```
mikeylockwood, line 6: Would prefer a block of css like this to be imported in stylesheet rather than defined in the template

mikeylockwood (approved):

MattHolmes123 (approved):

PR #634, opened 2021-10-14
ICMSLST-1099 Add reference and version to mailshot

kevinetienne:
A new version of a mailshot is created after a mailshot has been retracted. A mailshot can be retracted when it has been published.

* ICMSLST-1099 Add Publishing status to Mailshot
* ICMSLST-1171 Add republish when mailshot is retracted …
* ICMSLST-1151 Set version and reference
* ICMSLST-1100 Add search for current mailshot


web/domains/mailshot/views.py
```diff
@@ -120,14 +138,30 @@ def form_valid(self, form):
...
+
+        with transaction.atomic():
+            response = super().form_valid(form)
+
+            mailshot = self.get_object()
+            mailshot.status = Mailshot.Statuses.PUBLISHED
+            mailshot.published_datetime = timezone.now()
+            mailshot.published_by = self.request.user
+            if not mailshot.reference:
+                mailshot.reference = allocate_case_reference(
```
MattHolmes123, line 150: Hmm not 100% about this one  
I mean it works fine, I just wish the model wasn't called "CaseReference" as its grown beyond case references now.

web/tests/domains/mailshot/test_views.py
```diff
@@ -236,3 +251,97 @@ def test_importer_access_forbidden(self):
...
+
+    if reference_version:
+        icms = ICMSMiddlewareContext()
+        mailshot.reference = allocate_case_reference(
+            lock_manager=icms.lock_manager,
+            prefix="MAIL",
+            use_year=False,
+            min_digits=1,
+        )
+        mailshot.version += 1
```
MattHolmes123, line 344: See race condition comment

web/domains/mailshot/views.py
kevinetienne: Yes the same comment apply to access request in fact.
I can create a ticket to rename the function to `allocate_reference` and the model to `Reference`.

web/tests/domains/mailshot/test_views.py
kevinetienne: Fixed in Updated in https://github.com/uktrade/icms/pull/634/commits/7effcfd6a523bc3487dec674bbc103775accd906

PR #639, opened 2021-10-18
ICMSLST-983 Add list IMI Cases view.

MattHolmes123:


web/templates/layout/no-sidebar.html
```diff
@@ -1,11 +1,14 @@
 {% extends "layout/sidebar.html" %}
 
 {% block content %}
-<div class="no-sidebar" id="content">
-  <div class="content-ie8-float">
-    {% block content_actions %}{%endblock%}
-    {% include "partial/mainstart.html" %}
-    {% block main_content %}{%endblock%}
+  <div id="sidebar-substitute"></div>
+  <div id="sidebar"><h4></h4></div>
```
mikeylockwood, line 5: What is the empty `h4` for?

mikeylockwood, line 5: What is the empty `h4` for?

MattHolmes123: It does not :)
Removing now.

PR #655, opened 2021-11-02
ICMSLST-1185 Fix Docker build for the web container

davidwtbuxton:
Running `docker build -t [tag] .` was failing with an error about missing
the pg_config helper (when installing Python deps). Adding an explicit
dependency on libpq-dev fixes this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.9.5-slim
 
 # Install dependencies
-RUN apt-get update && apt-get install wget graphviz libgraphviz-dev gcc postgresql-client npm -y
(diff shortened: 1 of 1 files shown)


davidwtbuxton:
I'm still not sure what caused this breakage. If it is related to me building the container on an M1 Mac, there may be other things that are broken, and so perhaps the better fix is something to do with making sure we pick a consistent container architecture (i.e. force amd64).

davidwtbuxton:
I've tested the Docker container build steps again on an Intel Mac, and they work fine without this patch. So I am pretty sure that it is specifically me using an M1 Mac, which uses arm64 Docker images, that triggered the build failing. And having this explicit dependency makes it work again (and so far I haven't hit any other problems running the local Docker containers on an M1 machine).

MattHolmes123 (approved):

PR #658, opened 2021-11-10
ICMSLST-1029 Link to create a certificate application from a template

davidwtbuxton:
Added a link to create an application in the list of the user's saved application templates.

The "/export/?from-template=[ID]" path will redirect to the specific type depending on the saved template. If the template ID is invalid, or if the user is not the owner of the template, then the user sees the normal page for choosing an application type.

(Restricting template use to the owner prevents leaking PII - when we implement sharing a template with other users this policy will need to be updated.)

The certificate application edit form views now accept a "?from-template" query parameter and on GET (but not POST) will use the template to set initial values for the form data.

Dummy templates have been added for the ilb_admin and exporter_user users, for the different application types.


web/domains/case/export/views.py
```diff
@@ -268,7 +299,10 @@ def edit_cfs(request: AuthenticatedHttpRequest, *, application_pk: int) -> HttpR
                 )
 
         else:
-            form = EditCFSForm(instance=application, initial={"contact": request.user})
+            app_template = try_application_template_from_request(request)
+            initial = app_template.initial_data() if app_template else {}
+            initial["contact"] = request.user
+            form = EditCFSForm(instance=application, initial=initial)
```
davidwtbuxton, line 305: The CFS application form actually uses several views with forms for different parts of the application data:

- edit_cfs
- cfs_add_schedule
- cfs_edit_schedule
- cfs_set_manufacturer
- cfs_add_product
- cfs_edit_product
- cfs_add_ingredient
- cfs_edit_ingredient
- cfs_add_product_type
- cfs_edit_product_type

I have only updated the `edit_cfs` view to use the application template data because I'm not sure yet how all these views should best be updated to handle templates.

Like, this pattern of _get template, get initial data, add other initial data, create form with initial data_ is clearly repetitive and perhaps should be extracted into a separate function for re-use. Or maybe it shouldn't be in each view at all, and instead the template initial data should be applied somewhere in the `create_export_application` view, by directly updating the application database record (but then might it be weird / unexpected to create a related object such as `CFSProductActiveIngredient`?).

davidwtbuxton:
As discussed, I will update this PR to include a basic check that the user is the owner of the template. Later this will need to be updated if we implement sharing a template with others.

web/domains/case/export/views.py
```diff
@@ -82,6 +96,20 @@ def _exporters_with_agents(user: User) -> List[int]:
     return [exporter.pk for exporter in exporters_with_agents]
 
 
+def try_application_template_from_request(
+    request: HttpRequest,
```
mikeylockwood, line 100: We use `AuthenticatedHttpRequest` as a wrapper for `HttpRequest`

MattHolmes123: As it defines the extra attribute added by the middleware

web/management/commands/add_dummy_data.py
```diff
@@ -245,3 +250,34 @@ def create_superuser(self, username: str, password: str) -> User:
     def _assign_permission(self, user, permission_codename):
         permission = Permission.objects.get(codename=permission_codename)
         user.user_permissions.add(permission)
+
+
+def create_certificate_application_templates(
+    owner: User,
+) -> Collection[CertificateApplicationTemplate]:
```
MattHolmes123, line 257: This can be typed as `list[CertificateApplicationTemplate]`

MattHolmes123:
I'm going to have a play around with this branch after standup

web/domains/case/export/views.py
davidwtbuxton: Updated, thanks.

web/management/commands/add_dummy_data.py
davidwtbuxton: Updated, thanks.

MattHolmes123:
Added this as a suggestion to extend:
https://github.com/uktrade/icms/pull/659

PR #659, opened 2021-11-11
ICMSLST-1029 Create GMP app from template without query params.

MattHolmes123:
No need for query param or extra redirect hop to "export:choose".

Would still need to extend for the other two application types

### DIFF
--- a/web/domains/case/export/views.py
+++ b/web/domains/case/export/views.py
@@ -45,6 +45,7 @@ from .forms import (
     EditCFScheduleForm,
     EditCFSForm,
     EditGMPForm,
+    EditGMPTemplateForm,
     GMPBrandForm,
     PrepareCertManufactureForm,
     ProductsFileUploadForm,
@@ -112,7 +113,18 @@ def try_application_template_from_request(
 
 @login_required
 @permission_required("web.exporter_access", raise_exception=True)
-def create_export_application(request: AuthenticatedHttpRequest, *, type_code: str) -> HttpResponse:
+def create_export_application(
+    request: AuthenticatedHttpRequest, *, type_code: str, template_pk: int = None
+) -> HttpResponse:
+    """Create a certificate application.
+
+    Supports setting application data from an optional template.
+
+    :param request: Request
+    :param type_code: Type of application to create.
+    :param template_pk: Optional PK of a template to populate application
+    """
+
     application_type: ExportApplicationType = ExportApplicationType.objects.get(type_code=type_code)
 
     config = _get_export_app_config(type_code)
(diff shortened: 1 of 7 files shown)


davidwtbuxton (approved):

davidwtbuxton (approved):

web/domains/case/export/views.py
```diff
@@ -179,34 +161,29 @@ def create_export_application(
...
-        return None
-
-    return template if template.user_can_view(user) else None
+    if template.user_can_view(user):
+        return template
+    else:
+        raise PermissionDenied
 
 
 def set_template_data(
```
MattHolmes123, line 184: I only implemented the change the GMP application btw.
Are you going to do it for the other two application types?

web/domains/case/export/views.py
```diff
@@ -97,24 +88,10 @@ def _exporters_with_agents(user: User) -> List[int]:
...
-        return None
-
-    return template if template.user_can_view(request.user) else None
-
-
 @login_required
 @permission_required("web.exporter_access", raise_exception=True)
 def create_export_application(
-    request: AuthenticatedHttpRequest, *, type_code: str, template_pk: int = None
+    request: AuthenticatedHttpRequest, *, type_code: TextChoices, template_pk: int = None
```
MattHolmes123, line 94: The url converter used for type_code returns  a string:

```
class ExportApplicationTypeConverter:
    regex = "cfs|com|gmp"

    def to_python(self, value):
        return value.upper()

    def to_url(self, value):
        if value not in ["cfs", "com", "gmp"]:
            raise ValueError

        return value
```

web/domains/case/export/views.py
```diff
@@ -215,16 +195,21 @@ def set_template_data(
...
+    types_forms: dict[TextChoices, ModelForm] = {
+        # These form classes have no required fields, no data cleaning methods.
+        ExportApplicationType.Types.FREE_SALE: EditCFSTemplateForm,
+        ExportApplicationType.Types.GMP: EditGMPTemplateForm,
+        ExportApplicationType.Types.MANUFACTURE: PrepareCertManufactureTemplateForm,
+    }
+    try:
+        form_class = types_forms[type_code]
+    except KeyError:
+        raise Exception(f"Type not supported: {type_code}")
```
MattHolmes123, line 207: raise a `NotImplementedError` (everywhere else does).  
My bad as I raised a base Exception when I did my PR.

web/domains/case/export/views.py
mikeylockwood: This might be best as a literal so it is limited to the string choices

```..., type_code: Literal["cfs", "com", "gmp"]..

MattHolmes123: Even better :+1: 
Love a literal

web/domains/case/export/views.py
davidwtbuxton: Changed to `NotImplementedError`.

web/domains/case/export/views.py
davidwtbuxton: Changed the `types_forms` dictionary to `dict[Any, ModelForm]` which is loose enough to satisfy the type checker, and reverted the change to the `create_export_application(..)` signature.

PR #663, opened 2021-11-16
ICMSLST-1187 Add country form validation for Derogation application.

MattHolmes123:


davidwtbuxton (approved):
Approving, but is this list of countries serving the same purpose as the "Derogation from Sanctions COOs" `CountryGroup`  set?

https://github.com/uktrade/icms/blob/33b9d099e97caafbe357bea7d79d4ec18707ed49/web/migrations/0003_add_country_group_data.py#L1224

MattHolmes123:
When we go live we will probably need to update the "Derogation from Sanctions COOs" country group to reflect whatever it is on the live system.
We will also need to revisit all the commodity groups to work out how to populate them.
If you look in v1 it allows you to choose more than the four countries in the "Derogation from Sanctions COOs" country group.
But the form logic only checks those 4 countries.
This application isn't live anymore either (but I suppose it could be brought back)

mikeylockwood (approved):

PR #666, opened 2021-11-17
ICMSLST-1190 Allow editing of basic application template details

davidwtbuxton:
Added a new route "/cat/edit/[pk]/[step]/" for editing the template data for use with a certificate application template model. The view requires the same permissions as the template creation view, and checks that the user has edit permissions for that template (currently requires the user is the owner of the template).

Choice fields (select inputs) return their data as Django models, which are serialized as the object's primary key. This is different to Django's own serializer framework, which represents a model as a dict/object with the key, the app name and refs for foreign keys.


davidwtbuxton:
The live system shows the template editing form with the same layout as the certificate application form, but as discussed I haven't implemented that in this patch. It would be good to re-use all the layout from our current certificate application views, and that will take some refactoring of those templates so they can be used in the different contexts.

web/domains/cat/views.py
```diff
@@ -98,3 +114,51 @@ def _has_permission(user: User) -> bool:
     exporter_user = user.has_perm("web.exporter_access")
 
     return ilb_admin or exporter_user
+
+
+class CATEditStepView(PermissionRequiredMixin, LoginRequiredMixin, FormView):
```
mikeylockwood, line 119: I do prefer class based views.. except for the mixins

mikeylockwood (approved):

web/domains/cat/views.py
```diff
@@ -66,14 +69,17 @@ def create(request: AuthenticatedHttpRequest) -> HttpResponse:
 
 
 @login_required
-def edit(request: AuthenticatedHttpRequest, *, cat_pk: int) -> HttpResponse:
-    cat = get_object_or_404(CertificateApplicationTemplate, pk=cat_pk)
+def edit(request: AuthenticatedHttpRequest, *, pk: int) -> HttpResponse:
```
MattHolmes123, line 72: Any reason for changing this back to pk?  
By convention project tries to be less ambiguous about the view kwargs.
I don't mind pk but this was decided by the team a while back (admittedly most of that team has now left) :D

web/domains/cat/views.py
```diff
@@ -84,10 +90,20 @@ def edit(request: AuthenticatedHttpRequest, *, cat_pk: int) -> HttpResponse:
         else:
             form = EditCATForm(instance=cat)
 
+        type_ = cat.application_type
+        sidebar_links = [
+            (reverse("cat:edit", kwargs={"pk": cat.pk}), "Template"),
+            (
+                reverse("cat:edit-step", kwargs={"pk": cat.pk, "step": type_.lower()}),
+                f"{type_} Application",
```
MattHolmes123, line 98: Should we use the full names for the links?
```
>>> from web.models import CertificateApplicationTemplate
>>> cat = CertificateApplicationTemplate.objects.first()
>>> cat.application_type
'CFS'
>>> cat.get_application_type_display()
'Certificate of Free Sale'

```

web/domains/cat/views.py
davidwtbuxton: I originally used the generic object UpdateView for the CATEditStepView, and that view uses "pk" by default, so it made sense to use the same path key for the edit view too. I will switch them both to use "cat_pk".

web/domains/cat/views.py
davidwtbuxton: I'll update it to use the full name.

web/tests/domains/cat/test_views.py
```diff
@@ -0,0 +1,126 @@
...
+            "name": "Foo name",
+            "description": "Foo description",
+            "application_type": "CFS",
+            "sharing": "private",
+        }
+        self.login_with_permissions(["exporter_access"])
+        response = self.client.post(url, data)
+
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response["Location"], "/cat/")
```
MattHolmes123, line 24: These two asserts could be the following:
```
self.assertRedirects(response, reverse("cat:list")
```

web/tests/domains/cat/test_views.py
```diff
@@ -0,0 +1,126 @@
+from django.urls import reverse
+
+from web.domains.case.export.models import ExportApplicationType
+from web.domains.cat.models import CertificateApplicationTemplate
+from web.tests.auth import AuthTestCase
+
+
+class TestCATCreateView(AuthTestCase):
```
MattHolmes123, line 8: could easily be pytest style tests if an `exporter_client` fixture was added but no point changing it just for the sake of it.  
Tests have been added :heart: :heart: :heart:

MattHolmes123 (approved):

web/domains/cat/views.py
MattHolmes123: Cool.
For reference you can do this when you need to:

```
from django.views.generic.edit import UpdateView

class MyUpdateView(UpdateView):
    pk_url_kwarg = "cat_pk"
```

web/tests/domains/cat/test_views.py
davidwtbuxton: By default assertRedirects fetches the redirect URL, which I think is annoying. I think having the expected text of the URL makes the test easier to read and understand when something inadvertently breaks or changes the actual resolved URL paths.

PR #668, opened 2021-11-19
ICMSLST-1208 Enable case authorisation when application has been refused.

MattHolmes123:


web/domains/case/app_checks.py
```diff
@@ -40,44 +37,34 @@
...
 
-    prepare_errors = PageErrors(
-        page_name="Response Preparation",
-        url=reverse(
-            "case:prepare-response",
-            kwargs={"application_pk": application.pk, "case_type": case_type},
-        ),
-    )
+    # check for checklist errors
+    checklist_errors = get_checklist_errors(application)
```
mikeylockwood, line 47: Both import and export applications are passed to this but the function is typed only to `ImportApplication`

mikeylockwood (approved):

davidwtbuxton (approved):

web/domains/case/app_checks.py
MattHolmes123: Thanks I've fixed this.
Also removed a comment that was a lie :)

PR #670, opened 2021-11-21
ICMSLST-1178 Archiving for certificate application templates

davidwtbuxton:
Also ICMSLST-1204, restoring a previously archived template.

This changes the /cat/ template list view so that for each template in the list there is an action to archive or restore that template, which updates the CertificateApplicationTemplate.is_active boolean field appropriately.

The queryset manager for .objects.active() and .objects.inactive() will be useful when we implement the filtering feature for the template list.


web/domains/cat/managers.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+
+class CATManager(models.Manager):
+    def active(self):
```
MattHolmes123, line 5: Should we type the manager methods?

`-> "QuerySet[CertificateApplicationTemplate]"`

MattHolmes123 (approved):

web/domains/cat/managers.py
davidwtbuxton: Updated with types.

mikeylockwood (approved):

PR #671, opened 2021-11-22
ICMSLST-1043 Implement filtering for the template list view ("/cat/")

davidwtbuxton:


MattHolmes123 (approved):

web/templates/web/domains/cat/list.html
```diff
@@ -9,26 +9,27 @@
     </ul>
 
     <h4>Filter</h4>
-    {% call forms.form(action="#TODO: ICMSLST-1043 Filter templates", method='post', csrf_input=csrf_input) -%}
-        {% for field in form %}
+    {% call forms.form(action=".", method='GET') -%}
```
mikeylockwood, line 12: What does `action="."` resolve to? If you want to submit the form to the same page, you can remove the `action` kwarg

mikeylockwood (approved):

web/templates/web/domains/cat/list.html
davidwtbuxton: Updated with `action=""`.

PR #675, opened 2021-11-25
ICMSLST-1044 Add a read-only version of the application templates

davidwtbuxton:
In the list of templates ("/cat/"), clicking the _name_ of a template links to a read-only version of the templates, like "/cat/view/[id]/". The read-only view re-uses the editing views, but with form inputs disabled.

### DIFF
--- a/web/domains/cat/urls.py
+++ b/web/domains/cat/urls.py
@@ -6,8 +6,10 @@ app_name = "cat"
 urlpatterns = [
     path("", views.CATListView.as_view(), name="list"),
     path("create/", views.create, name="create"),
-    path("edit/<int:cat_pk>/", views.edit, name="edit"),
+    path("edit/<int:cat_pk>/", views.CATEditView.as_view(), name="edit"),
     path("edit/<int:cat_pk>/archive/", views.CATArchiveView.as_view(), name="archive"),
     path("edit/<int:cat_pk>/restore/", views.CATRestoreView.as_view(), name="restore"),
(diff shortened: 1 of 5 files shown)


web/domains/cat/urls.py
```diff
@@ -6,8 +6,18 @@
...
+        "edit/<int:cat_pk>/<str:step>/",
+        views.CATEditStepView.as_view(editing=True),
+        name="edit-step",
+    ),
+    path("view/<int:cat_pk>/", views.CATEditStepView.as_view(editing=False), name="view"),
+    path(
+        "view/<int:cat_pk>/<str:step>/",
+        views.CATEditStepView.as_view(editing=False),
+        name="view-step",
+    ),
```
davidwtbuxton, line 22: So this takes the CATEditStepView and 

1. adds a .editing attribute, which is used in the view methods and in the web/domains/cat/base_html.html template to disable the form's <input> elements and remove the <button> elements.
2. uses the CATEditStepView for the initial "/cat/edit/[id]/" template edit form, and inside the view methods test for whether "step" is in the path keyword arguments to support both "/cat/edit/[id]/" and "/cat/edit/[id]/[step]/".

I am happy enough with using the edit view for a read-only view by introducing that ".editing" attribute, but I am not sure it is worth making the CATEditStepView work for both the initial template form and the template's data form(s).

Alternative to removing the separate "edit" view would be either trying to extend it so it too takes an `editing` parameter, or duplicating some of that logic for a pure read-only view.

MattHolmes123 (approved):

mikeylockwood (approved):

PR #677, opened 2021-11-25
ICMSLST-1206 Add Search actions for Import applications.

MattHolmes123:


web/templates/web/domains/case/search/import/actions.html
```diff
@@ -0,0 +1,12 @@
+<ul class="menu-out">
+  {% for action in rec.actions  %}
+    <li>
+      <form method="post" action="{{ action.url }}">
+        <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}"/>
```
davidwtbuxton, line 5: Worth using `{{ csrf_input }}` here? https://docs.djangoproject.com/en/3.2/ref/csrf/#using-csrf-in-jinja2-templates

davidwtbuxton (approved):

web/templates/web/domains/case/search/import/actions.html
MattHolmes123: yes my bad.
I copied the workbasket template

MattHolmes123: Thanks

web/templates/web/domains/case/search/import/actions.html
```diff
@@ -0,0 +1,12 @@
...
+        <button type="submit" class="link-button {{ action.icon }} button">
+          {{ action.link_text }}
```
MattHolmes123, line 7: Bug from a refactor.
Should be {{ action.name }}

mikeylockwood (approved):

PR #678, opened 2021-11-26
ICMSLST-1004 Add Reopen Case search action

MattHolmes123:


web/domains/case/views/mixins.py
```diff
@@ -0,0 +1,53 @@
...
+    """View for updating an application.
+
+    Supports checking and updating status and the current active task
+    """
+
+    model = Process
+    pk_url_kwarg = "application_pk"
+
+    expected_status: ClassVar[Union[str, list[str]]]
+    current_task: ClassVar[Optional[Task]] = None
```
MattHolmes123, line 22: Note: Type this as `str`

web/domains/case/urls.py
```diff
@@ -28,6 +28,18 @@
...
+        include(
+            [
+                path(
+                    "reopen-case",
+                    views_search.ReopenApplicationView.as_view(),
+                    name="search-reopen-case",
+                ),
+            ]
+        ),
+    ),
```
MattHolmes123, line 42: I initially considered adding this to the views_misc.py file as that's where a lot of the "manage" views are.
However after some testing this view only appears on the search screen.
When you stop a case it disappears from the workbasket and you have to go to the search screen to reopen it.

web/flow/models.py
```diff
@@ -57,6 +58,32 @@ class Process(models.Model):
...
+        """Get the expected active current task"""
+
+        if not self.is_active:
+            raise errors.ProcessInactiveError("Process is not active")
+
+        try:
+            task = self.get_active_tasks().get(task_type=task_type)
+
+        except ObjectDoesNotExist:
+            raise errors.TaskError(f"Expected active task doesn't exist: {task_type}")
```
davidwtbuxton, line 71: Would using `raise TaskError(..) from err` be useful here to provide more context? May also make sense to handle both ObjectDoesNotExist and MultipleObjectsReturned with a single except since both cases raise a TaskError.

davidwtbuxton (approved):

web/domains/case/views/mixins.py
```diff
@@ -0,0 +1,100 @@
...
+from django.utils import timezone
+from django.views import View
+from django.views.generic.detail import SingleObjectMixin
+
+from web.domains.case.types import ImpOrExp
+from web.flow.models import Process, Task
+from web.types import AuthenticatedHttpRequest
+
+
+class ApplicationUpdateView(PermissionRequiredMixin, LoginRequiredMixin, SingleObjectMixin, View):
```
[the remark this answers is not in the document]
MattHolmes123: and If this was done we'd need to refactor the select_for_update call to only happen for post requests etc.

mikeylockwood (approved):

web/domains/case/views/mixins.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def get_queryset(self) -> models.QuerySet:
+        """Return the model instance ready for update."""
+
+        return self.model.objects.select_for_update().all()
+
+    def get_object(self, queryset: models.QuerySet = None) -> ImpOrExp:
+        """Downcast to specific model class."""
+
+        application = super().get_object(queryset).get_specific_model()
```
mikeylockwood, line 67: This gets the model but not the application?

web/flow/models.py
MattHolmes123: What you've suggested is a lot better :+1:

web/domains/case/views/mixins.py
MattHolmes123: Discussed offline

PR #680, opened 2021-12-01
ICMSLST-1173 Fix success message when a mailshot is published.

MattHolmes123:


davidwtbuxton (approved):

web/domains/mailshot/views.py
```diff
@@ -175,7 +183,20 @@ def publish(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRes
             form.add_error(None, "A document must be uploaded before publishing")
 
         if form.is_valid():
-            return self.form_valid(form)
+            response = self.form_valid(form)
```
mikeylockwood, line 186: Would it be easier to not call `super` on line 151 and instead just do
`response = HttpResponseRedirect(self.get_success_url())` instead of undoing what `super` is doing.

MattHolmes123: `response = super().form_valid(form)`
That does the form saving. 

I could change it I suppose is

mikeylockwood: If that's the case then your change is the simplest. I think we should remove the `SuccessMessageMixin` from the class, but not worth the time now

mikeylockwood (approved):

davidwtbuxton (approved):
Shiny

MattHolmes123:
Closing this PR as we are going with the following:
https://github.com/uktrade/icms/pull/681

PR #683, opened 2021-12-06
ICMSLST-686 Add Request Variation view. 

MattHolmes123:


web/domains/case/views/views_search.py
```diff
@@ -171,6 +179,91 @@ def post(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRespon
...
+        if self.request.session.get("search_results_url"):
+            return self._get_return_url()
+
+        return self._default_return_url()
+
+    def _get_return_url(self) -> str:
+        """Check for a search_results_url and rebuild the search request."""
+
+        url = self.request.session.get("search_results_url")
+        return_url: parse.ParseResult = parse.urlparse(url)
```
mikeylockwood, line 237: Will this break if `search_results_url` is not in the session?

mikeylockwood: No it won't, it can handle `None`

mikeylockwood (approved):

web/domains/case/views/mixins.py
```diff
@@ -99,3 +114,21 @@ def update_application_tasks(self) -> None:
...
+        return super().get(request, *args, **kwargs)
+
+    def post(self, request: HttpRequest, *args, **kwargs) -> Any:
+        self.set_application_and_task()
+
+        return super().post(request, *args, **kwargs)
+
+    @property
+    def is_post(self) -> bool:
+        return self.request.method.lower() == "post"
```
davidwtbuxton, line 130: Since Django guarantees that request.method is always upper case, I think it makes sense to test that with `request.method == "POST"`. Or is there a situation where it _isn't_ upper case?

https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.method

Although I think I would favour removing both `is_post` and `is_get`, I don't think they simplify much.

davidwtbuxton (approved):
With comments.

Also I see this updates things in a models.py file, was wondering if this would impact the initial migrations.

web/domains/case/views/mixins.py
MattHolmes123: Didn't know it was guaranteed :+1: 
I will update.

PR #685, opened 2021-12-10
ICMSLST-1236 Manage variation requests

MattHolmes123:
Other changes: 

- ICMSLST-1244 Redirect users to workbasket after login
- Update dev script to create applications.


davidwtbuxton (approved):

web/domains/case/models.py
```diff
@@ -308,6 +299,8 @@ def get_workbasket_subject(self) -> str:
         raise NotImplementedError
 
     def get_reference(self) -> str:
+        # TODO: ICMSLST-1287 Need to work out how to increment the variations requested.
```
mikeylockwood, line 302: Could this just be a count of the variations?

mikeylockwood (approved):

web/domains/case/_import/models.py
```diff
@@ -184,10 +184,14 @@ class Meta:
     under_appeal_flag = models.BooleanField(blank=False, null=False, default=False)
 
     variation_decision = models.CharField(
-        max_length=10, choices=ApplicationBase.DECISIONS, blank=True, null=True
+        max_length=10,
+        choices=ApplicationBase.DECISIONS,
+        null=True,
+        verbose_name="Variation Decision",
+    )
```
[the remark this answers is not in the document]
mikeylockwood: Yes it does

web/domains/case/views/views_search.py
```diff
@@ -208,18 +208,35 @@ def get(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRespons
...
+                "-requested_datetime"
+            ),
+        }
 
     def form_valid(self, form: RequestVariationForm) -> HttpResponseRedirect:
         """Store the variation request before redirecting to the success url."""
 
         variation_request: VariationRequest = form.save(commit=False)
-        variation_request.status = VariationRequest.DRAFT
+        variation_request.status = VariationRequest.OPEN
```
mikeylockwood, line 226: Is there a check to make sure there isn't an active variation request open for this application before creating a new one?

MattHolmes123: Not directly no (you don't get the action to do it though if there is an open one).

web/domains/case/models.py
MattHolmes123: I did think that, and it could, however its an extra db query every time.
I need to have a proper think about updating the CaseReference table

MattHolmes123: Problem for ICMSLST-1287 either way

web/domains/case/views/views_search.py
MattHolmes123: Im going to add some logic to the form now for this @mikeylockwood

MattHolmes123: Discussed offline, no change needed.

PR #697, opened 2022-01-12
ICMSLST-1284 Add partial form saving to main FA SIL application form.

MattHolmes123:


web/domains/case/_import/fa_sil/views.py
```diff
@@ -319,18 +321,28 @@ def edit(request: AuthenticatedHttpRequest, *, application_pk: int) -> HttpRespo
...
+            else:
+                messages.error(request, "Failed to save application data, please correct errors.")
 
         else:
             initial = {} if application.contact else {"contact": request.user}
-            form = forms.PrepareSILForm(instance=application, initial=initial)
+            form_kwargs = {"instance": application, "initial": initial}
+
+            # query param to validate the form (useful when returning from submit link)
+            if request.GET.get("validate", None):
```
davidwtbuxton, line 341: We could omit the `None` argument to get here because it is the default.

davidwtbuxton (approved):

web/domains/case/_import/fa_sil/views.py
MattHolmes123: Ahh yes good point

mikeylockwood, line 341: Just thinking, should validating be the default? So a query param will turn off validation rather than turn it on

mikeylockwood (approved):

PR #698, opened 2022-01-12
Update run-tests.sh to not run coverage by default.

MattHolmes123:
Also fix error in testing section of README.md


davidwtbuxton (approved):
Assuming it actually works… ;)

PR #699, opened 2022-01-13
ICMSLST-1269 Stub methods for CHIEF integration

davidwtbuxton:
The path for the call-back is "/chief/license-data-callback".

New settings for configuring Hawk authentication:

- CHIEF_LICENSE_URL, the icms-hmrc API to send license requests.
- HAWK_AUTH_ID, client ID shared by icms and icms-hmrc (not a secret).
- HAWK_AUTH_KEY, signing key shared by icms and icms-hmrc (secret).

Both HAWK_AUTH_ID and HAWK_AUTH_KEY must be shared with the sender (in our case, the sender is icms-hmrc).

When sending data to the call-back, we must return either a 207 multi status or 208 already reported status. For a 207 response, the body is a JSON object with keys for "accepted" and "rejected", and with values a list of objects like `{"id": <license-id>}`.

### DIFF
--- a/web/domains/chief/client.py
+++ b/web/domains/chief/client.py
@@ -1,0 +1,48 @@
+import typing
+
+import mohawk
+import requests
+from django.conf import settings
+
+HTTPMethod = typing.Literal["GET", "OPTIONS", "HEAD", "POST", "PUT", "PATCH", "DELETE"]
+
+
+def make_hawk_sender(method: HTTPMethod, url: str, **kwargs) -> mohawk.Sender:
+    creds = {
+        "id": settings.HAWK_AUTH_ID,
+        "key": settings.HAWK_AUTH_KEY,
+        "algorithm": "sha256",
+    }
+
+    return mohawk.Sender(creds, url, method, **kwargs)
+
+
+def make_request(method: HTTPMethod, url: str, **kwargs) -> tuple[mohawk.Sender, requests.Response]:
+    # Requests allows calling with a dict which is converted to a JSON body,
+    # but we need the final body and type to create the Hawk signature, which
+    # is then added as an auth header to the request.
+    request = requests.Request(method, url, **kwargs)
+    prepped = request.prepare()
+    hawk_sender = make_hawk_sender(
+        method, url, content=prepped.body, content_type=prepped.headers["Content-type"]
+    )
+    prepped.headers["Authorization"] = hawk_sender.request_header
+    session = requests.Session()
+    response = session.send(prepped)
+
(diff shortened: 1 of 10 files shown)


web/domains/chief/client.py
```diff
@@ -0,0 +1,44 @@
...
+    creds = {
+        "id": settings.HAWK_AUTH_ID,
+        "key": settings.HAWK_AUTH_KEY,
+        "algorithm": "sha256",
+    }
+
+    return mohawk.Sender(creds, url, method, **kwargs)
+
+
+def make_request(method: str, url: str, **kwargs) -> tuple[mohawk.Sender, requests.Response]:
```
MattHolmes123, line 16: Consider a literal?

davidwtbuxton: Do we have one already for HTTP methods?

web/domains/chief/views.py
```diff
@@ -0,0 +1,81 @@
...
+from django.utils.crypto import constant_time_compare
+from django.views.generic import View
+
+HAWK_ALGO = "sha256"
+HAWK_NONCE_EXPIRY = 60  # seconds
+NONCE_CACHE_PREFIX = "hawk-nonce"
+HAWK_RESPONSE_HEADER = "Server-Authorization"
+
+
+def get_credentials_map(access_id: str) -> dict:
```
MattHolmes123, line 19: `dict[str, Any]` as the return type.

web/domains/chief/client.py
MattHolmes123: haven't got a clue :)

web/tests/domains/chief/test_client.py
```diff
@@ -0,0 +1,27 @@
+from unittest import mock
+
+from django.test import override_settings
+
+from web.domains.chief import client
+
+
+# N.B. Requires you install requests-mock.
```
MattHolmes123, line 8: this comment needed?

MattHolmes123 (approved):

mikeylockwood (approved):

web/tests/domains/chief/test_client.py
davidwtbuxton: Removed comment.

web/domains/chief/client.py
davidwtbuxton: Updated function annotation to use `HTTPMethod = typing.Literal["GET", "OPTIONS", "HEAD", "POST", "PUT", "PATCH", "DELETE"]`.

PR #705, opened 2022-01-19
ICMSLST-982 Add all countries to deactivated firearms form choices

davidwtbuxton:
The "Firearms and Ammunition (Deactivated) Issuing Countries" country group
is used on the "Firearms and Ammunition (Deactivated Firearms Licence)" form
and its related "add goods certificate" form.

On the legacy ICMSv1 form we have "Burma (Myanmar)", but I changed that to
"Burma" to match the existing country names we have in ICMSv2. We should
confirm what name to use for this country and update where necessary.


web/migrations/0003_add_country_group_data.py
```diff
@@ -42,7 +42,7 @@ def load_country_group_data_iron_and_steel_coc(apps, schema_editor):
         "Brazil",
         "British Virgin Islands",
         "Brunei",
-        "Burma (Myanamar)",
+        "Burma",
```
MattHolmes123, line 45: See this ticket:
https://uktrade.atlassian.net/browse/ICMSLST-763
I think its supposed to be this: Myanmar (Burma)

https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/974911/FCDO_Geographical_Names_Index.csv/preview

davidwtbuxton: The countries created by [the initial migrations include "Burma"][1]. We would need either to change that initial country name, and update references to that name in the groups migration, or use "Burma".

This specific line 45 in the group data for "Burma (Myanamar)" means it doesn't get added to the group at all, because it doesn't match an existing Country record in the database (and the [add group logic filters by name][1]).

[1]: https://github.com/uktrade/icms/blob/a17d7894cb17cfdfd5caae8e4a69a452e5048962/web/migrations/0002_add_country_data.py#L186[2]: https://github.com/uktrade/icms/blob/a17d7894cb17cfdfd5caae8e4a69a452e5048962/web/migrations/0003_add_country_group_data.py#L9

MattHolmes123: ICMS V1 displays it as Burma right? 
That's wrong I think, we should just use the correct name throughout "Myanmar (Burma)".
What do you think?

MattHolmes123 (approved):

web/migrations/0003_add_country_group_data.py
davidwtbuxton: Updated, change all references to "Myanmar (Burma)".

PR #707, opened 2022-01-20
ICMSLST-1366 Add view to withdraw a variation request update request.

MattHolmes123:


web/domains/case/views/views_variation_request.py
```diff
@@ -206,6 +200,55 @@ def get_success_url(self):
...
+        self.set_application_and_task()
+
+        # Reset the update request reason
+        variation_request.update_request_reason = None
+        variation_request.save()
+
+        # Close the task
+        self.update_application_tasks()
+
+        return redirect(
```
davidwtbuxton, line 241: I think one can remove the `reverse(..)` here and pass the view name and keyword arguments directly to `redirect(..)`.

davidwtbuxton (approved):

web/domains/case/views/views_variation_request.py
MattHolmes123: Good spot!

mikeylockwood (approved):

PR #708, opened 2022-01-20
ICMSLST-1388 New command "./manage.py countries export"

davidwtbuxton:
The command writes the country records and country group names to `<stdout>`
in CSV format. Where a country belongs to a group, there is a "Y" in the
cell (the group names are the column headers).

Use this like "./manage.py countries export > report.csv" to save the
output to a file named "report.csv".

There is also a "./manage.py countries import" command that reads from
`<stdin>`, but all it does at the moment is print how many countries and
groups were in the CSV.

### DIFF
--- a/web/management/commands/countries.py
+++ b/web/management/commands/countries.py
@@ -1,3 +1,4 @@
+import argparse
 import csv
 import operator
 import sys
@@ -18,7 +19,7 @@ MEMBERSHIP = "Y"
 class Command(BaseCommand):
     help = "Import countries data from <stdin> or export to <stdout>"
 
-    def add_arguments(self, parser):
+    def add_arguments(self, parser: argparse.ArgumentParser):
         parser.add_argument(
             "operation", choices=["import", "export"], help="Import or export countries data"
         )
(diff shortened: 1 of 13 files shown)


web/management/commands/countries.py
```diff
@@ -0,0 +1,106 @@
...
+# when reading a CSV and loading data.
+COUNTRY_FIELDS = ["id", "name", "is_active", "type", "commission_code", "hmrc_code"]
+# Marker in a spreadsheet cell to indicate group membership.
+MEMBERSHIP = "Y"
+
+
+class Command(BaseCommand):
+    help = "Import countries data from <stdin> or export to <stdout>"
+
+    def add_arguments(self, parser):
```
mikeylockwood, line 21: `parser: ArgumentParser`

mikeylockwood (approved):

web/management/commands/countries.py
davidwtbuxton: Fixed.

MattHolmes123 (approved):

PR #710, opened 2022-01-21
ICMSLST-1387 Move country and country group initial data to a spreadsheet

davidwtbuxton:
Instead of having all the countries and what groups they belong to defined
in the migration code, the configuration has been moved to a CSV spreadsheet
(generated using `./manage.py countries export`). Then in the initial data
migrations, we read the country names and stuff from the CSV.


davidwtbuxton:
It's difficult to see from the diff in GitHub exactly how the 2 migrations have changed, so it may help to view the complete files:

- https://github.com/uktrade/icms/blob/ICMSLST-1387/web/migrations/0002_add_country_data.py
- https://github.com/uktrade/icms/blob/ICMSLST-1387/web/migrations/0003_add_country_group_data.py

web/migrations/0002_add_country_data.py
```diff
@@ -1,1259 +1,33 @@
...
-            Country(
-                name="Uganda",
-                is_active=True,
-                type="SOVEREIGN_TERRITORY",
-                commission_code="350",
-                hmrc_code="UG",
-            ),
-        ]
-    )
+    return Country.objects.bulk_create(countries)
```
MattHolmes123, line 30: Do you need to return here?

davidwtbuxton: Nope, but at one point i was calling this function in a shell and wanted to see what was created. I'll remove it.

MattHolmes123 (approved):

mikeylockwood (approved):

PR #711, opened 2022-01-24
ICMSLST-1362 WIP workbasket actions refactor & "Acknowledge Completed Application" refactor

MattHolmes123:


web/domains/workbasket/actions/applicant_actions.py
```diff
@@ -0,0 +1,364 @@
...
+class EditApplicationAction(Action):
+    """Applicant action to resume editing the application"""
+
+    def show_link(self) -> bool:
+        show_link = False
+
+        if self.status == ImpExpStatus.IN_PROGRESS:
+            show_link = True
+
+        return show_link
```
mikeylockwood, line 26: `return self.status == ImpExpStatus.IN_PROGRESS` unless there is more logic to be added to it

web/domains/case/models.py
```diff
@@ -680,7 +413,9 @@ def history(self) -> None:
 
         print("All tasks in order:")
         for t in all_tasks:
-            print(f"Task: {t.get_task_type_display()}, {t.created}, {t.finished}")
+            created = t.created.strftime("%Y/%m/%d %H:%M")
+            finished = t.finished.strftime("%Y/%m/%d %H:%M") if t.finished else ""
+            print(f"Task: {t.get_task_type_display()}, created={created!r}, finished={finished!r}")
```
davidwtbuxton, line 418: Why use a specific date format here instead of the repr? Isn't this a debugging thing?

davidwtbuxton (approved):
That queryset annotation for the active task looks cool.

web/domains/workbasket/views.py
```diff
@@ -47,12 +55,16 @@ def _get_queryset_admin(user: User) -> chain[QuerySet]:
         is_active=True, status=AccessRequest.Statuses.SUBMITTED
     ).prefetch_related(Prefetch("tasks", queryset=Task.objects.filter(is_active=True)))
 
-    export_applications = ExportApplication.objects.filter(is_active=True).prefetch_related(
-        Prefetch("tasks", queryset=Task.objects.filter(is_active=True))
+    export_applications = (
+        ExportApplication.objects.filter(is_active=True)
+        .select_related("exporter", "contact", "application_type", "submitted_by", "case_owner")
```
mikeylockwood, line 60: 👍

mikeylockwood (approved):
Minor comment about simplifying the `show_link` methods, feel free to ignore. LGTM :+1:

web/domains/case/models.py
MattHolmes123: I just found it easier to read when I was using it without the seconds and timeszone info.
I can revert it.
But yes this is a debugging tool

web/domains/workbasket/actions/applicant_actions.py
MattHolmes123: I agree with you but I resisted such changes to make it easier to change later.  
I tried to be like that throughout

PR #714, opened 2022-01-27
Data Migration Import Reference Data

mikeylockwood:
* Management command `export_from_v1` runs queries from `data_migration/queries` and outputs the data into `data_migration/models`
* Management command `import_v1_data` gets data from `data_migration/models` and imports into `web/models`


pii-ner-exclude.txt
```diff
@@ -188,7 +188,7 @@ Import Licence - Edit Goods
 Import Licence - Edit Certificate
 Importer Edit
 non-European
-| Field         | Value        
+| Field         | Value
```
MattHolmes123, line 191: Has your editor stripped these?
because they are part of what the tool matches and shouldn't be stripped

Makefile
```diff
@@ -72,7 +77,6 @@ check-staging-with-deploy: ## run Django check for staging environment settings
 	export DATABASE_URL="unset" && \
 	docker-compose run --rm web python ./manage.py check --deploy --settings=config.settings.staging
 
-COMMAND="help"
 manage: ## execute manage.py
```
MattHolmes123, line 80: This manage command just doesn't work does it?  
might as well remove the whole thing

mikeylockwood: It does work now, it was just that the command was always being set to `help` (which is the default anyway). To run

`COMMAND=some_command make manage`

data_migration/models/reference.py
```diff
@@ -0,0 +1,128 @@
...
+        data["unit_id"] = self.unit and self.unit.pk
+        return data
+
+
+class CommodityGroupCommodity(MigrationBase):
+    commodity = models.ForeignKey(Commodity, on_delete=models.CASCADE)
+    commoditygroup = models.ForeignKey(CommodityGroup, on_delete=models.CASCADE)
+
+
+class Usage(MigrationBase):
```
MattHolmes123, line 122: I don't see Usage being exported anywhere

data_migration/management/commands/export_from_v1.py
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        with cx_Oracle.connect(**connection_config) as connection:
+            cursor = connection.cursor()
+            self._export_reference_data(cursor, batchsize)
+            cursor.close()
+
+    def _format_data(self, column: str, data: Any) -> tuple[str, Any]:
+        if column.endswith("_datetime") and data:
+            data = timezone.utc.localize(data)
```
MattHolmes123, line 44: Is this correct?
```
>>> from django.utils.timezone import *
>>> import datetime
>>> dt_val = datetime.datetime(2013, 4, 26, 13, 55, 4)
>>> utc.localize(dt_val)
datetime.datetime(2013, 4, 26, 13, 55, 4, tzinfo=<UTC>)
>>> get_default_timezone().localize(dt_val)
datetime.datetime(2013, 4, 26, 13, 55, 4, tzinfo=<DstTzInfo 'Europe/London' BST+1:00:00 DST>)
>>> get_current_timezone().localize(dt_val)
datetime.datetime(2013, 4, 26, 13, 55, 4, tzinfo=<DstTzInfo 'Europe/London' BST+1:00:00 DST>)
>>> 
```

MattHolmes123: Importing with a different timezone to what icms has as the default timezone?

data_migration/models/base.py
```diff
@@ -0,0 +1,14 @@
...
+
+class MigrationBase(models.Model):
+    class Meta:
+        abstract = True
+
+    def excludes(self) -> list[str]:
+        return ["_state"]
+
+    def data_export(self) -> dict[str:Any]:
+        return {k: v for k, v in self.__dict__.items() if k not in self.excludes()}
```
MattHolmes123, line 14: As discussed call self.excludes once

data_migration/models/reference.py
```diff
@@ -0,0 +1,128 @@
...
+    start_datetime = models.DateTimeField(null=True)
+    end_datetime = models.DateTimeField(null=True)
+
+    def excludes(self) -> list[str]:
+        return super().excludes() + ["status"]
+
+    def data_export(self) -> dict[str, Any]:
+        data = super().data_export()
+        data["is_active"] = self.status.lower() == "active"
+        data["commodity_type_id"] = self.commodity_type.pk
```
MattHolmes123, line 89: I think you may need to consider adding a select_related to the queries at some point when importing.

data_migration/management/commands/import_v1_data.py
```diff
@@ -0,0 +1,45 @@
...
+
+        batch_size = options["batchsize"]
+        self._import_reference_data(batch_size)
+
+    def _import_reference_data(self, batch_size):
+        self.stdout.write("Importing Reference Data...")
+        for st in ref_source_target:
+            self.stdout.write(f"Importing {st.target.__name__} from {st.source.__name__}")
+            st.target.objects.bulk_create(
+                [st.target(**obj.data_export()) for obj in st.source.objects.all()],
```
MattHolmes123, line 32: `st.source.objects.all()`
at some point consider a `st.source.get_source_data()` method or something.
That way you could call selected related where needed.
I added an example

data_migration/models/reference.py
mikeylockwood: I was thinking about ways to do this. I will try to add something in ICMSLST-1422

data_migration/management/commands/import_v1_data.py
mikeylockwood: Good idea, will look at changing this in ICMSLST-1422

PR #715, opened 2022-01-31
ICMSLST-1402 Bump pii-secret-check-hooks to version 0.0.0.34

davidwtbuxton:
This version has a change to how Python source code is scanned, which
we hope will result in fewer false positives.


davidwtbuxton:
Done as part of #711

PR #719, opened 2022-02-04
ICMSLST-1421: Move test data creation out of migrations

mikeylockwood:
* Moved data being added out of migrations into `web/management/commands/utils/`
* `web/management/commands/utils/load_data.py` is used to load all the data in order
* Updated `add_dummy_data` and `add_test_data` to call `load_data()`


davidwtbuxton (approved):

web/management/commands/utils/load_data.py
```diff
@@ -0,0 +1,57 @@
...
+    add_email_templates()
+    add_letter_templates()
+    add_letter_fragment_templates()
+    add_export_application_type_data()
+    add_import_application_type_data()
+    add_units()
+    add_commodity_types()
+    add_commodity_groups()
+    add_commodities()
+    add_commodities_to_commity_groups()
```
MattHolmes123, line 53: A typo (I know it was Kevin not you) :)

pii-ner-exclude.txt
```diff
@@ -1001,3 +1001,20 @@ Application Authorisation
...
+docker_flake8
+SQL
+python manage.py npm &&
+shell
+##
+@unset UID && \
+PGPASSWORD
+--rm
+release_patch
+test_agent_export_user
```
MattHolmes123, line 1020: I don't understand why all these are now false positives :/ 
All you are doing is moving files around.

web/management/commands/utils/load_data.py
```diff
@@ -0,0 +1,57 @@
...
+    add_declaration_templates,
+    add_email_templates,
+    add_endorsement_templates,
+    add_letter_fragment_templates,
+    add_letter_templates,
+    add_schedule_translation_templates,
+)
+
+
+def load_data():
```
MattHolmes123, line 34: The name is a bit vague.
`load_app_test_data` maybe?

pii-ner-exclude.txt
mikeylockwood: I guess because no changes to the `Makefile` were made since this was added?

MattHolmes123: I ran this after David's update to the pii tool `pre-commit run --all-files`.
So I don't understand how that would matter

PR #720, opened 2022-02-08
ICMSLST-1422 Export Import Application Type data from V1

mikeylockwood:


data_migration/management/commands/export_from_v1.py
```diff
@@ -50,10 +70,19 @@ def _format_data(self, column: str, data: Any) -> tuple[str, Any]:
...
-        self.stdout.write("Exporting Reference Data...")
+    def _export_data(
+        self, data_type: DATA_TYPE, cursor: cx_Oracle.Cursor, batchsize: int, skip: bool
+    ) -> None:
+        query_models = DATA_TYPE_QUERY_MODEL[data_type]
+        name = " ".join(dt.capitalize() for dt in data_type.split("_"))
+
+        if skip:
+            self.stdout.write(f"Skipping {name} Data Export\n")
+            return None
```
MattHolmes123, line 81: The function never returns a value right?
Just use `return` then

data_migration/management/commands/import_v1_data.py
```diff
@@ -1,9 +1,23 @@
...
+from data_migration.queries import ia_source_target, ref_m2m, ref_source_target
+
+DATA_TYPES = Literal["reference", "import_application"]
+
+DATA_TYPE_SOURCE_TARGET = {
+    "reference": ref_source_target,
+    "import_application": ia_source_target,
+}
+
+DATA_TYPE_M2M = {
```
MattHolmes123, line 17: Would this add clarity? 
`DATA_TYPE_M2M: dict[str, list[M2M]]` = {}

MattHolmes123: Hmm maybe not... Just a thought :thinking:

data_migration/management/commands/import_v1_data.py
```diff
@@ -16,34 +30,66 @@ def add_arguments(self, parser: argparse.ArgumentParser):
...
-        self._import_reference_data(batch_size)
 
-    def _import_reference_data(self, batch_size):
-        self.stdout.write("Importing Reference Data...")
-        for st in ref_source_target:
+        self._import_data("reference", batch_size, options["skip_ref"])
+        self._import_data("import_application", batch_size, options["skip_ia"])
+
+    def _import_data(self, data_type: DATA_TYPES, batch_size: int, skip: bool) -> None:
+        sts = DATA_TYPE_SOURCE_TARGET[data_type]
```
MattHolmes123, line 54: could we have a more descriptive variable name instead of `sts`?

data_migration/management/commands/import_v1_data.py
```diff
@@ -16,34 +30,66 @@ def add_arguments(self, parser: argparse.ArgumentParser):
...
+        m2ms = DATA_TYPE_M2M[data_type]
+        name = " ".join(dt.capitalize() for dt in data_type.split("_"))
```
davidwtbuxton, line 56: A comment would be useful here, i.e. `foo_bar -> Foo Bar`.

davidwtbuxton (approved):

data_migration/models/base.py
```diff
@@ -7,9 +7,44 @@ class MigrationBase(models.Model):
...
+    def get_includes() -> list[str]:
+        return []
+
+    @classmethod
+    def get_values(cls) -> list[str]:
+        excludes = cls.get_excludes()
+        values = [
+            f"{f.name}_id" if f.is_relation else f.name
+            for f in cls._meta.fields
+            if (f.is_relation and f"{f.name}_id") or f.name not in excludes
```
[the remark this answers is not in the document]
MattHolmes123: :bug: :lady_beetle:

data_migration/models/base.py
```diff
@@ -7,9 +7,44 @@ class MigrationBase(models.Model):
...
+        ]
+        return values + cls.get_includes()
+
+    @classmethod
+    def get_related(cls) -> list[str]:
+        includes = cls.get_includes()
+        return [i.split("__")[0] for i in includes]
+
+    @classmethod
+    def get_source_data(cls) -> Generator:
```
MattHolmes123, line 47: This is a lot of generic code that will only get more complex as you implement it for every model we are importing.
Can you add some tests to make sure the correct fields / related modes are included & excluded etc?

I honestly prefer a much more simple approach of every import model defining `data_export` and `get_source_data` methods if im being honest.
We are only abstracting away django orm code here.
But I'm willing to just go with the flow if it has some tests.
As we discussed offline we found a bug in the get_values method

MattHolmes123 (approved):

PR #721, opened 2022-02-08
ICMSLST-1382 Add preview licence for FA-OIL application.

MattHolmes123:
Notes:
- The css and template will probably be changed when I do the next firearms certificates. So I have resisted refactoring it (as its probably going to change in the next pr)
- The endorsement splitting isn't perfect and I need to come back to it (I will create a story for this)
- I've left in my html query param as its useful when creating the pdfs


web/domains/case/views/views_pdf.py
```diff
@@ -43,6 +43,12 @@ def get(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRespons
             request=self.request,
         )
 
+        # TODO: Remove this when all the pdfs have been created
+        # Useful when debugging pdf layout
+        if self.request.GET.get("html"):
```
davidwtbuxton, line 48: Alternatively, `if "html" in self.request.GET`, if one doesn't care what the value is (or even no value, i.e. `/foo/?html`.

web/templates/pdf/import/fa-oil-licence.html
```diff
@@ -0,0 +1,111 @@
...
+
+  <div class="flex-row" id="reference-and-licence">
+    <div></div>
+    <div class="bordered">
+      <p class="content">Applicant's own reference</p>
+      <p class="content">{{ applicant_reference }}</p>
+    </div>
+
+    <div>
+      <p class="content">Quote both reference in all correspondence</p>
```
davidwtbuxton, line 24: Should `reference` be plural?

web/utils/pdf/utils.py
```diff
@@ -1 +1,64 @@
...
+    return "Licence End Date not set"
+
+
+def _get_importer_eori_numbers(application) -> list[str]:
+    # TODO: Check the Country of Consignment logic for other firearm licence types
+    # TODO: If the applicant’s address has a BT (Belfast) post code AND the Country of Consignment is an EU country:
+
+    importer = application.importer
+    office = application.importer_office
+    is_ni = office.postcode.upper().startswith("BT")
```
davidwtbuxton, line 53: A more verbose name here would be helpful, like `is_belfast` or `is_northern_ireland`. I would not expect all devs to know Belfast is in Northern Ireland.

web/tests/utils/pdf/conftest.py
```diff
@@ -0,0 +1,56 @@
...
+        "licence_end_date": "Licence End Date not set",
+        "licence_number": "[[Licence Number]]",
+        "eori_numbers": ["GB123456789"],
+        "importer_address": ["22 Some Avenue", "Some Way", "Some Town"],
+        "importer_postcode": "S93bl",  # /PS-IGNORE
+        "endorsements": [],
+    }
+
+
+def compare_expected_and_actual(expected: dict[str, Any], actual: dict[str, Any]) -> None:
```
davidwtbuxton, line 49: This seems worse than `assert dictA == dictB` which would show all the differences. Or does pytest not show a diff like unittest.assertDictEqual?

davidwtbuxton (approved):

web/templates/pdf/import/fa-oil-licence.html
MattHolmes123: yes good spot!

mikeylockwood (approved):

web/utils/pdf/utils.py
mikeylockwood: I would be happier with `is_ni` than either of those. BT is the postcode for all NI addresses, so not just Belfast

web/tests/utils/pdf/conftest.py
MattHolmes123: This is worse you're right.
It was useful originally (as I wasn't comparing all the keys only a subset.
But after I changed that it makes no sense to keep this.
I will remove it thanks

PR #727, opened 2022-02-16
ICMSLST-1357 Import Textiles and Wood data from V1

mikeylockwood:
* Export of Textiles and Wood application data from V1
* Handles creation of `Process` and  `ImportApplication` models when importing into V2
* Creates dummy user data to associate with the applications prior to user migration
* Includes the creation of checklists

* Excludes the creation of `Task` models


data_migration/management/commands/import_v1_data.py
```diff
@@ -50,31 +58,31 @@ def _import_data(self, data_type: DATA_TYPE, batch_size: int, skip: bool) -> Non
 
         self.stdout.write(f"Importing {name} Data")
 
-        for st in source_target_list:
-            self.stdout.write(f"Importing {st.target.__name__} from {st.source.__name__}")
-            objs = st.source.get_source_data()
+        for source_target in source_target_list:
```
MattHolmes123, line 61: This could be `for source, target in source_target_list` I think

data_migration/management/commands/import_v1_data.py
```diff
@@ -50,31 +58,31 @@ def _import_data(self, data_type: DATA_TYPE, batch_size: int, skip: bool) -> Non
...
             while True:
-                batch = [
-                    st.target(**st.source.data_export(obj)) for obj in islice(objs, batch_size)
-                ]
+                batch = [target(**source.data_export(obj)) for obj in islice(objs, self.batch_size)]
                 if not batch:
                     break
 
-                st.target.objects.bulk_create(batch)
+                # TODO: Increment the db sequence to fix auto increment
```
MattHolmes123, line 71: is this something you still need to do?

data_migration/management/commands/export_from_v1.py
```diff
@@ -78,17 +87,143 @@ def _export_data(
...
+    def _format_data(self, column: str, data: Any) -> tuple[str, Any]:
+        """Applies formatting to the columns to be able to import into Django models
+
+        :param column: The column name being formatted
+        :param data: The data being formatted
+        """
+
+        if column.endswith("_datetime") and data:
+            # TODO: Check timezone how timezones work in django.
+            # Assumption that source is UTC and datetime is passed to models with source tz
```
davidwtbuxton, line 157: I'm not clear whether the `data` datetime object here is expected to be time zone-aware or -naive before we apply `timezone.utc.localize(..)` to it.

data_migration/management/commands/import_v1_data.py
mikeylockwood: Ah no not there. `objects.bulk_create` will handle that

data_migration/management/commands/export_from_v1.py
```diff
@@ -78,17 +87,143 @@ def _export_data(
...
+        :param row: The row of data being formatted
+        :param includes: The fields to be used when creating the model instance
+        :param pk: The pk to set when creating the model instance
+        """
+
+        data = dict(
+            self._format_data(columns[i], data)
+            for i, data in enumerate(row)
+            if not includes or columns[i] in includes
+        )
```
davidwtbuxton, line 181: I think this would be easier to read as a basic for loop. Am I reading it right in that we are trying to pair each item in `row` with each item in `includes`? If so, use `for colname, value in zip(columns, data)`.

davidwtbuxton: Is the following equivalent?

```
data = {}

for colname, value in zip(columns, row):
    key, value = self._format_data(colname, value)
    data[key] = value

if includes:
    # Remove keys that are not in includes.
    data = {k: data[k] for k in data if k in includes}```

**Review by davidwtbuxton — APPROVED** [review]

File: data_migration/management/commands/utils/db.py [hunk 5]
```diff
@@ -0,0 +1,18 @@
+from django.db import connections
+from django.db.models import Model, sql
+
+
+def bulk_create(model: Model, items: list[Model], db: str = "default") -> None:
```
MattHolmes123, line 5: As discussed I think we should document this and explain its limitations (just link the stack overflow)

data_migration/management/commands/export_from_v1.py
mikeylockwood: The source database is currently in `UTC` timezone but the datetimes don't have timezone information when pulling. I am adding the timezone information to them before passing to the django models

MattHolmes123:
test: ## run tests (circleci; don't use locally as it produces a coverage report)
	./run-tests.sh \
		--cov=web \
		--cov=config \
		--cov-report xml:test-reports/cov.xml \
		--numprocesses 2 \
		--cov-fail-under 68

Michael can you update the test command to include `data_migration` in the coverage report  please.
You will probably need to update `.coveragerc` to exclude the sql files but we could do with knowing how much of the migration code is covered by tests

data_migration/management/commands/export_from_v1.py
mikeylockwood: It looks the same. I can meet you in the middle with

```
        data = dict(
            self._format_data(column, data)
            for column, data in zip(columns, row)
            if not includes or column in includes
        )
```

MattHolmes123 (approved):

data_migration/management/commands/export_from_v1.py
davidwtbuxton: Spreading generator / list comprehensions across multiple lines makes them harder to read and is more verbose, and in this case one has to also parse the `not includes or column in includes`. I think doing less in each step is simpler to understand. Another version:

```
data = dict(self._format(column, data) for column, data in zip(columns, row))

if includes:
    # Remove keys that are not in includes.
    data = {k: data[k] for k in data if k in includes}
```

mikeylockwood: You're killing me here. Comprehensions are my favourite thing about python. I want to avoid formatting data that won't be included

```
        data = {}

        for column, value in zip(columns, row):
            if includes:
                if column in includes:
                    k, v = self._format_data(column_value)
                    data[k] = v
            else:
                k, v = self._format_data(column_value)
                data[k] = v
```

PR #728, opened 2022-02-17
ICMSLST-1409 Add preview and pre-sign versions of FA SIL licence.

MattHolmes123:
Other changes:

- Exclude "if TYPE_CHECKING" blocks from coverage report
- Add minimum coverage check to CI pipeline

### DIFF
--- a/web/tests/application_utils.py
+++ b/web/tests/application_utils.py
@@ -8,6 +8,7 @@ from pytest_django.asserts import assertRedirects
 
 from web.domains.case._import.fa_dfl.models import DFLApplication
 from web.domains.case._import.fa_oil.models import OpenIndividualLicenceApplication
+from web.domains.case._import.fa_sil.models import SILApplication
 from web.domains.case._import.wood.models import WoodQuotaApplication
 from web.domains.case.export.models import (
     CertificateOfManufactureApplication,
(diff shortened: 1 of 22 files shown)


web/tests/application_utils.py
```diff
@@ -265,4 +349,7 @@ def submit_app(*, client: "Client", view_name: str, app_pk: int) -> None:
     form_data = {"confirmation": "I AGREE"}
     response = client.post(submit_url, form_data)
 
+    if response.status_code == 200:
+        print(response.context["errors"])
```
davidwtbuxton, line 353: Is this left over from earlier development work? Or should it be an assert?

mikeylockwood (approved):

web/utils/pdf/utils.py
```diff
@@ -1,13 +1,26 @@
...
+    from web.domains.case._import.fa_dfl.models import DFLApplication
+    from web.domains.case._import.fa_oil.models import OpenIndividualLicenceApplication
+    from web.domains.case._import.fa_sil import models as sil_models
+    from web.models import ImportApplication
+
+    SILGoods = Union[
+        sil_models.SILGoodsSection1,
+        sil_models.SILGoodsSection2,
+        sil_models.SILGoodsSection5,
+        sil_models.SILGoodsSection582Obsolete,  # /PS-IGNORE
```
davidwtbuxton, line 17: Yeah, I'm guessing this is the post code PII checker, which does not do the Python tokenizing but instead scans the entire file as text.

davidwtbuxton (approved):

web/tests/application_utils.py
MattHolmes123: Its just useful to see when the application has an error.
the next line asserts what it should do when its working correctly:
`assertRedirects(response, reverse("workbasket"), 302)`

MattHolmes123: I'm leaving it in as I always add it when I'm creating a new application type (to use in tests).

web/utils/pdf/utils.py
MattHolmes123: :cry:

PR #730, opened 2022-02-18
ICMSLST-1223/1224 Add licence reference for import and export applications.

MattHolmes123:


mikeylockwood (approved):

davidwtbuxton (approved):
I feel like the length of the sequence padding is a bit magic here, 7 digits for import, 5 for export, and it may be useful to have a comment pointing out this detail of the 2  licence reference formats.

PR #732, opened 2022-02-21
ICMSLST-1405 Licence & Certificate reference changes

MattHolmes123:


web/domains/case/_import/models.py
```diff
@@ -368,6 +359,33 @@ def licence_issue_date(self):
     def get_specific_model(self) -> "ImportApplication":
         return super().get_specific_model()
 
+    def get_most_recent_licence(self) -> "ImportApplicationLicence":
+        return self.licences.filter(
+            status__in=[
+                ImportApplicationLicence.Status.DRAFT,
+                ImportApplicationLicence.Status.ACTIVE,
+            ]
+        ).order_by("-created_at")[0]
```
mikeylockwood, line 368: I tend to use `.first()` instead of [0] for django queries

MattHolmes123: I do when it's optional.
its a bug if there isn't one and I want the system to fail

web/templates/web/domains/case/import/manage/preview-licence.html
```diff
@@ -136,8 +136,9 @@
             <br />
             <br />
             This licence will cease to be valid for importations after
-            {% if process.licence_end_date %}
-              {{ process.licence_end_date.strftime("%d %B %Y") }}.
+            {% if {{ licence_end_date }} %}
+              {# {{ licence_end_date.strftime("%d %B %Y") }} .#}
```
mikeylockwood, line 140: Commented out?

MattHolmes123: Yes for now.
preview-licence.html is the old template that is a load of rubbish.
It is only used so preview licence doesn't error for applications that haven't implemented it yet.

mikeylockwood (approved):

PR #734, opened 2022-03-03
ICMSLST-1441 Add import / export document reference model.

MattHolmes123:
Changes:
  - Add function to generate import licence document reference.
  - Add function to gererate export certificate document reference.
  - Refactor all reference code to reference service.
  - Add tests for settings document references.
  - ICMSLST-1462 Fix cancel application bug.


web/domains/case/services/reference.py
```diff
@@ -1,46 +1,119 @@
...
+
+    if ref == application.DEFAULT_REF:
+        raise ValueError("Application has not been assigned yet.")
+
+    variations = application.variation_requests.all()
+
+    if not application.is_import_application():
+        # Can't import VariationRequest due to circular dependency
+        # Should use VariationRequest.OPEN & VariationRequest.CLOSED
+        variations = variations.filter(status__in=["OPEN", "CLOSED"])
```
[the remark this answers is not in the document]
MattHolmes123: I'd prefer to move this out of the model that causes the circular import:
image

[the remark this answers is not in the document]
MattHolmes123: I think its a small price to pay at the moment to have all the reference code in the same service.
If we did refactor submit_application elsewhere (which we should) the problem would go away.

web/domains/case/services/reference.py
```diff
@@ -58,11 +131,83 @@ def get_next_reference(
...
+    Reference formats:
+        - XXX/YYYY/NNNNN
+
+    Reference breakdown:
+        - XXX: licence category
+        - YYYY: year certificate issued
+        - NNNNN: Next sequence value (padded to 5 digits)
+    """
+
+    today = datetime.date.today()
```
mikeylockwood, line 208: Instead of importing both `datetime` and `timezone`, you can get date from `timezone.now().date()`
Or even just `timezone.now().year` as that's all it's being used for

MattHolmes123: Yer I will use timezone.
Tbh I just forget about it normally

web/domains/case/export/models.py
```diff
@@ -183,6 +184,14 @@ def get_status_display(self):
 
         return super().get_status_display()
 
+    def get_most_recent_certificate(self) -> "ExportApplicationCertificate":
+        return self.certificates.filter(
+            status__in=[
+                ExportApplicationCertificate.Status.DRAFT,
+                ExportApplicationCertificate.Status.ACTIVE,
+            ]
+        ).order_by("-created_at")[0]
```
davidwtbuxton, line 193: Is it worth using `.latest('created_at)` here in place of `.order_by('-created_at')[0]`? Semantics may be slightly different, in that the docs say `latest()` raises `DoesNotExist` if there is no latest.

davidwtbuxton (approved):

web/domains/case/export/models.py
MattHolmes123: Agreed!

PR #735, opened 2022-03-04
ICMSLST-1452 Creation of task objects during migration

mikeylockwood:


MattHolmes123 (approved):
I take it you will revisit this when importing applications in more complex states?
image

mikeylockwood:
> I take it you will revisit this when importing applications in more complex states? image

Yes I will raise tickets for the other task types. It's not possible to determine them yet with the limited data that has been migrated over at this early stage. This initial PR is just a method of being able to create the tasks during the migration process

data_migration/models/task.py
```diff
@@ -0,0 +1,44 @@
+from data_migration.models import ImportApplication
+from web.models import Task
+
+
+class TaskBase:
+    TASK_TYPE: str = ""
+    TARGET = Task
+
+    @classmethod
+    def obj_dicts(cls) -> list[Task]:
```
davidwtbuxton, line 10: Better name for this method? It doesn't appear to return dicts.

davidwtbuxton (approved):

data_migration/models/task.py
mikeylockwood: Good point, I changed what it does and didn't change the name

PR #743, opened 2022-03-10
ICMSLST-1476 Update licence/certificate when a variation is requested.

MattHolmes123:
Changes:
  - Create a draft licence/certificate when opening a VR.
  - Archive the licence/certificate when cancelling a VR.
  - Filter out archived licences from search results.
  - Create an export licence certificate when creating an export app.
  - Archive old licence when completing a case
  - Set completion date when completing an application.

### DIFF
--- a/web/domains/case/views/views_misc.py
+++ b/web/domains/case/views/views_misc.py
@@ -12,14 +12,12 @@ from django.views.decorators.debug import sensitive_post_parameters
 from django.views.decorators.http import require_POST
 from guardian.shortcuts import get_users_with_perms
 
-from web.domains.case._import.models import (
-    ImportApplicationLicence,
-    get_paper_licence_only,
-)
+from web.domains.case._import.models import ImportApplicationLicence
 from web.domains.case.models import CaseDocumentReference, VariationRequest
 from web.domains.case.services import reference
 from web.domains.case.shared import ImpExpStatus
 from web.domains.case.utils import (
+    archive_application_licence_or_certificate,
     check_application_permission,
     create_acknowledge_notification_task,
     get_application_current_task,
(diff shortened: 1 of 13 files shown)


mikeylockwood (approved):

web/domains/case/views/views_misc.py
```diff
@@ -461,12 +450,18 @@ def create_application_document_references(
         if licence.status != ImportApplicationLicence.Status.DRAFT:
             raise ValueError("Can only create references for a draft application")
 
-        licence.document_references.create(document_type=CaseDocumentReference.Type.COVER_LETTER)
-        licence.document_references.create(
-            document_type=CaseDocumentReference.Type.LICENCE,
-            reference=reference.get_import_licence_reference(lock_manager, application),
+        licence.document_references.get_or_create(
```
davidwtbuxton, line 453: Is there a unique constraint on this CaseDocumentReference.document_type thing? Just wondering if it matters if we ever had get_or_create resulting in 2 CaseDocumentReference objects being created.

web/domains/case/views/views_misc.py
```diff
@@ -461,12 +450,18 @@ def create_application_document_references(
...
+            document_type=CaseDocumentReference.Type.COVER_LETTER
         )
 
+        if not licence.document_references.filter(
+            document_type=CaseDocumentReference.Type.LICENCE
+        ).exists():
```
davidwtbuxton, line 459: Similar to previous comment on the get_or_create, does this need to be bullet proof against concurrency bugs resulting in the CaseReference being created twice?

I see that lock manager thing, which appears to guard against duplicate reference IDs, but it still looks like we could assign 2 CaseReference objects (with different IDs).

davidwtbuxton (approved):

PR #748, opened 2022-03-15
OIL Application Export

mikeylockwood:


data_migration/models/base.py
```diff
@@ -68,13 +77,37 @@ def get_source_data(cls) -> Generator:
...
+        """Parses xml to bulk_create model objects from a list of xml strings
+
+        :param batch: A list of xml strings containing data for populating the model
+        """
+        return [cls.parse_xml_fields(pk, etree.fromstring(xml)) for pk, xml in batch]
+
+    @classmethod
+    def parse_xml_fields(cls, pk: int, xml: etree.ElementTree) -> models.Model:
+        """Retrrives data for fields from an lxml ElementTree object
+
```
mikeylockwood, line 91: More explanation here

web/domains/case/_import/fa/models.py
```diff
@@ -99,8 +99,10 @@ class TransportType(models.TextChoices):
     class Meta:
         abstract = True
 
-    transport = models.CharField(choices=TransportType.choices, max_length=4, blank=False)
-    date_received = models.DateField(verbose_name="Date Received")
+    transport = models.CharField(
```
mikeylockwood, line 102: Check supplementary report to make sure that doesn't affect form validation when saving

PR #751, opened 2022-03-21
ICMSLST-1483 Set export certificate references when start auth is pressed.

MattHolmes123:


web/domains/case/export/models.py
```diff
@@ -528,3 +529,39 @@ class ExportApplicationCertificate(CaseLicenceCertificateBase):
...
+        CaseDocumentReference, on_delete=models.CASCADE, related_name="reference_data"
+    )
+    country = models.ForeignKey(Country, on_delete=models.PROTECT)
+
+    # Extra information for GMP applications
+    gmp_brand = models.ForeignKey("GMPBrand", on_delete=models.PROTECT, null=True)
+
+    def __str__(self):
+        return (
+            f"ExportCertificateCaseDocumentReferenceData("
```
mikeylockwood, line 562: Might be able to get this sting on one line
```
name = self.__name__
cdr_id = self.case_document_reference_id
c_id = self.country_id
gmp_id = self.gmp_brand_id

f"{name}(case_document_reference_id={cdr_id}, country_id={c_id}`, gmp_brand_id={gmp_id)"
```

mikeylockwood (approved):

web/domains/case/export/models.py
```diff
@@ -528,3 +529,39 @@ class ExportApplicationCertificate(CaseLicenceCertificateBase):
...
+    case_document_reference = models.OneToOneField(
+        CaseDocumentReference, on_delete=models.CASCADE, related_name="reference_data"
+    )
+    country = models.ForeignKey(Country, on_delete=models.PROTECT)
+
+    # Extra information for GMP applications
+    gmp_brand = models.ForeignKey("GMPBrand", on_delete=models.PROTECT, null=True)
+
+    def __str__(self):
+        return (
```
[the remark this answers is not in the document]
MattHolmes123: As discussed django plays more nicely when adding the `__str__` method.
I should instead add a __str__ and a __repr__ method.
But as no templates are ever going to print model instances (by design) then I will stick to adding the more useful __str__method.

web/domains/case/export/models.py
MattHolmes123: I did originally have a version quite similar to the above, but then I though its not really saving me anything (one line at most).
So decided to do it the way I did.

PR #755, opened 2022-03-25
ICMSLST-1499 Generate case document pack and upload to s3

MattHolmes123:
Changes:

- Upgrade Celery / Redis to match Cloudfoundry deployment
- Add Check Document Generation view.
- Add Check Document Generation workbasket actions.
- Generate documents when "Authorise Documents" is pressed.


web/domains/file/models.py
```diff
@@ -36,3 +36,9 @@ def date_created_formatted(self):
 
     def human_readable_file_size(self):
         return humanize.naturalsize(self.file_size or 0)
+
+    def __str__(self):
+        props = ("is_active", "filename", "content_type", "file_size", "path")
+        attribute_values = ", ".join([f"{p}={getattr(self, p)!r}" for p in props])
```
davidwtbuxton, line 42: `[]` is optional here, without it would be a generator, i.e. `", ".join(f"{p}" for p in props)`.

web/domains/case/tasks.py
```diff
@@ -0,0 +1,208 @@
...
+        elif document_reference.document_type == CaseDocumentReference.Type.COVER_LETTER:
+            doc_type = DocumentTypes.COVER_LETTER
+            filename = "cover-letter.pdf"
+
+        else:
+            raise ValueError(
+                f"Unable to generate document - unsupported document type {document_reference.document_type}"
+            )
+
+        with io.BytesIO() as file_obj:
```
davidwtbuxton, line 76: I think using `file_obj = io.BytesIO()` and removing the context manager would be more obvious. The code doesn't rely on the file object being closed/flushed.

web/domains/case/tasks.py
```diff
@@ -0,0 +1,208 @@
...
+def upload_case_document_file(
+    file_obj: io.BytesIO, cdr: "CaseDocumentReference", filename: str, created_by: "User"
+) -> None:
+    """Upload a case document file to s3 and create related data."""
+
+    # application id - document type - case document reference id - timestamp - filename
+    time_stamp = f'{timezone.now().strftime("%Y%m%d%H%M%S")}'
+    key = f"{cdr.object_id}_{cdr.document_type}_{cdr.id}_{time_stamp}_{filename}"
+
+    file_size = sys.getsizeof(file_obj.getvalue())
```
davidwtbuxton, line 159: Shouldn't this be the number of bytes written to `file_obj`, rather than the size of the Python object?

Also it may be possible to take advantage of the fact the file pointer hasn't been moved yet, and use `file_obj.tell()` to find how many bytes have been written. This would save reading the contents into memory again.

davidwtbuxton (approved):

web/domains/case/tasks.py
davidwtbuxton: Yee, object overhead:

```
>>> import sys
>>> sys.getsizeof(b'foo')
36
```

MattHolmes123: being honest I wrote this a while a go and forgot to come back to it.
I will test it now

web/domains/case/tasks.py
```diff
@@ -0,0 +1,208 @@
...
+    upload_file_obj_to_s3(file_obj, key)
+
+    cdr.document = File.objects.create(
+        is_active=True,
+        filename=filename,
+        content_type="application/pdf",
+        file_size=file_size,
+        path=key,
+        created_by=created_by,
```
mikeylockwood, line 168: Could `upload_file_obj_to_s3` return a dict of this information to pass into `File.objects.create`

mikeylockwood (approved):

web/domains/case/tasks.py
MattHolmes123: You will see in my next commit :)

docker-compose.yml
```diff
@@ -73,7 +74,7 @@ services:
     celery:
         build:
             context: .
-        command: celery --app=config.celery:app worker --loglevel=debug
+        command: celery --app=config.celery:app worker --loglevel=info
```
davidwtbuxton, line 77: How annoying is debug level?

davidwtbuxton (approved):

docker-compose.yml
MattHolmes123: very

PR #760, opened 2022-04-06
ICMSLST-1505 Firearm OIL User Certificates Data Migration

mikeylockwood:


data_migration/utils/format.py
```diff
@@ -10,22 +11,55 @@ def get_xml_val(xml: etree.ElementTree, xpath: str) -> Any:
...
+
+    if len(date_split) > 3 or len(date_split[-1]) > 4:
+        raise ValidationError("Date not in parsable format")
+
+    if len(date_split[0]) == 4:
+        date_format = "%Y-%m-%d"
+    elif len(date_split[-1]) == 4:
+        date_format = "%d-%m-%Y"
+    else:
+        date_format = "%d-%m-%y"
```
MattHolmes123, line 45: Isn't that the same as this:
```
    if len(date_split[0]) == 4:
        date_format = "%Y-%m-%d"
    else:
        date_format = "%d-%m-%y"
```

MattHolmes123: Nope... I see it now

web/domains/case/_import/fa_sil/views.py
```diff
@@ -306,6 +307,25 @@ def _get_sil_errors(application: models.SILApplication) -> ApplicationErrors:
...
+        )
+
+        create_page_errors(
+            UserImportCertificateForm(
+                data=model_to_dict(cert), instance=cert, application=application
+            ),
+            page_errors,
+        )
+        errors.add(page_errors)
+
```
MattHolmes123, line 328: Does FA DFL have these kind of certificates?

MattHolmes123: Looks like they have certificates but are different to these?

MattHolmes123 (approved):

PR #762, opened 2022-04-11
ICMSLST-1501 & ICMSLST-1502 Application history views for import and export applications

MattHolmes123:
Other changes:
- ICMSLST-1546 Update applicant's View Case view to include extra links.


web/domains/case/views/views_case_history.py
```diff
@@ -0,0 +1,155 @@
...
+            .order_by("-created_at")
+        )
+
+        return {
+            "licences": [
+                {
+                    "case_reference": lic.case_reference,
+                    "variation_request": lic.case_reference.split("/")[3:],
+                    "issue_paper_licence_only": lic.issue_paper_licence_only,
+                    "licence_start_date": lic.licence_start_date.strftime("%d %b %Y"),
```
davidwtbuxton, line 74: This date format (`%d %b %Y`) is used 3 times in this class, make sense to define it once as a class variable and reference that instead?

davidwtbuxton (approved):

web/domains/case/views/views_case_history.py
MattHolmes123: yep good idea.

web/templates/web/domains/case/import/fa/provide-report/report-info.html
```diff
@@ -17,6 +15,24 @@
   </div>
 {% endblock %}
 
+{% block sidebar %}
+    {% if process.supplementary_info.is_complete %}
```
mikeylockwood, line 19: Thanks for adding this. In V1, this can be viewed regardless of whether the report is complete or not

mikeylockwood (approved):

web/templates/web/domains/case/import/fa/provide-report/report-info.html
MattHolmes123: Yep no problem.
The is complete check stops the other links being shown until you've completed it.
The other links being "view application" and "sent responses"

PR #765, opened 2022-04-14
ICMSLST-1289 Update export application search actions logic.

MattHolmes123:


web/utils/search/app_data.py
```diff
@@ -435,6 +435,7 @@ def _add_export_certificate_data(model: "QuerySet[Model]") -> "QuerySet[Model]":
         )
         .annotate(
             latest_certificate_pk=F("certificates__pk"),
+            latest_certificate_issue_date=F("certificates__issue_date"),
```
mikeylockwood, line 438: A bit confused by these F queries? Wouldn't the latest be the max?

MattHolmes123: Not needed, its already grouped by the latest using the distinct on.

davidwtbuxton (approved):

PR #766, opened 2022-04-19
ICMSLST-1454 Add ILB feedback changes for firearms contact list and forms.

MattHolmes123:
Changes:
  - Add Delete action to list view when creating an application.
  - Update redirect links after saving a firearms contact.
  - Add tests for several views.

### DIFF
--- a/web/domains/case/_import/fa/views.py
+++ b/web/domains/case/_import/fa/views.py
@@ -155,22 +155,15 @@ def create_import_contact(
                 _update_know_bought_from(application)
 
                 if application.status == application.Statuses.COMPLETED:
-                    return redirect(
-                        reverse(
-                            "import:fa:provide-report", kwargs={"application_pk": application.pk}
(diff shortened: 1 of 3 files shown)


web/domains/case/_import/fa/views.py
```diff
@@ -155,22 +155,15 @@ def create_import_contact(
...
-                    reverse(
-                        "import:fa:edit-import-contact",
-                        kwargs={
-                            "application_pk": application_pk,
-                            "entity": entity,
-                            "contact_pk": import_contact.pk,
-                        },
+                else:
+                    url = reverse(
+                        "import:fa:list-import-contacts", kwargs={"application_pk": application_pk}
```
mikeylockwood, line 163: Why do we use the `application_pk` variable here but `application.pk` above? Are they the same?

mikeylockwood (approved):

PR #772, opened 2022-04-21
ICMSLST-1460 Move FA-SIL section 5 documents to certificates page.

MattHolmes123:
Changes:
  - Remove section 5 documents from edit view.
  - Add section 5 documents to certificates list view.
  - Update error URL's in submit view to navigate to certificate list view.
  - Delete section 5 documents when submitting if section5 is not selected.

### DIFF
--- a/web/domains/case/_import/fa_sil/views.py
+++ b/web/domains/case/_import/fa_sil/views.py
@@ -261,9 +261,8 @@ def _get_sil_errors(application: models.SILApplication) -> ApplicationErrors:
 
     # Verified Section 5
     if application.section5 and importer_has_section5 and not selected_section5:
-        section_errors = PageErrors(
-            page_name="Application Details - Certificates - Section 5", url=edit_url
-        )
+        url = reverse("import:fa:manage-certificates", kwargs={"application_pk": application.pk})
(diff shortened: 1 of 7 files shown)


web/domains/case/_import/fa_sil/views.py
```diff
@@ -586,6 +577,12 @@ def submit(request: AuthenticatedHttpRequest, *, application_pk: int) -> HttpRes
             form = SubmitForm(data=request.POST)
 
             if form.is_valid() and not errors.has_errors():
+
+                # Check if we need to delete any section five documents
+                if not application.section5:
+                    application.user_section5.clear()
+                    application.verified_section5.clear()
```
mikeylockwood, line 584: Do `user_section5` and `verified_section5` models hold data for just one application per instance or do they for many applications? Calling `clear` will remove the M2M relationship but not the model itself, leaving orphaned data if they have no relationships to any applications. Maybe instead set the `is_active` flag to `False`

mikeylockwood: Actually looking at the code I think these models are used across different applications

mikeylockwood (approved):

PR #775, opened 2022-04-22
Fix bought from JS to show warning when user presses "no".

MattHolmes123:


MattHolmes123:
image
When pressing "no" this message is shown straight away now

web/static/web/js/pages/bought-from.js
```diff
@@ -1,10 +1,10 @@
 window.addEventListener('load', (event) => {
-  const boughtFrom = document.querySelector('#id_know_bought_from');
-  const boughtFromInfoBox = document.querySelector('#know-bought-from-info-box');
+  "use strict";
```
mikeylockwood, line 2: What's this string for?

web/static/web/js/pages/bought-from.js
```diff
@@ -1,10 +1,10 @@
...
 
-  boughtFrom.addEventListener('change', (e) => {
-    const displayStyle = e.target.value == 'false' ? 'block': 'none';
-    boughtFromInfoBox.style.display = displayStyle;
-  });
+  const showBoughtFrom = UTILS.getShowElementFunc("#know-bought-from-info-box")
+  const boughtFromRadios = document.querySelectorAll("input[type=radio][name=know_bought_from]")
 
+  boughtFromRadios.forEach((radio) => radio.addEventListener("change", (e) => {
+    showBoughtFrom(e.target.value === "False")
```
mikeylockwood, line 8: I would `.toLowerCase` this

web/static/web/js/pages/bought-from.js
MattHolmes123: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

web/static/web/js/pages/bought-from.js
MattHolmes123: I mean its the django form value so it will never change but I can do

PR #777, opened 2022-04-25
ICMSLST-1522 Add check for unsaved form data before unloading page.

MattHolmes123:
Changes:

- Add form-data-checker.js to check for form data changes.
- Add form data checking to import applications.
- Add form data checking to export applications.
- Add form data checking to all application sub forms.
- Add Jinja template utility function to return a UUID.


mikeylockwood (approved):

web/static/web/js/components/form-data-checker.js
```diff
@@ -0,0 +1,91 @@
+window.addEventListener('load', (event) => {
+  "use strict";
```
davidwtbuxton, line 2: Should this be at the script level instead of just this anonymous function scope?

davidwtbuxton (approved):

PR #780, opened 2022-04-27
ICMSLST-1550 Fix rejected applications not being returned by search.

MattHolmes123:
Changes:
  - Create a filtered annotation rather than filter applications that  have an archived licence.
  - Fix a bug for export applications causing the search view to  an exception when no manufacturing countries are saved.
  - Renamed a test function to match the view being called.


MattHolmes123:
Here is the sql generated:
image
You can see the filtering is now done on the join and the sub query correctly references that table.

mikeylockwood (approved):

PR #782, opened 2022-04-29
ICMSLST-1558 Add order_datetime column to sort the workbasket.

MattHolmes123:
Changes:
  - Add column that will be used to sort the workbasket.
  - Update multiple views to update the new column.


web/domains/case/utils.py
```diff
@@ -43,6 +43,15 @@ def check_application_permission(application: ImpOrExpOrAccess, user: User, case
         raise PermissionDenied
 
 
+def _has_importer_exporter_access(user: User, case_type: str) -> bool:
```
MattHolmes123, line 46: I moved this to be right next to the only function that called it.

davidwtbuxton: High cohesion, I love it! You deserve a nice cup of tea and a biscuit.

davidwtbuxton (approved):
I see that `application.update_order_datetime()` call followed by `application.save()` in several places. Would it make sense to move that call inside the `application.save()`?

web/domains/case/utils.py
MattHolmes123: :tea: :cookie:

mikeylockwood (approved):

PR #784, opened 2022-05-04
SIL Report Firearms Migration

mikeylockwood:
SIL Supplementary Report Firearms need to be connected to the applications goods which fall under specific sections, therefore the supplementary report firearms need to be created with the correct section model.

A SILSection was created to store the legacy ordinal and section of each goods item as the supplementary reports don't contain any section information in V1. This model is then used to form the correct relationship between the supplementary reports goods and the application goods.


davidwtbuxton (approved):

data_migration/utils/xml_parser/base.py
```diff
@@ -4,7 +4,7 @@
 from django.db.models import Model, QuerySet
 from lxml import etree
 
-BatchT = list[Tuple[int, str]]
+BatchT = list[Tuple]
```
MattHolmes123, line 7: Do you not want to know the types of each list element?
`for parent_pk, xml, section, ordinal in batch:`

MattHolmes123: also you can use `tuple` instead of `Tuple`

MattHolmes123 (approved):

data_migration/utils/xml_parser/base.py
mikeylockwood: I only want that for sil report firearms and the rest  are just `int, str`. It creates a mypy typing mess, so I just removed the specifics of it

data_migration/models/import_application/fa/sil.py
```diff
@@ -121,3 +141,75 @@ class SILSupplementaryReport(SupplementaryReportBase):
...
+
+class SILSupplementaryReportFirearmSection5(SILReportFirearmBase):
+    GOODS_MODEL = SILGoodsSection5
+
+    report = models.ForeignKey(
+        SILSupplementaryReport, related_name="section5_firearms", on_delete=models.CASCADE
+    )
+
+
+class SILSupplementaryReportFirearmSection582Obsolete(SILReportFirearmBase):  # /PS-IGNORE
```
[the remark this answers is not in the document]
mikeylockwood: Unfortunately it does. Stick some numbers between text and it thinks you have a postcode

data_migration/utils/xml_parser/base.py
MattHolmes123: yer ok :+1:

PR #787, opened 2022-05-05
ICMSLST-1518/1521 Add partial saving to firearms forms.

MattHolmes123:


web/domains/case/utils.py
```diff
@@ -215,3 +218,27 @@ def create_acknowledge_notification_task(application: ImpOrExp, previous_task: O
...
+
+    if request.method == "POST":
+        form = edit_form(data=request.POST, instance=application)
+    else:
+        initial = {} if application.contact else {"contact": request.user}
+        form_kwargs = {"instance": application, "initial": initial}
+
+        # query param to fully validate the form.
+        if "validate" in request.GET:
+            form_kwargs |= {"data": model_to_dict(application)}
```
davidwtbuxton, line 239: For a single key, do `form_kwargs["data"] = ..`.

davidwtbuxton (approved):

web/domains/case/utils.py
MattHolmes123: good point. 
I just like |= so damn much :D

mikeylockwood (approved):

PR #808, opened 2022-05-20
ICMSLST-1599 Add confirm message for close case action.

MattHolmes123:


MattHolmes123:
Popup when pressing Stop Case:
image

MattHolmes123:
Success message after closing:
image

mikeylockwood (approved):

web/domains/case/views/views_misc.py
```diff
@@ -352,6 +352,12 @@ def manage_case(
 
                     send_email(subject, body, recipients)
 
+                messages.success(
+                    request,
+                    "This case has been stopped and removed from your workbasket."
+                    " It will still be available from the search screen.",
+                )
+
                 return redirect(reverse("workbasket"))
```
davidwtbuxton, line 361: Not part of this change, I know, but one can use `redirect("workbasket")` directly https://docs.djangoproject.com/en/4.0/topics/http/shortcuts/#redirect

davidwtbuxton (approved):

web/domains/case/views/views_misc.py
MattHolmes123: :+1:

PR #809, opened 2022-05-20
SILUserSection5 and UserImportCertificate migration

mikeylockwood:
* Fixes the `UserImportCertificate` migration for `SILApplication`
* Includes the migration of `SILUserSection5` data - obtained by getting the files linked to the application with the correct folder type and target type
* Change to the `UserImportCertificate` model in web to allow `expiry_date` to be null as there are complete applications in V1 with this
* The submit validation for this model change will already have been handled in https://github.com/uktrade/icms/commit/3e3ee4740


web/templates/partial/certificates/list.html
```diff
@@ -25,7 +25,7 @@
         <td>{{ certificate.get_certificate_type_display() }}</td>
         <td>{{ certificate.constabulary.name }}</td>
         <td>{{ certificate.date_issued and certificate.date_issued.strftime('%d-%b-%Y') }}</td>
-        <td>{{ certificate.expiry_date.strftime('%d-%b-%Y') }}</td>
+        <td>{{ certificate.expiry_date and certificate.expiry_date.strftime('%d-%b-%Y') }}</td>
```
MattHolmes123, line 28: This will render as "None" in the template.
e.g.
```
<P>{{ True and "You see me" }}</P>
<P>{{ None and "You don't see me" }}</P>
```

image

data_migration/models/import_application/fa/base.py
```diff
@@ -100,15 +101,16 @@ def get_source_data(cls) -> Generator:
...
         exclude_query = Q(import_application__submit_datetime__isnull=True) & Q(
-            Q(reference__isnull=True) | Q(constabulary__isnull=True) | Q(expiry_date__isnull=True)
+            Q(reference__isnull=True) | Q(constabulary__isnull=True)
         )
         m2m_id = f"{target._meta.model_name}_id"
         return (
             cls.objects.select_related("target")
             .prefetch_related("target__files")
             .filter(import_application__ima__process_type=target.PROCESS_TYPE)
+            .filter(constabulary__isnull=False)  # TODO One completed record has no constabulary
```
MattHolmes123, line 113: What is the TODO for here?
Is this what you need to speak to Roy about?

mikeylockwood: Yes, Roy confirmed that this User Import Certificate does not need to be migrated with the application

PR #811, opened 2022-05-23
ICMSLST-1611 Add form validation to COM edit form.

MattHolmes123:
Add extra logic to prevent user from saving a value that will never be valid.


MattHolmes123:
Extra validation on Edit form:
image

davidwtbuxton (approved):
I know it was there already, not introduced by this change, but that big block of "safe" html is just waiting for someone to helpfully add some request data and expose a XSS hole.

mikeylockwood (approved):

PR #812, opened 2022-05-24
ICMSLST-1347 Fix edit url not found bug.

MattHolmes123:
Other changes:
  - Order queryset by active status and date created.
  - Add is_active filter to list view


MattHolmes123:
New filter added:
image
New default ordering (active before archived and the newest records appear first):
image

davidwtbuxton (approved):

PR #814, opened 2022-05-24
ICMSLST-1376 / ICMSLST-1587 Updates to FIR / Case Notes and Application Updates

MattHolmes123:
Further information requests, Case Notes and Application Updates should all work in a similar way.
Also updated a load of incorrect `if request.POST` lines.

### DIFF
--- a/web/domains/case/views/views_update_request.py
+++ b/web/domains/case/views/views_update_request.py
@@ -64,7 +64,7 @@ def manage_update_requests(
         email_subject = template.get_title({"CASE_REFERENCE": application.reference})
         email_content = template.get_content(placeholder_content)
 
-        if request.POST and not readonly_view:
+        if request.method == "POST" and not readonly_view:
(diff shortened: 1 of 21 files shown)


MattHolmes123:
New list application update screen: 
image

When clicked:
image

Same screen for Export applications:
image

davidwtbuxton (approved):

web/domains/case/views/views_update_request.py
```diff
@@ -23,11 +23,53 @@
...
 
+@login_required
+@permission_required("web.ilb_admin", raise_exception=True)
+@require_GET
+def list_update_requests(
+    request: AuthenticatedHttpRequest, *, application_pk: int, case_type: str
+) -> HttpResponse:
+    model_class = get_class_imp_or_exp(case_type)
+
+    with transaction.atomic():
```
MattHolmes123, line 34: Actually I don't need to do this as we aren't updating anything yet

PR #818, opened 2022-05-25
ICMSLST-1619 Show DEBUG logging by default

davidwtbuxton:
This is useful for local development. When deployed to Cloud Foundry, we
also use this as a base logging config, except for the "production" env.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -208,26 +208,6 @@ CLAM_AV_DOMAIN = env.str("CLAM_AV_DOMAIN", default="clamav.london.cloudapps.digi
 # Storage Folders
 PATH_STORAGE_FIR = "/documents/fir/"  # start with /
 
-# Structured logging shared configuration
-structlog.configure(
-    processors=[
-        structlog.stdlib.filter_by_level,
-        structlog.processors.TimeStamper(fmt="iso"),
-        structlog.stdlib.add_logger_name,
-        structlog.stdlib.add_log_level,
-        structlog.stdlib.PositionalArgumentsFormatter(),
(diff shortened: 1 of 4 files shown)


davidwtbuxton:
This works! But I don't know if it is what we actually want. Only the `config.settings.production` settings appear to configure the structlog handler, but I think we may want that for `config.settings.development` and `config.settings.staging` too.

The other way of doing this would be to set the full "production" logging config as the base, then override it in `config.settings.local` to be a simple straight-to-console config.

MattHolmes123:
> This works! But I don't know if it is what we actually want. Only the `config.settings.production` settings appear to configure the structlog handler, but I think we may want that for `config.settings.development` and `config.settings.staging` too.
> 
> The other way of doing this would be to set the full "production" logging config as the base, then override it in `config.settings.local` to be a simple straight-to-console config.

I would vote for this, set the logging correct once and simplify it for local development.
Possibly disable it in config.settings.test.py?

What do you think @davidwtbuxton @mikeylockwood ?

config/settings/base.py
```diff
@@ -325,3 +305,52 @@
...
+        },
+    },
+    "loggers": {
+        "django": {
+            "handlers": ["console"],
+            "level": "ERROR",
+        },
+        "web": {
+            "handlers": ["console"],
+            "level": "INFO",
```
davidwtbuxton, line 353: @mikeylockwood Updated to use INFO by default, but DEBUG in local and test.

mikeylockwood (approved):

MattHolmes123 (approved):

PR #819, opened 2022-05-25
ICMSLST-1589 Add default message to several admin management views.

MattHolmes123:


web/templates/partial/case/firs.html
```diff
@@ -25,4 +25,6 @@
 
 {% for fir in firs.all() %}
   {% include 'partial/case/case-fir-box.html' %}
+{% else %}
```
mikeylockwood, line 28: I didn't know you could use `else` on a `for` loop. Nice!

davidwtbuxton (approved):

PR #824, opened 2022-06-01
ICMSLST-1639 Add start point to data migration commands

mikeylockwood:


data_migration/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+The export can also be restarted from a specific point using the `--start` parameter in the format `--start=<data_type>.<index>`
+e.g. `./manage.py export_from_v1 --start=ia.2`  will start the export from point 2 of the import_application data type
+
+The following data_types are valid in the `--start` parameter
+*  `r`, `ref` or `reference` for the reference data type
+* `f` or `file` for the file data type
+* `ia` or `import_application` for the import_application data type
+
+Before new versions of the data migration are deployed, the data base should be wiped and rolled back ready to deploy the latest version
```
MattHolmes123, line 49: database

data_migration/management/commands/import_v1_data.py
```diff
@@ -67,26 +81,37 @@ def handle(self, *args, **options):
         self._create_tasks(options["skip_task"])
 
     def _import_data(self, data_type: DATA_TYPE, skip: bool) -> None:
+        start_m2m = self.start_type.endswith("m2m")
+
+        if self.start_type:
```
MattHolmes123, line 86: You do this block of code three times.  
You could have a base class for the jobs and have it raise a CommandError("Invalid start type") rather than an empty return?

MattHolmes123: A base class to have a function to set the start_type etc.

PR #838, opened 2022-07-01
ICMSLST-1663 Add fa-sil chief serializer.

MattHolmes123:


MattHolmes123:
Example of payload that gets created, the red box is the subsection change that will be changed in ICMSLST-1686:
image

mikeylockwood (approved):

PR #843, opened 2022-07-11
ICMSLST-1659 Store Office address as five fields instead of one.

MattHolmes123:
Changes:
  - Save address as five fields.
  - Fix importer/export save bug that duplicates offices
  - Fix importer/exporter list office queryset showing inactive records.
  - Fix office postcode lookup to store address correctly.


web/templates/tables/tables.html
```diff
@@ -31,12 +31,19 @@
       {{ caller() }}
     {% endwith %}
   {% elif config and config.show_all %}
+
+    {% if config.query_filter %}
+      {% set _objects = object[field].filter(**config.query_filter) %}
+    {% else %}
+      {% set _objects = object[field].all()  %}
```
mikeylockwood, line 38: Very minor but a double space

mikeylockwood (approved):

PR #861, opened 2022-08-15
ICMSLST-1553 Move CFS product form to wrap all form attributes.

MattHolmes123:
Other change:
  - Add extra field to ImportApplicationLicence __str__ method.


web/templates/web/domains/case/export/cfs-edit-product.html
```diff
@@ -31,9 +32,7 @@ <h4>{{ page_title }}</h4>
         </thead>
         <tbody>
           <tr>
-            {% call forms.form(action='', method='post', csrf_input=csrf_input, enctype='multipart/form-data', id='id_cfs-product-form') -%}
```
MattHolmes123, line 34: A form is not allowed to be a child element of a `tr`

mikeylockwood (approved):

PR #872, opened 2022-08-19
ICMSLST-1632 Add SIL legacy goods model to store historical records.

MattHolmes123:


web/domains/case/_import/fa_sil/models.py
```diff
@@ -300,6 +305,18 @@ class IgnitionDetail(models.TextChoices):
     quantity = models.PositiveBigIntegerField(help_text="Enter a whole number")
 
 
+class SilLegacyGoods(models.Model):
```
mikeylockwood, line 308: Should SIL not be capatalised?

mikeylockwood (approved):

web/domains/case/_import/fa_sil/models.py
MattHolmes123: Sure

PR #891, opened 2022-09-15
ICMSLST-1723 Export Application Case Note Data Migration

mikeylockwood:
* Combines case note data from different sources in V1 (export application and import application) into case note model
* Combines files from different folder schemas in V1 into one unified file model in V2


data_migration/models/case.py
```diff
@@ -225,7 +226,8 @@ def get_m2m_data(cls, target: models.Model) -> Generator:
 
 
 class CaseNote(MigrationBase):
-    ima = models.ForeignKey(Process, on_delete=models.CASCADE, to_field="ima_id")
+    ima = models.ForeignKey(Process, on_delete=models.CASCADE, to_field="ima_id", null=True)
+    export_application = models.ForeignKey(ExportApplication, on_delete=models.CASCADE, null=True)
```
MattHolmes123, line 230: will there be an import_application field later?

MattHolmes123 (approved):

data_migration/models/case.py
mikeylockwood: `ima` is the relationship for import applications. `ima_id` is the V1 name for import application's parent which is stored on the `Process` model in data migration tables (`ca_id` is for the export application parent). They are different to `import_application_id` and `export_application_id` which are new ids generated for V2 models. At the point of parsing the xml on the `ExportApplication` model during the data migration process, the `export_application_id` has already been generated so I can FK directly to that. However the case note data for `ImportApplication` is fetched via a query and I only know the `ima_id` and not the generated `import_application_id` at that point.

TLDR; no.

MattHolmes123: Cool :)

PR #907, opened 2022-10-12
ICMSLST-1761 Add time logging to data migration scripts

mikeylockwood:


data_migration/management/commands/export_from_v1.py
```diff
@@ -83,6 +85,8 @@ def _export_data(self, data_type: DATA_TYPE, cursor: cx_Oracle.Cursor, skip: boo
 
                 self._export_model_data(columns, rows, model)
 
+            self._log_time()
```
MattHolmes123, line 88: Do you not want a label to indicate what section you're timing?
Or is this already being logged.

data_migration/management/commands/_base.py
```diff
@@ -55,6 +56,17 @@ def handle(self, *args, **options):
...
+        self.split_time = time.time()
+
+    def _log_time(self) -> None:
+        time_taken = time.time() - self.split_time
+        self.stdout.write(f"\t\t--> {time_taken} seconds", ending="\n\n")
+        self.split_time = time.time()
+
+    def _log_script_end(self) -> None:
+        time_taken = time.time() - self.start_time
+        self.stdout.write(f"Execuction Time --> {time_taken} seconds", ending="\n\n")
```
MattHolmes123, line 70: All seems fair enough.
I added this a while back in to ICMS: 
https://github.com/uktrade/icms/blob/develop/web/utils/__init__.py

It uses perf_counter which is more accurate.

MattHolmes123 (approved):

data_migration/management/commands/export_from_v1.py
mikeylockwood: It always follows a previous log line

```
        7 - Importing ExportApplication_case_emails from CaseEmail
                --> 0.30349302291870117 seconds

        8 - Importing ExportApplication_further_information_requests from FurtherInformationRequest
                --> 0.19294953346252441 seconds

        9 - Importing ExportApplication_update_requests from UpdateRequest
                --> 0.6755120754241943 seconds
```

PR #915, opened 2022-10-25
ICMSLST-1691 SIGL Transmission Data Migration

mikeylockwood:
* Add a model for SIGL Transmissions in V2
* Migrate data to the new model


MattHolmes123:
Could you left align the Response Details content?

data_migration/models/file.py
```diff
@@ -121,7 +121,7 @@ def get_from_combined(cls) -> QuerySet:
 
 
 class FileM2MBase(MigrationBase):
-    TARGET_TYPE: Union[str, list[str]] = "IMP_SUPPORTING_DOC"
+    TARGET_TYPE: str | list[str] = "IMP_SUPPORTING_DOC"
```
MattHolmes123, line 124: What a time to be alive :)

data_migration/tests/utils/ia_data.py
```diff
@@ -1933,4 +1933,74 @@
...
+            ("response_message",),
+            ("response_code",),
+        ],
+        [
+            (
+                100,  # ima_id
+                "ACCEPTED",  # status
+                "WEB_SERVICE",  # transmission_type
+                "INSERT",  # request_type
+                datetime.now(),  # sent_datetime
```
MattHolmes123, line 1954: I'd just get in to the habbit of using django.utils.timezone.now when writing django apps.

https://docs.djangoproject.com/en/4.1/topics/i18n/timezones/#concepts

MattHolmes123:
```def now():
    """
    Return an aware or naive datetime.datetime, depending on settings.USE_TZ.
    """
    return datetime.now(tz=timezone.utc if settings.USE_TZ else None)```

File: web/templates/web/domains/case/manage/manage.html [hunk 3]
```diff
@@ -48,6 +48,58 @@ <h4>SIGL</h4>
           link:</p>
         <p><a href="https://webgate.ec.europa.eu/siglwood" target="_blank">https://webgate.ec.europa.eu/siglwood</a></p>
       </div>
+    {% elif process.application_type.sigl_flag %}
+      <table class="setoutList">
```
MattHolmes123, line 52: Do we need any if of this if there aren't going to be any more sigl transmissions going forward?

It will be empty on all new applications (if any sigl_flag enable app types are live).

MattHolmes123: e.g.
```
{% elif process.application_type.sigl_flag and process.sigl_transmissions.exist() %}
```

MattHolmes123 (commented):
added a few comments.

web/templates/web/domains/case/manage/manage.html
mikeylockwood: ` sigl_flag` is only live for wood applications. This is only to display legacy sigl transmissions for now

data_migration/tests/utils/ia_data.py
mikeylockwood: This is mimicking data that is pulled from the V1 database which has no timezone info associated

PR #926, opened 2022-11-25
ICMSLST-1498 User Data Migration

mikeylockwood:


data_migration/management/commands/export_from_v1.py
```diff
@@ -160,6 +161,11 @@ def _create_user_data(self):
             self.stdout.write("User data exists. Skipping.")
             return
 
+        hasher = FOXPBKDF2SHA1Hasher()
+        salt = hasher.salt()
+        password_str = hasher.encode(password, "2:" + salt)
```
MattHolmes123, line 166: "2:" isn't this supposed to be the user pk or something?

MattHolmes123 (approved):

data_migration/management/commands/export_from_v1.py
mikeylockwood: Yes, the pk is set to 2 when the user is created below

PR #928, opened 2022-11-28
ICMSLST-1793 Add end to end tests using playwright.

MattHolmes123:
Changes:
  - Add a playwright-runner container for executing end to end tests.
  - Refactor end_to_end test directory and tests


web/end_to_end/conftest.py
```diff
@@ -0,0 +1,179 @@
...
+def _get_test_name() -> str:
+    """Get the current test name
+
+    Converts this: test_create_fa_dfl_application.py::test_can_create_fa_dfl[chromium] (call)
+    to this: "test_can_create_fa_dfl"
+    """
+
+    current_test = os.environ.get("PYTEST_CURRENT_TEST", "")
+
+    return current_test.split(":")[-1].split(" ")[0].split("[")[0]
```
mikeylockwood, line 179: Don't know if there is a way to make this less ugly

```
return re.search('::([a-z_]*)\[', current_test).group(1)
```

Or a testable utility function

```
def extract_substr(start, end, string, substr='[a-z_]*'):
    match = re.search(f'{start}({substr}){end}', current_test).group(1)
    return match and match.group(1)
```

mikeylockwood (approved):

web/end_to_end/conftest.py
MattHolmes123: I agree - I googled how to get the current test name and that's what it gave me.

MattHolmes123: test_name.split("::")[-1].split("[")[0]

PR #929, opened 2022-11-29
ICMSLST-1792 Replace dummy user ids with data in data migration

mikeylockwood:


data_migration/management/commands/export_from_v1.py
```diff
@@ -147,37 +141,26 @@ def _extract_file_data(self) -> None:
...
-            is_active=True,
-            title="Mr",
-            password_disposition="FULL",
-            salt=salt,
-            encrypted_password=encrypted_password,
-        )
-        user.save()
+        # TODO ICMSLST-1811
+        models.File.objects.filter(created_by_id__isnull=True).update(created_by_id=0)
+        guest_user = models.User.objects.filter(pk=0).first()
```
MattHolmes123, line 153: If there should be only one guest_user then I would prefer this call `.get()`

data_migration/models/file.py
```diff
@@ -83,13 +83,15 @@ def get_from_combined(cls) -> QuerySet:
 
 
 class File(MigrationBase):
+    # TODO: created_by_str can be an email or id. Need to process and fix data
```
MattHolmes123, line 86: issue number for TODO

data_migration/queries/export_application/export.py
```diff
@@ -121,19 +142,19 @@
 ORDER BY cardc.id
 """
 
-
+# TODO: Check what status in V2 updates closed_by_id
```
MattHolmes123, line 145: issue number for TODO

data_migration/utils/xml_parser/user.py
```diff
@@ -36,10 +36,10 @@ def parse_xml_fields(cls, parent_pk: int, xml: etree.ElementTree) -> Model | Non
 
         status = get_xml_val(xml, "./STATUS")
         request_date = datetime_or_none(get_xml_val(xml, "./REQUEST_DATE"))
-        requested_by = 2  # get_xml_val(xml, "./REQUEST_CREATED_BY_WUA_ID")
-        requested_from = 2  # get_xml_val(xml, "./CONTACT_WUA_ID") ?
+        requested_by = get_xml_val(xml, "./REQUEST_CREATED_BY_WUA_ID")
+        requested_from = get_xml_val(xml, "./CONTACT_WUA_ID")  # TODO check post-migrate
```
MattHolmes123, line 40: issue number for TODO

data_migration/management/commands/export_from_v1.py
mikeylockwood: There is but the guest user doesn't exist in the test data. I thought `.filter()` would be the simplest way to get round it as I'm just going to try and remove the user from the data in the next ticket

PR #937, opened 2022-12-12
ICMSLST-1811 Get file created_by_id from strings

mikeylockwood:


data_migration/queries/import_application/licence.py
```diff
@@ -34,7 +34,12 @@
 #  , dd.signed_datetime
 #  , dd.signed_by_id signed_by_str
 
-ia_licence_docs = """
+ia_licence_docs = r"""
```
mikeylockwood, line 37: The precommit changes this to a raw string

MattHolmes123 (approved):

data_migration/queries/import_application/licence.py
MattHolmes123: https://github.com/asottile/pyupgrade#invalid-escape-sequences

PR #946, opened 2022-12-19
ICMSLST-1828 Add ssl_cert_reqs query param to Redis / Celery connection.

MattHolmes123:


config/settings/base.py
```diff
@@ -211,7 +212,11 @@
 
 # Celery & Redis shared configuration
 if "redis" in VCAP_SERVICES:
-    REDIS_URL = VCAP_SERVICES["redis"][0]["credentials"]["uri"]
+    secure_redis_uri = VCAP_SERVICES["redis"][0]["credentials"]["uri"]
+    parts = urlsplit(secure_redis_uri)
+    ssl_param = "ssl_cert_reqs=required"
+    query_params = ssl_param if not parts.query else f"{parts.query}&{ssl_param}"
```
mikeylockwood, line 218: I would prefer to handle this as a dict

```
qs_dict = parse_qs(parts.query)
qs_dict["ssl_cert_reqs"] = ["required"]
query_params = urlencode(qs_dict)
```

mikeylockwood (approved):

config/settings/base.py
MattHolmes123: I can change it but the value type is list[str] not str

mikeylockwood (approved):

PR #947, opened 2022-12-21
ICMSLST-1827 Rename Permissions to GlobalPermission. 

MattHolmes123:
Other changes:
  - Add ContentType to admin for debugging.
  - Update Permission qs to only return relevant permissions.
  - Restrict permissions available when creating / editing Group records.
  - Simplify logic of assigning / deleting contact permissions.
  - Remove several unused management commands.
  - Remove broken BDD tests.
  - Remove Pa11y.


MattHolmes123:
Example of permission list being restricted to only show permissions currently available:
image

Every model in django gets 4 permissions by default and we don't want to see any of them

MattHolmes123:
Also filter the available permissions that can be added to a group:
image

mikeylockwood (approved):

PR #948, opened 2023-01-03
ICMSLST-1791 Update auto created timestamps post migrate

mikeylockwood:


data_migration/queries/export_application/export.py
```diff
@@ -94,6 +94,7 @@
   , last_updated_by_wua_id last_updated_by_id
   , xcad.variation_number variation_no
   , xcad.submitted_by_wua_id submitted_by_id
+  , ca.created_datetime created
```
MattHolmes123, line 97: How come you decided on "created" as an alias in a few places?  
Just curious more than anything as "created_datetime" seems a better name

data_migration/tests/test_e2e_inactive.py
```diff
@@ -117,7 +117,7 @@ def test_import_export_data(mock_connect, dummy_dm_settings):
     assert opt1.customs_office_address == "Test Address"
     assert opt1.rate_of_yield == 0.5
     assert opt1.rate_of_yield_calc_method == "abc"
-    assert opt1.last_export_day == datetime(2023, 4, 23).date()
+    assert opt1.last_export_day == dt.datetime(2023, 4, 23).date()
```
MattHolmes123, line 120: `dt.date(2023, 4, 23)`

data_migration/utils/xml_parser/import_application.py
```diff
@@ -90,6 +91,7 @@ def parse_xml_fields(cls, parent_pk: int, xml: "ET") -> dm.ImportContact:
                 "region": region,
                 "dealer": str_to_yes_no(dealer),
                 "country_id": country_id,
+                "created_datetime": timezone.now(),
```
MattHolmes123, line 94: Whats this we are defaulting to now?
Is it not present in v1?

web/tests/domains/chief/test_client.py
```diff
@@ -73,7 +73,7 @@ def test_send_application_to_chief(self, fa_sil_app_doc_signing, monkeypatch):
                 "licence_reference": "dummy-reference",
                 "country_code": "DE",
                 "country_group": None,
-                "end_date": "2023-12-01",
+                "end_date": "2024-12-01",
```
MattHolmes123, line 76: Kick it down the road by a year :D 
Love it!

MattHolmes123: In the tests this is the line that sets the licence end date:
`licence_end_date=datetime.date(datetime.date.today().year + 1, 12, 1),`

data_migration/management/commands/import_v1_data.py
```diff
@@ -226,3 +230,29 @@ def _create_missing_export_certificates(self, skip: bool) -> None:
...
+                u_model = getattr(web, model)
+                u_table = getattr(web, model)._meta.db_table
+                pk_u = u_model._meta.pk.column
+                s_model = getattr(dm, model)
+                s_table = s_model._meta.db_table
+                pk_s = s_model._meta.pk.column
+                query = update_sql.format(
+                    u_table=u_table, s_table=s_table, u_field=field, pk_s=pk_s, pk_u=pk_u
+                )
+                cursor.execute(query)
```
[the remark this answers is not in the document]
mikeylockwood: I tried that but it quotes all the params so the sql doesn't work. I added ICMSLST-1840 to just store the queries rather than parameterise them

data_migration/queries/export_application/export.py
mikeylockwood: I agree created_datetime is better

data_migration/utils/xml_parser/import_application.py
mikeylockwood: It's not present under the endorsement data. I had a look at the import application responses model to see if I could get a reasonable datetime from that but it wasn't straight forward

data_migration/management/commands/import_v1_data.py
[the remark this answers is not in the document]
MattHolmes123: cool

data_migration/utils/xml_parser/import_application.py
MattHolmes123: ok cool.
I don't think the created_datetime on the Endorsement model is used anywhere (I haven't checked).

MattHolmes123 (approved):

data_migration/management/commands/print_export_query.py
```diff
@@ -1,18 +1,14 @@
...
-    import_application,
-    reference,
-    user,
-)
+from data_migration import queries
 
 
 class Command(BaseCommand):
-    QUERY_MODULES = [files, import_application, reference, export_application, user]
+    ALL_QUERIES = OrderedDict(getmembers(queries, lambda obj: isinstance(obj, str)))
```
MattHolmes123, line 11: So get members returns (name, value) pairs sorted by name.
Should the predicate lambda have a value param as that is what is it instead of obj?

Just reading here:
https://docs.python.org/3/library/inspect.html#inspect.getmembers

data_migration/queries/import_application/fa/dfl.py
```diff
@@ -1,12 +1,6 @@
-from data_migration.queries.import_application.import_application import (
-    common_xml_fields,
-    import_application_base,
-    import_checklist_base,
-)
 from web.flow.models import ProcessTypes
 
-__all__ = ["dfl_application", "dfl_checklist"]
-
+from ..base import common_xml_fields, import_application_base, import_checklist_base
```
MattHolmes123, line 3: Can we stick to absolute imports or relative to the same module - that's what we've done everywhere in ICMS so far.

MattHolmes123: In fact as its in queries I don't really care here :D 
But in general either the full path or from `.foo import bar`

MattHolmes123 (approved):

data_migration/management/commands/print_export_query.py
mikeylockwood: If you look at the predicates they pass `object` in
https://docs.python.org/3/library/inspect.html#inspect.ismodule

PR #954, opened 2023-01-13
ICMSLST-1843 Add static Firearms Act data to data migration

mikeylockwood:


data_migration/migrations/0002_data_firearms_act.py
```diff
@@ -0,0 +1,53 @@
...
+        "act": "Section 1 Ammunition",
+        "id": 5,
+    },
+    {
+        "act": "Expanding Ammunition 5(1A)(f)",
+        "id": 6,
+    },
+    {
+        "act": "Suppressors",
+        "id": 7,
```
MattHolmes123, line 32: why do you have to specifiy the primary key?

mikeylockwood: Because it referenced by a foreign key in `FirearmsActQuantity`. There are ids in the V1 db (extracted using `ActQuantityParser`) but not the info about the acts they refer to. This is stored in the application xml

MattHolmes123: Yer I wasn't sure if bulk_create allows you to set the primary key of if it was silently ignoring them as the id's you have defined happen to match the order they are being inserted.

PR #955, opened 2023-01-16
ICMSLST-1849 Refactor licence and certificate code 

MattHolmes123:


web/domains/case/services/document_pack.py
```diff
@@ -0,0 +1,405 @@
...
+    pack.status = pack.Status.ARCHIVED
+    pack.save()
+
+
+def pack_active_get(application: ImpOrExp) -> DocumentPack:
+    """Get the current active document pack."""
+
+    packs = _get_qm(application).filter(status=PackStatus.ACTIVE)
+
+    return packs.get()
```
mikeylockwood, line 164: Seen this `filter(param=something).get()` in a couple places in this file. Why not just `.get(param=something)`?

web/domains/case/services/document_pack.py
```diff
@@ -0,0 +1,405 @@
...
+
+if TYPE_CHECKING:
+    from web.utils.lock_manager import LockManager
+
+
+PackStatus = DocumentPackBase.Status
+DocumentType = CaseDocumentReference.Type
+
+# Available service functions
+__all__ = [
```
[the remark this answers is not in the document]
MattHolmes123: Not really - Just exposing what is importable from other modules

web/domains/case/services/document_pack.py
MattHolmes123: Older versions looked like this:
```
def get_something(foo, allow_optional) -> SomeModel | None:
    qs = SomeModel.objects.filter(name=foo)
    
    if allow_optional:
        return qs.first()
    
    return qs.get()
```

MattHolmes123: I can change it

web/domains/case/services/document_pack.py
```diff
@@ -0,0 +1,405 @@
...
+                reference.get_export_certificate_reference(lock_manager, app),
+                country=country,
+            )
+
+    elif application.process_type == ProcessTypes.GMP:
+        gmp_app: CertificateOfGoodManufacturingPracticeApplication = (
+            application.get_specific_model()
+        )
+        countries = gmp_app.countries.all().order_by("name")
+        brands = gmp_app.brands.all().order_by("brand_name")
```
mikeylockwood, line 293: Not a PR without a couple extra `.all()`

mikeylockwood (approved):

web/domains/case/services/document_pack.py
MattHolmes123: updated.

PR #962, opened 2023-01-26
ICMSLST-1888 Fix several bugs relating to refusing an application

MattHolmes123:


web/templates/web/domains/case/manage/prepare-response.html
```diff
@@ -182,6 +182,10 @@ <h4>Endorsement</h4>
         </div>
       {% endif %}
     {% endblock %}
+
+  {% elif readonly_view and process.decision == "REFUSE" %}
```
mikeylockwood, line 186: Can you use `== process.REFUSE` ?

web/templates/web/domains/case/view_case.html
```diff
@@ -17,7 +17,7 @@
         {% if process.is_import_application() %}
           {{ icms_link(request, icms_url('case:history', kwargs={'application_pk': process.pk, 'case_type': case_type}), 'Sent Responses' ) }}
 
-          {% if process.application_type.type == "FA" %}
+          {% if process.application_type.type == "FA" and process.decision == "APPROVE" %}
```
mikeylockwood, line 20: As above, `process.APPROVE`?

mikeylockwood (approved):

PR #977, opened 2023-02-13
ICMSLST-1912 Fix Endorsement Implementation

mikeylockwood:


web/domains/case/_import/models.py
```diff
@@ -109,37 +109,22 @@ class SubTypes(models.TextChoices):
...
-                self.SubTypes.SIL: ProcessTypes.FA_SIL,
-            }
-        }
-
-        if self.type in mapped_types.keys():
-            process_type = mapped_types[self.type]
-
-        elif self.type in mapped_types_with_subtypes.keys():
-            process_type = mapped_types_with_subtypes[self.type][self.sub_type]
+    def __str__(self) -> str:
```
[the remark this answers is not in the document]
MattHolmes123: Ok so I've done some digging and it was to improve the ImportApplicationType select options on the Maintain Commodity Group form (It generates a select list).

Here are the values what used to appear:
- ADHOC (ADHOC1)
- FA (DEACTIVATED)
- FA (OIL)
- FA (SIL)
- IS (QUOTA)
- OPT (QUOTA)
- SAN (SAN1)
- SPS (SPS1)
- TEX (QUOTA)
- WD (QUOTA)

The version I added displayed the values like this:

- Sanctions and Adhoc Licence Application
- Firearms and Ammunition (Deactivated Firearms Licence)
- Firearms and Ammunition (Open Individual Import Licence)
- Firearms and Ammunition (Specific Individual Import Licence)
- Iron and Steel (Quota)
- Outward Processing Trade
- Derogation from Sanctions Import Ban
- Prior Surveillance
- Textiles (Quota)
- Wood (Quota)

Your version displays like this (extra "import" on the DFL record):

- Sanctions and Adhoc Licence Application
- Firearms and Ammunition (Deactivated Firearms Import Licence)
- Firearms and Ammunition (Open Individual Import Licence)
- Firearms and Ammunition (Specific Individual Import Licence)
- Iron and Steel (Quota)
- Outward Processing Trade
- Derogation from Sanctions Import Ban
- Prior Surveillance
- Textiles (Quota)
- Wood (Quota)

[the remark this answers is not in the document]
MattHolmes123: I think this would be a better way of showing labels:
```
    def __str__(self):
        if self.sub_type in self.SubTypes.values:
            return f"{self.get_type_display()} ({self.get_sub_type_display()})"
        else:
            return f"{self.get_type_display()}"
```

It shows labels exactly the same as your version.

web/domains/case/_import/models.py
```diff
@@ -109,37 +109,22 @@ class SubTypes(models.TextChoices):
...
+        # TODO ICMSLST-1918 type should be None if not in Types
+        if self.type in self.Types.values:
+            return self.Types(self.type).label
 
-        return ProcessTypes(process_type).label
+        return f"{self.type} ({self.sub_type})"
 
     class Meta:
+        # TODO ICMSLST-1918 ordering strange as type codes and their labels don't always match alphabetically
+        # Store a label / name field on the model?
```
MattHolmes123, line 127: The order isn't strange, its using the values stored in the db - it doesn't know anything about the labels:

- ADHOC (ADHOC1)
- FA (DEACTIVATED)
- FA (OIL)
- FA (SIL)
- IS (QUOTA)
- OPT (QUOTA)
- SAN (SAN1)
- SPS (SPS1)
- TEX (QUOTA)
- WD (QUOTA)

It's just unfortunate that our values in the db don't match the labels.

MattHolmes123: I suspect all this came about because whoever migrated ImportApplicationType from V1 originally preserved the type / sub_type values from the database, and thats why values like `ADHOC` have labels like `Sanctions`.

I think adding a column to use for ordering this one model is ok.

something like an integer `sort_order` column `ordering = ("sort_order",)`

web/domains/template/forms.py
```diff
@@ -81,15 +81,20 @@ class Meta:
...
+            label="",
+            help_text="""
+                Search an endorsement to add. Endorsements returned are matched against
+                name and content.
+            """,
+            queryset=Template.objects.filter(template_type=Template.ENDORSEMENT).exclude(
+                endorsement_application_types=application_type_pk
+            ),
+            widget=EndorsementTemplateWidget,
+        )
```
MattHolmes123, line 97: Isn't it more standard to define the form field as a class property with `Template.objects.none()` and then in the `__init__` method you override the queryset

e.g.
`self.fields["endorsement"].queryset = ...`

web/management/commands/utils/add_application_type_data.py
```diff
@@ -274,6 +275,7 @@ def add_import_application_type_data():
         importer_printable=False,
     )
 
+    # TODO ICMSLST-1918 Remove from data if not implemented
```
MattHolmes123, line 278: Do these comments mean we are removing ImportApplicationType records that we haven't implemented in V2?

web/domains/template/views.py
```diff
@@ -165,34 +166,43 @@ def form_valid(self, form):
...
+    return render(
+        request,
+        "web/domains/template/list-endorsement-usages.html",
+        {"objects": import_application_types},
+    )
 
 
 @login_required
 @permission_required("web.ilb_admin", raise_exception=True)
 def edit_endorsement_usage(request, pk):
```
MattHolmes123, line 182: Is it worth renaming this now it has changed?
Its really `edit_import_application_type_endorsement` now right?

web/domains/template/views.py
```diff
@@ -165,34 +166,43 @@ def form_valid(self, form):
...
-            usage.linked_endorsements.add(linked_endorsement)
+            endorsement = form.cleaned_data["endorsement"]
+            application_type.endorsements.add(endorsement)
     else:
-        form = EndorsementUsageForm()
+        form = EndorsementUsageForm(application_type_pk=pk)
 
-    context = {"object": usage, "form": form}
+    context = {"object": application_type, "form": form}
     return render(request, "web/domains/template/edit-endorsement-usage.html", context)
```
MattHolmes123, line 193: Same comment with the template name (renaming to reflect what it is)

web/tests/domains/case/services/test_response_preparation.py
```diff
@@ -0,0 +1,68 @@
...
+from web.domains.case.services import response_preparation
+from web.models import (
+    DFLApplication,
+    EndorsementImportApplication,
+    ImportApplicationType,
+    Template,
+)
+
+
+@pytest.mark.django_db
```
MattHolmes123, line 12: I don't think you need this because of the fixtures you're using

web/tests/domains/case/services/test_response_preparation.py
```diff
@@ -0,0 +1,68 @@
...
+def test_add_endorsements_from_applciation_type(test_import_user, importer, office):
```
MattHolmes123, line 13: Spelling error: `applciation`

It appears in two places

web/tests/domains/case/services/test_response_preparation.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    endorsements = Template.objects.filter(template_name__startswith="Test Endorsement")
+    application_type.endorsements.add(*endorsements)
+
+    inactive_endorsement = Template.objects.filter(
+        template_type=Template.ENDORSEMENT, is_active=False
+    ).first()
+
+    application_type.endorsements.add(inactive_endorsement)
```
MattHolmes123, line 25: I assume the test here is suggesting that there are three templates (2 active and 1 inactive)?

Maybe add a comment for clarity

web/utils/pdf/utils.py
```diff
@@ -101,7 +101,7 @@ def get_fa_sil_licence_context(
 def get_licence_endorsements(application: "ImpOrExp") -> list[str]:
     endorsements = [
         content.split("\r\n")
-        for content in application.endorsements.all().values_list("content", flat=True)
+        for content in application.endorsements.values_list("content", flat=True)
```
MattHolmes123, line 104: How dare you :)

web/domains/case/services/response_preparation.py
```diff
@@ -0,0 +1,27 @@
+from web.domains.case._import.models import (
+    EndorsementImportApplication,
```
MattHolmes123, line 2: I don't think `response_preparation` is a good name for the code we're adding here.

I think `endorsement` is better.

But if this is the only code we are going to be adding relating to adding endorsements to an application then what about adding it to `icms/web/domains/case/utils.py` 

e.g.
```
from web.domains.case import utils

utils.add_application_endorsements(application)
```

web/domains/case/_import/models.py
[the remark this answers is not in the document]
mikeylockwood: I don't think either is a good way to show the labels. There shouldn't be any sub type or type in our that are not included in the `SubType` or `Type` enum choices, although in our data there currently is (including in the data being migrated from V2). We need to be clearer on what the types and subtypes in V2 are going to be and not have invalid data, which is why I raised ICMSLST-1918 to look into it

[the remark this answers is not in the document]
MattHolmes123: Clear out the extra sub_type values if you feel the need but they came from v1 and aren't really doing any harm.

web/domains/case/_import/models.py
mikeylockwood: Yes the sloppy ordering has come about because the labels which were stored in the db in V1 are not stored in V2. If adding a column, I would prefer just to add a label or title field

web/management/commands/utils/add_application_type_data.py
mikeylockwood: Yes, or make changes to support them. There's no data for these in V1, so unless they are developed for future I don't see the value of doing extra work for them

web/domains/case/_import/models.py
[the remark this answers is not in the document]
MattHolmes123:
```
>>> from web.models import ImportApplicationType
>>> for iat in ImportApplicationType.objects.all():
...     print(f"{iat.type}, {iat.sub_type}, {str(iat)}")
... 
ADHOC, ADHOC1, Sanctions and Adhoc Licence Application
FA, DEACTIVATED, Firearms and Ammunition (Deactivated Firearms Import Licence)
FA, OIL, Firearms and Ammunition (Open Individual Import Licence)
FA, SIL, Firearms and Ammunition (Specific Individual Import Licence)
GS, CARS, GS
GS, IS, GS
IS, QUOTA, Iron and Steel (Quota)
OPT, QUOTA, Outward Processing Trade
SAN, SAN1, Derogation from Sanctions Import Ban
SPS, SPS1, Prior Surveillance
TEX, QUOTA, Textiles (Quota)
WD, QUOTA, Wood (Quota)
```
I'd just delete the records we haven't implemented and move on with your life :)

web/domains/template/views.py
mikeylockwood: I'd rather keep view / templates / urls consistent to what it seem on the frontend rather than their database tables. These are part of the "maintain endorsement usages" section of the site

web/domains/case/_import/models.py
MattHolmes123: Adding another column to store a label makes the Type and SubType labels pointless though.
A sort_order column has a single purpose and works with everything we currently use?

web/management/commands/utils/add_application_type_data.py
MattHolmes123: Until someone says add a new application type I say 🗑️

web/domains/case/_import/models.py
MattHolmes123: If you don't need the values for type and sub_type for your migration then just change the values in the TextChoice classes to match the labels and then the ordering problem goes away.

You'd just have to go through all the tests to make sure we don't manually create any IAT records or do any filtering based on those values with hard coded strings.

web/domains/case/services/response_preparation.py
mikeylockwood: Yeah, wasn't sure where to put it. There are more functions that will be added regarding templates. I think case/utils is probably better

web/domains/template/views.py
MattHolmes123: Fair enough 👍 I just found it confusing as you'd deleted the usage model.

web/domains/case/_import/models.py
mikeylockwood: A sort order column means that existing rows would need to be updated when new rows are added which is not good practice. We have to expect that there will be more application types in future.

I think those type codes only exist in V1 as unique references for queries. Maybe we don't need them at all in V2. I wouldn't look into that as part of an endorsements template ticket though

mikeylockwood: Rather than making any db changes, could just sort the data in python / in the frontend. It's always going to be a small list of data

MattHolmes123: Also the type and sub_type columns are used in a lot of places in v2 so we can't just get rid easily

mikeylockwood: I have seen an ordering column to sort data the exists in a series for values that code cannot sort ("small", "medium", "large") but have never seen an ordering column used to sort an alphabetical list of independent items, and nor do I wish to.

MattHolmes123 (approved):

web/domains/case/_import/models.py
MattHolmes123: Then you have never lived

PR #978, opened 2023-02-14
ICMSLST-1904 Add revoke licence feature.

MattHolmes123:
This change allows payloads to be sent to icms-hmrc which in turn will forward revoke licence requests to CHIEF.


web/domains/chief/types.py
```diff
@@ -102,15 +116,20 @@ class SanctionGoodsData(BaseModel):
...
     type: Literal["SAN"]
     goods: list[SanctionGoodsData]
 
 
-class CreateLicenceData(BaseModel):
-    licence: FirearmLicenceData | SanctionsLicenceData
+class LicenceDataPayload(BaseModel):
+    """Payload that gets sent to ICMS-HRMC."""
+
+    licence: FirearmLicenceData | SanctionsLicenceData | CancelLicencePayload
```
mikeylockwood, line 127: I think this would make more sense as `payload` rather than `licence`

mikeylockwood (approved):

web/domains/chief/types.py
MattHolmes123: I agree - its a historical name that lite-hmrc came up with. 
I'd have to change it in a bunch of tests and in view in icms-hmrc so I've never bothered.

PR #979, opened 2023-02-15
ICMSLST-1915 Import Application Declaration Templates

mikeylockwood:


web/domains/template/utils.py
```diff
@@ -0,0 +1,34 @@
+from web.domains.case._import.models import (
+    EndorsementImportApplication,
+    ImportApplication,
+)
+
+
+def add_endorsements_from_application_type(application: ImportApplication) -> None:
+    """Adds active endorsements to application based on application type"""
+
+    application_type = application.application_type
```
MattHolmes123, line 10: You only use this variable once now you've removed the endorsements_flag.

Just use it directly on line 16?

MattHolmes123 (approved):

PR #987, opened 2023-02-24
ICMSLST-1914 Cover letter templates

mikeylockwood:


web/tests/domains/template/test_utils.py
```diff
@@ -105,3 +117,83 @@ def test_add_template_data_on_submit(test_import_user, importer, office):
...
+            {"FOO": "Test"},
+            "Test some example content [[BAR]]",
+        ),
+        (
+            "[[FOO]] some example content [[BAR]]",
+            {"FOO": "Test", "BAR": "please", "EXTRA": "extra"},
+            "Test some example content please",
+        ),
+    ],
+)
```
MattHolmes123, line 149: Based on my earlier comments I thing we should change these tests.

1. Don't support optionals 
2. Don't support unknown placeholders
3. We could get fancy with our context, DEATH to inline dictionaries everywhere!

web/tests/domains/template/test_utils.py
```diff
@@ -105,3 +117,83 @@ def test_add_template_data_on_submit(test_import_user, importer, office):
...
+    context = get_cover_letter_context(app)
+
+    assert context["APPLICATION_SUBMITTED_DATE"] == "25 December 2022"
+    assert context["CONTACT_NAME"] == test_import_user
+    assert context["COUNTRY_OF_CONSIGNMENT"] == "Germany"
+    assert context["COUNTRY_OF_ORIGIN"] == "Albania"
+    assert "LICENCE_NUMBER" not in context
+    assert "LICENCE_END_DATE" not in context
+
+    context = get_cover_letter_context(app, final=True)
```
MattHolmes123, line 196: if this indeed testing a final cover letter template then we should be testing the correct licence details are shown not checking for None.
Create a draft licence for the app you've created.

MattHolmes123 (changes requested):

web/domains/template/utils.py
```diff
@@ -24,9 +35,86 @@ def add_endorsements_from_application_type(application: ImportApplication) -> No
...
+
+        if licence:
+            licence_doc = document_pack.doc_ref_licence_get(licence)
+            licence_end_date = licence.licence_end_date and licence.licence_end_date.strftime(
+                date_str
+            )
+
+        return context | {
+            "LICENCE_NUMBER": licence and licence_doc.reference,
+            "LICENCE_END_DATE": licence and licence_end_date,
```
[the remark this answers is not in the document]
mikeylockwood: Will also be using the function to validate placeholders in ICMSLST-1921. At this point there may not be a licence yet but the licence placeholder should still be valid in the editor

web/domains/template/utils.py
```diff
@@ -24,9 +35,86 @@ def add_endorsements_from_application_type(application: ImportApplication) -> No
...
+    """Returns the template content with the placeholders replaced with their value
+
+    Calling this function with replacements={'foo': 'bar'} will return the template content
+    with all occurences of [[foo]] replaced with bar"""
+
+    if content is None:
+        return ""
+
+    if context is None:
+        return content
```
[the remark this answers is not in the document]
mikeylockwood: 76 can never be `None` because if `content is None` it returns an empty string on 73. Will remove 75 and 76 though if not needed

web/domains/template/utils.py
```diff
@@ -24,9 +35,86 @@ def add_endorsements_from_application_type(application: ImportApplication) -> No
...
+
+    for replacement, value in context.items():
+        content = re.sub(TEMPLATE_CONTENT_REGEX.format(replacement), str(value), content)
```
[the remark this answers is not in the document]
mikeylockwood: I know the replacement implementation is not finished yet.  I was only doing enough to be able to implement the cover letters in this ticket. 
I think I want to be able to also retrieve all keys for validation but will think about it more in ICMSLST-1921

web/domains/template/utils.py
```diff
@@ -24,9 +35,86 @@ def add_endorsements_from_application_type(application: ImportApplication) -> No
...
+def add_application_default_cover_letter(application: ImportApplication) -> None:
+    """Adds default cover letters to an import application"""
+
+    match application.process_type:
+        case ProcessTypes.FA_DFL:
+            template = Template.objects.get(template_code="COVER_FIREARMS_DEACTIVATED_FIREARMS")
+            add_application_cover_letter(application, template)
+        case ProcessTypes.FA_OIL:
+            template = Template.objects.get(template_code="COVER_FIREARMS_OIL")
+            add_application_cover_letter(application, template)
```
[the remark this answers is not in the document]
mikeylockwood: `cover_letter_flag` will be `True` for `SILApplication` but needs to be manually set rather than automatically. Could add a pass and a comment there I guess and then a value error

web/tests/domains/template/test_utils.py
mikeylockwood: Extending the test as part of my in progress ticket where the cover letter data will be used to generate the cover letter as the licence will have been created at that point

ICMSLST-1917

web/domains/template/utils.py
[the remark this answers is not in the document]
mikeylockwood: Nice idea fora context class

web/domains/template/utils.py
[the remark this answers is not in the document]
MattHolmes123: I read this as 
```
if context is None:
    return context
```

web/domains/template/utils.py
```diff
@@ -1,7 +1,21 @@
...
 )
+from web.flow.models import ProcessTypes
+from web.utils.pdf.types import DocumentTypes
+
+from .context import CoverLetterTemplateContext
+from .models import Template
+
+# e.g. [[CONTACT_NAME]]
+TEMPLATE_CONTENT_REGEX = r"\[\[{}\]\]"
+CONTEXT_VARIABLE_REGEX = r"\[\[([A-Z]+(_[A-Z]+)*)\]\]"
```
MattHolmes123, line 15: This regex has multiple matching groups which is a bit strange.
e.g. `APP_REF_VALUE` would return a match with two values `('APP_REF_VALUE', '_VALUE')`

It also doesn't support lower case letters.
This regex works: `\[\[([A-Za-z_]+)]]` or better yet use regex flags to make it more clear what is happening and we can also ignore case:
```
CONTEXT_VARIABLE_REGEX = re.compile(
    r"""
        \[\[        # Opening braces
        ([a-z_]+)   # Match group of a-z and underscore
        ]]          # Closing braces
    """,
    flags=re.IGNORECASE | re.VERBOSE,
)
```

web/domains/template/utils.py
```diff
@@ -24,9 +38,68 @@ def add_endorsements_from_application_type(application: ImportApplication) -> No
     )
 
 
+def get_context_dict(content: str, context: ContextT) -> dict[str, str]:
+    return {hit[0]: context[hit[0]] for hit in set(re.findall(CONTEXT_VARIABLE_REGEX, content))}
```
MattHolmes123, line 42: Using my regex means you won't need to index in to each hit (as each hit matches one value)

web/domains/template/utils.py
```diff
@@ -1,7 +1,21 @@
...
     ImportApplication,
 )
+from web.flow.models import ProcessTypes
+from web.utils.pdf.types import DocumentTypes
+
+from .context import CoverLetterTemplateContext
+from .models import Template
+
+# e.g. [[CONTACT_NAME]]
+TEMPLATE_CONTENT_REGEX = r"\[\[{}\]\]"
```
MattHolmes123, line 14: The `\` isn't need for the closing brakets if the opening ones have been escaped.

This will work: `r"\[\[{}]]"`

web/domains/template/utils.py
```diff
@@ -1,7 +1,21 @@
...
+CONTEXT_VARIABLE_REGEX = r"\[\[([A-Z]+(_[A-Z]+)*)\]\]"
+
+# Types
+ContextT = CoverLetterTemplateContext
```
MattHolmes123, line 18: ContextT should be replaced with a protocol if this is going to be extended later.

```
from typing import Protocol

class TemplateContextProcessor(Protocol):
    def __init__(self, application: ImportApplication, *args, **kwargs):
        ...

        ...
```

MattHolmes123: I did args / kwargs instead of document type as that might not be required for other TemplateContextProcessor classes.

MattHolmes123: But either way right now you're saying every ContextT class will be a cover letter template processor

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #990, opened 2023-03-03
Add permission test harness and refactor permissions in to one file.

MattHolmes123:
Changes:
  - Refactor permissions.py (page, system & object level perms)
  - Add basic permission test harness to later expand with permission code.
  - Add support to open up menu item in new tab.


web/permissions.py
```diff
@@ -1,8 +1,85 @@
...
+    def _remove_prefix(v: tuple[str, str]) -> tuple[str, str]:
+        value, label = v
+
+        return value.removeprefix("web."), label
+
+
+class _PagePermissions(PermissionTextChoice):
+    view_permission_harness = (
+        "can_view_permission_harness",
+        "Can the user view the permission test harness",
```
mikeylockwood, line 41: This should be a statement but it looks like a question
`"Can view the permission test harness"` ?

MattHolmes123: Good point :)

mikeylockwood (approved):

PR #992, opened 2023-03-07
ICMSLST-1934 Add user_obj_perms to global template context.

MattHolmes123:
Changes:
  - Add context object to query object permissions in templates.
  - Add several comments relating to django-guardian config.
  - Register models with object permissions with GuardedModelAdmin.
  - Reorder User class mixins (mixins to the left).
  - Update permissions test harness to test user_obj_perms context.


web/perm_harness/views.py
```diff
@@ -9,3 +12,13 @@ class PermissionTestHarness(PermissionRequiredMixin, LoginRequiredMixin, Templat
...
     http_method_names = ["get"]
     template_name = "web/perm_harness/index.html"
+
+    def get_context_data(self, **kwargs: dict[str, Any]) -> dict[str, Any]:
+        context = super().get_context_data(**kwargs)
+
+        return context | {
+            "dummy_importer": Importer.objects.get(name="Dummy importer"),
+            "dummy_exporter": Exporter.objects.get(name="Dummy exporter"),
+            "importers": Importer.objects.all(),
```
mikeylockwood, line 22: If this is going to be used for anything other than test purposes we may want to filter on active users only

web/permissions_context.py
```diff
@@ -0,0 +1,24 @@
...
+def request_user_object_permissions(request: HttpRequest) -> UserObjectPermissionsContext:
+    """Return context variables required by apps that use the django-guardian permission system.
+
+    If there is no 'user' attribute in the request, use get_anonymous_user (from guardian.utils).
+    """
+
+    if hasattr(request, "user"):
+        user = request.user
+    else:
+        from guardian.utils import get_anonymous_user
```
mikeylockwood, line 20: Is there some sort of import clash?

web/templates/web/perm_harness/index.html
```diff
@@ -2,17 +2,43 @@
...
+  <title>Permission Test Harness</title>
 </head>
 <body>
 
-<h1>Permission Test harness</h1>
-<p>This is a test to see if perms is usable in the templates:</p>
-<p>{{ perms }}</p>
+{% macro importer_object_permissions(imp) %}
+  <p style="font-weight: bold">Importer object permissions - Importer: {{ imp }}</p>
+  {% for p in ["web.is_contact_of_importer", "web.is_agent_of_importer"] %}
```
mikeylockwood, line 11: I know this comes from the last pr, but `p` variable looks strange in html templates. Especially when using `<p>` elements

mikeylockwood (approved):

web/perm_harness/views.py
MattHolmes123: No this is code in the "PermissionTestHarness" view.

iI will change it to only include active importers though

web/permissions_context.py
MattHolmes123: No I have followed how django does it's auth context processor:
https://github.com/django/django/blob/main/django/contrib/auth/context_processors.py

As this context is added to every request I suspect it's only ever imported when it's needed for speed.
I don't know though.

web/templates/web/perm_harness/index.html
MattHolmes123: I can change it

PR #994, opened 2023-03-08
ICMSLST-1934 Add guardian direct foreign keys for Importer and Exporter.

MattHolmes123:
Changes:
  - Add direct foreign keys for models requiring object level permissions.
  - Add view to populate permission test harness with data.
  - Test prefetch_perms to optimize queries in template.

Permission Test harness performance with the following data:
  - 100 Importer records
  - 100 Exporter records
  - each test user had permissions assigned at random to each object.

Benchmark:
  - Total CPU time	17593.042 msec
  - Total DB time 1086.50 ms
  - 2431 DB queries including 2424 similar and 808 duplicates

After using prefetch_perms:
  - Total CPU time	412.736 msec
  - Total DB time 26.69 ms
  - 43 DB queries including 36 similar and 20 duplicates


web/templates/web/perm_harness/index.html
```diff
@@ -38,6 +44,8 @@ <h2>System permissions (Django) for request user {{ request.user }}</h2>
 
 <hr>
 <h2>Object permissions (Django Guardian) for request user {{ request.user }}</h2>
+{% set _ = user_obj_perms.prefetch_perms(importers) %}
```
MattHolmes123, line 47: prefetch_perms returns a boolean.
If I use `{{ user_obj_perms.prefetch_perms(importers) }}` you get `True` in the template as you'd expect.

MattHolmes123: if you don't use `set` jinja thinks `user_obj_perms` is some kind of tag and throws an error.

mikeylockwood (approved):

PR #995, opened 2023-03-09
ICMSLST-1917 Cover Letter PDF

mikeylockwood:


web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="icon" href="{{ static('web/img/favicon.png') }}" type="image/png">
```
MattHolmes123, line 4: Do we need this?

web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
...
+    <link rel="stylesheet" href="{{ static('web/css/pdfs/cover-letter.css') }}"/>
+    {% if preview %}
+      <link rel="stylesheet" href="{{ static('web/css/pdfs/preview.css') }}"/>
+    {% endif %}
+    {%- if page_title %}
+     {% set title=page_title %}
+    {% else %}
+      {% set title='Import Case Management System' %}
+    {% endif -%}
```
MattHolmes123, line 13: This if else block feels like you've copied it from somewhere else.
Do we need a title for a cover letter template?

web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
+<html>
```
MattHolmes123, line 2: Missing required 'lang' attribute 
`<html lang="en">`

web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
...
+        alt="DBT Logo"
+        class="header-block"
+        src="{{ static('web/img/dbt_red_transparent_bg.png') }}" />
+
+        <div id="address-text" class="header-block">
+          <p class="address-line">Import Licencing Branch</p>
+          <p class="address-line">Queensway House</p>
+          <p class="address-line">West Precinct</p>
+          <p class="address-line">Billingham</p>
+          <p class="address-line">TS23 2NF</p>  <!-- /PS-IGNORE -->
```
MattHolmes123, line 30: I wonder if this should be configurable somehow?
```
{% for line in ilb_address %}
  <p class="address-line">{{line}}</p>
{% endfor %}
```
Could always sort it later I suppose.

web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
...
+        </div>
+        <div class="header-block" id="dbt-url">
+          <span class="url">gov.uk/dbt</span>
```
MattHolmes123, line 33: I get page not found going to this url: `https://www.gov.uk/dbt`

web/templates/pdf/import/cover-letter.html
```diff
@@ -0,0 +1,83 @@
...
+      <div class="endorsements-block page-margin">
+        {% for endorsement in process.endorsements.all() %}
+          <p>{{ endorsement.content }}</p>
+        {% endfor %}
+      </div>
+    {% elif preview %}
+      <div class="endorsements-block page-margin">
+        <p class="additional-remarks">[[ ADDITIONAL_REMARKS ]]</p>
+      </div>
+    {% endif %}
```
MattHolmes123, line 59: If this were me I wouldn't have this `if else endif` logic in the template.
I'd supply a list of endorsements in the context.
e.g. something like this
```
if preview:
    endorsements = ["[[ ADDITIONAL_REMARKS ]]"]
else:
    endorsements = process.endorsements.values_list("content", flat=True")
```
The less logic in the template the better - templates should be as dumb as they can be

web/tests/domains/template/test_utils.py
```diff
@@ -3,26 +3,49 @@
 
 import pytest
 
+from web.domains.case._import.fa_dfl.models import DFLApplication
```
MattHolmes123, line 6: we can import all models from `web.models` .

MattHolmes123: Just seen further down you've deleted the existing imports from web.models I don't understand why.
Importing from web.models allows for easier refactoring as if you change the underlying structure web.models isn't going anywhere

MattHolmes123 (changes requested):

web/domains/template/context.py
```diff
@@ -1,12 +1,24 @@
...
+
+    def __getitem__(self, item: str) -> str:
+        ...
+
 
 class CoverLetterTemplateContext:
     date_fmt = "%d %B %Y"
 
-    def __init__(self, application: ImportApplication, document_type: DocumentTypes):
+    def __init__(self, application: "ImportApplication", document_type):
```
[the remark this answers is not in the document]
mikeylockwood: Trying to figure out why DocumentTypes was causing a circular import and forgot to put it back in

web/tests/domains/template/test_utils.py
mikeylockwood: It was again part of trying to figure out the circular import

web/templates/pdf/import/cover-letter.html
mikeylockwood: Yes as it doesn't inherit from the base template. The favicon image seems to just be a placeholder at the minute so will need to be replaced at some point

web/templates/pdf/import/cover-letter.html
mikeylockwood: I want different classes for the p elements still. It is nicer just to do this in the template

web/templates/pdf/import/cover-letter.html
mikeylockwood: I'll just update wit when it's created if that makes you happy. The dit link says this page has a big banner saying this organisation is changing so will need to be updated

web/templates/pdf/import/cover-letter.html
MattHolmes123: Oh yer I just noticed the extra class on the p tag.
You could still do that without the neeed for the if else and django db logic in the template.

MattHolmes123 (approved):

MattHolmes123 (commented):
It is generating cover letters for sanctions applications

PR #1015, opened 2023-03-23
ICMSLST-1939 Update permissions for Importer/Exporter admin views.

MattHolmes123:
ILB Admins and Org contacts can now manage Importers & Exporter records.

Changes:
  - Add permission checks to all Importer views
  - Add permission checks to all Exporter views.
  - Add Exporter Details view.
  - Restrict firearm authority sections with new permission check.
  - Refactor agent list to only allow ILB Admins to delete agents.
  - Update the editable user permissions on the Edit Importer/Exporter views.
  - Remove permissions not required.


web/permissions/context_processors.py
```diff
@@ -3,9 +3,24 @@
 from django.http import HttpRequest
 from guardian.core import ObjectPermissionChecker
 
+from . import can_user_edit_org, can_user_manage_org_contacts, can_user_view_org
+
+
+class UserObjectPerms(ObjectPermissionChecker):
+    # TODO: These methods don't use the ObjectPermissionChecker cache (extra db call every time)
+    # each call to user.has_perm() creates a new ObjectPermissionChecker instance
```
MattHolmes123, line 11: This is a performance problem if the list of importers / exporters associated with a user is large.

The importer / exporter detail pages only ever show a max of 10 records at a time so I don't see it being a problem we need to revisit any time soon.

PR #1030, opened 2023-04-12
ICMSLST-1169 Update assignee details search results section.

MattHolmes123:


web/utils/search/api.py
```diff
@@ -332,18 +314,28 @@ def _get_export_result_row(
...
         agent_name=rec.agent.name if rec.agent else None,
         actions=get_export_record_actions(rec, user_org_permissions),
         order_by_datetime=rec.order_by_datetime,  # This is an annotation
     )
 
 
+def _get_assignee_details(app: ImportApplication | ExportApplication) -> types.AssigneeDetails:
+    assignee_name = f"{app.case_owner.full_name} ({app.case_owner.email})" if app.case_owner else ""
+    reassignment_at = (
+        app.reassign_datetime.strftime("%d %b %Y %H:%M") if app.reassign_datetime else ""
```
mikeylockwood, line 327: This datetime string format is used a few times, might be worth making it a constant

MattHolmes123: I've thought this a few times but it isn't worth it.
We actually need to localise the app correctly before Django 5.0

https://docs.djangoproject.com/en/4.2/ref/settings/#use-l10n

Currently ICMS v2 has this setting and breaks without it:
`USE_L10N = False`

PR #1033, opened 2023-04-19
ICMSLST-1986 Update FA-DFL views to use new permissions

MattHolmes123:


web/permissions/service.py
```diff
@@ -24,6 +26,91 @@
...
+    org: ORGANISATION = dataclasses.field(init=False)
+    agent_org: ORGANISATION = dataclasses.field(init=False)
+    obj_perms: IMP_OR_EXP_PERMS_T = dataclasses.field(init=False)
+
+    def __post_init__(self):
+        self.is_ilb_admin = self.user.has_perm(Perms.sys.ilb_admin)
+
+        if self.app.is_import_application():
+            self.has_org_access = self.user.has_perm(Perms.sys.importer_access)
+            self.org = self.app.importer if self.app.is_import_application() else self.app.exporter
```
mikeylockwood, line 50: No need to check if import application here

mikeylockwood (approved):

web/permissions/service.py
MattHolmes123: Ooops! That was from a former version :) 
Thanks for spotting it.

PR #1035, opened 2023-04-20
ICMSLST-1924 Revoke licence email

mikeylockwood:


web/notify/email.py
```diff
@@ -29,37 +33,65 @@ def send_email(
...
+                subject,
+                message,
+                utils.get_notification_emails(importer.user),
+                html_message=html_message,
+            )
+
+        return
+
+    # Importer organisation
+    for user in utils.get_importer_contacts(
```
MattHolmes123, line 52: There is already a function in `web/permissions` to get contacts for an organisation:
`organisation_get_contacts`
 
 You pass it an importer or exporter instance and it returns you all active contacts at that org.

web/notify/utils.py
```diff
@@ -40,3 +47,27 @@ def get_case_officers_emails():
...
+def get_exporter_contacts(exporter: "Exporter", permission: str | None = None) -> "QuerySet[User]":
+    """Return a queryset of users for an exporter with specified permission"""
+    if not permission:
+        permission = ExporterObjectPermissions.edit.codename
+
+    return User.objects.prefetch_related("exporteruserobjectpermission_set").filter(
+        exporteruserobjectpermission__content_object=exporter,
+        exporteruserobjectpermission__permission__codename=permission,
+        account_status=User.ACTIVE,
+    )
```
MattHolmes123, line 73: Can all this be replaced by:
`icms/web/permissions/service.py -> organisation_get_contacts`

web/notify/email.py
```diff
@@ -1,14 +1,18 @@
...
 
 from config.celery import app
 from web.models import Exporter, Importer, User
+from web.permissions.perms import ExporterObjectPermissions, ImporterObjectPermissions
 
 from . import utils
 
+if TYPE_CHECKING:
+    from web.domains.case.types import ImpOrExp
+
```
MattHolmes123, line 15: I've changed my opinion on when to use `TYPE_CHECKING` btw.
Unless it is expensive to import the object or it causes a circular dependency I just import the object directly.
It removes the need to add double quotes to your type annotations as well.

web/domains/template/context.py
```diff
@@ -75,3 +76,49 @@ def __getitem__(self, item: str) -> str:
...
+        if isinstance(self.process, ImportApplication):
+            return self._import_context(item)
+        elif isinstance(self.process, ExportApplication):
+            return self._export_context(item)
+        elif isinstance(self.process, AccessRequest):
+            return self._access_context(item)
+        else:
+            raise ValueError(
+                "Process must be an instance of ImportApplication / ExportApplication / AccessRequest"
+            )
```
MattHolmes123, line 95: This shoulf work if you want to use `match`
```
        match self.process:
            case ImportApplication():
                return self._import_context(item)
            case ExportApplication():
                return self._export_context(item)
            case AccessRequest():
                return self._access_context(item)
            case _:
                raise ValueError(
                    "Process must be an instance of ImportApplication / ExportApplication / AccessRequest"
                )
```

Or is this what you were asking the other day about matching subclasses?
In which case I don't know you'd have to test it :) 

**edit:**
It does work with subclasses so you can use the match version if you like.

MattHolmes123 (changes requested):

web/domains/case/views/views_search.py
```diff
@@ -453,38 +454,30 @@ def form_valid(self, form: RevokeApplicationForm) -> HttpResponseRedirect:
...
+            if is_import:
+                email_subject, email_body = get_email_template_subject_body(
+                    self.application, "LICENCE_REVOKE", RevokedEmailTemplateContext
+                )
+            else:
+                email_subject, email_body = get_email_template_subject_body(
+                    self.application, "CERTIFICATE_REVOKE", RevokedEmailTemplateContext
+                )
+
+            send_to_application_contacts(self.application, email_subject, email_body)
```
MattHolmes123, line 476: Please move this block to the end of the method

web/domains/template/utils.py
```diff
@@ -92,6 +92,22 @@ def get_cover_letter_content(
     return replace_template_values(application.cover_letter_text, context)
 
 
+def get_email_template_subject_body(
+    process: "Process",
+    template_code: str,
+    Context: type[EmailTemplateContext] = EmailTemplateContext,
```
MattHolmes123, line 98: Lets follow pep8 please: https://peps.python.org/pep-0008/#function-and-variable-names
`context_cls` maybe?

web/domains/template/context.py
```diff
@@ -1,14 +1,15 @@
-from typing import TYPE_CHECKING, Protocol
+from typing import Protocol
 
+from web.domains.case._import.models import ImportApplication
+from web.domains.case.access.models import AccessRequest
+from web.domains.case.export.models import ExportApplication
```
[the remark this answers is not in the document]
MattHolmes123: Please do this unless we have a good reason not to.
It was a large refactor to fix all the model imports

web/notify/email.py
```diff
@@ -29,37 +30,71 @@ def send_email(
...
+        send_to_importer_contacts(importer, subject, message, html_message)
+
+
+def send_to_exporter_contacts(
+    exporter: Exporter, subject: str, message: str, html_message: str | None = None
+) -> None:
+    obj_perms = get_org_obj_permissions(exporter)
+    contacts = organisation_get_contacts(exporter, perms=[obj_perms.is_contact.codename])
+
+    # Importer organisation
```
MattHolmes123, line 70: `# Exporter organisation`

MattHolmes123 (approved):

PR #1064, opened 2023-05-10
ICMSLST-1970 Sanctions Email

mikeylockwood:
Sanction emails aren't currently used in the system (and probably never will be) but implementing for completeness

### DIFF
--- a/web/domains/template/context.py
+++ b/web/domains/template/context.py
@@ -96,6 +96,16 @@ def _get_import_goods_description(app: ImportApplication) -> str:
                 ]
             )
 
+        case ProcessTypes.SANCTIONS:
+            return "\n".join(
+                [
+                    f"{str(quantity).rstrip('0').rstrip('.')} x {desc}"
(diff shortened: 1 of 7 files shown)


web/domains/template/context.py
```diff
@@ -96,6 +96,16 @@ def _get_import_goods_description(app: ImportApplication) -> str:
                 ]
             )
 
+        case ProcessTypes.SANCTIONS:
+            return "\n".join(
+                [
+                    f"{str(quantity).rstrip('0').rstrip('.')} x {desc}"
```
mikeylockwood, line 102: A bit of an ugly way to strip trailing 0s in a decimal number, but saves inaccuracies of parsing to a float. `quantity_amount` will always have a decimal places `quantity_amount = models.DecimalField(max_digits=14, decimal_places=3, verbose_name="Quantity")`

MattHolmes123 (approved):

PR #1073, opened 2023-05-16
ICMSLST-1980 - Case Refused email

marcuspp:


web/notify/utils.py
```diff
@@ -65,3 +67,12 @@ def create_case_email(
         case_email.attachments.add(*attachments)
 
     return case_email
+
+
+def _render_email(template, context):
```
MattHolmes123, line 72: Remove the leading understore if we are moving this to a utils file?

MattHolmes123 (approved):

PR #1077, opened 2023-05-17
ICMSLST-1961 Reopen case email

marcuspp:


web/domains/case/views/views_search.py
```diff
@@ -195,20 +196,21 @@ class ReopenApplicationView(
...
-        self.application.update_order_datetime()
-        self.update_application_status()
-        self.update_application_tasks()
-        document_pack.pack_draft_create(self.application)
-
-        messages.success(
-            request,
-            "The case has been reopened, and can be taken ownership of by any case manager.",
-        )
+        with transaction.atomic():
```
MattHolmes123, line 199: The post method is already atomic:
`@method_decorator(transaction.atomic, name="post")`

MattHolmes123: I don't really mind which we use but not both.

web/notify/constants.py
```diff
@@ -0,0 +1,8 @@
+import enum
+
+
+class DatabaseEmailTemplate(enum.Enum):
```
MattHolmes123, line 4: We are running ICMS using pythin 3.11.
Let's use StrEnum

```
>>> from enum import Enum, StrEnum
>>> class Foo(Enum):
...     VAL = "ASDF"
... 
>>> class Bar(StrEnum):
...     VAL = "ASDF"
... 
>>> print(Foo.VAL == "ASDF")
False
>>> print(Foo.VAL.value == "ASDF")
True
>>> print(Bar.VAL == "ASDF")
True
>>> print(Bar.VAL.value == "ASDF")
True
>>> 
```

MattHolmes123 (changes requested):
added a few comments.

web/domains/case/views/views_search.py
marcuspp: 👍 I didn't notice the decorator so I have just removed the with statement

web/notify/constants.py
marcuspp: Updated! I need to the read the whats new in 3.11 :-)

PR #1079, opened 2023-05-22
ICMLST-1965 Withdraw Application emails

marcuspp:


web/domains/case/models.py
```diff
@@ -163,11 +163,13 @@ class WithdrawApplication(models.Model):
...
     STATUS_REJECTED = "rejected"
     STATUS_ACCEPTED = "accepted"
+    STATUS_DELETED = "deleted"
 
     OPEN = (STATUS_OPEN, "OPEN")
     REJECTED = (STATUS_REJECTED, "REJECTED")
     ACCEPTED = (STATUS_ACCEPTED, "ACCEPTED")
-    STATUSES = (OPEN, REJECTED, ACCEPTED)
+    DELETED = (STATUS_DELETED, "DELETED")
+    STATUSES = (OPEN, REJECTED, ACCEPTED, DELETED)
```
mikeylockwood, line 172: I think it might be worth switching this for a `TypedTextChoices` class

```
class Statuses(TypedTextChoices):
    OPEN = ("open", "OPEN")
   ...

status = models.CharField(max_length=10, choices=Statuses.choices, default=Statuses.OPEN)
```
Would jus need to make sure `WithdrawApplication.STATUS_OPEN`, `WithdrawApplication.STATUS_REJECTED`, etc. are replaced in the code with `WithdrawApplication.Statuses.OPEN`

An example of this is in `AccessRequest`
`web/domains/case/access/models.py`

web/notify/email.py
```diff
@@ -178,6 +185,36 @@ def send_html_email(
...
+        status = " Cancelled"
+    elif withdrawal.status != WithdrawApplication.STATUS_OPEN:
+        status = " " + withdrawal.get_status_display().title()
+    return f"Withdrawal Request{status}: {application.reference}"
+
+
+def send_withdrawal_email(withdrawal: WithdrawApplication) -> None:
+    contacts = None
+
+    if withdrawal.status not in dict(WithdrawApplication.STATUSES).keys():
```
mikeylockwood, line 200: If the `Statuses` `TypedTextChoices` class is added to the model, this could change to `if withdrawal.status not in WithdrawApplication.Statuses`

mikeylockwood (approved):

PR #1088, opened 2023-05-30
ICMLST-1969 Variation Request emails

marcuspp:
- Refactored VariationRequest statuses to use `TypedTextChoices`
 - Refactored tests to use a shared `add_variation_request_to_app` function
 - Added emails for VariationRequest
 - Fixed cta on VariationRequest form
 - Removed TextArea widget for `variation_refuse_reason` as `text-editor.js` does not render the Text Areas correctly when there is 2 on a page.
- Tested for export applications. Only email sent is cancelled


web/notify/email.py
```diff
@@ -219,6 +226,45 @@ def send_withdrawal_email(withdrawal: WithdrawApplication) -> None:
...
+    contacts = None
+
+    if description not in list(VariationRequestDescription):
+        return
+
+    if description in [VariationRequestDescription.CANCELLED]:
+        contacts = [variation_request.requested_by]
+    elif description == VariationRequestDescription.UPDATE_RECEIVED:
+        contacts = get_ilb_admin_users()
+
```
marcuspp, line 258: Not 100% sure if the recipients are correct.

web/domains/case/views/views_misc.py
```diff
@@ -457,15 +460,16 @@ def start_authorisation(
...
                     next_task = None
                     application.status = model_class.Statuses.COMPLETED
-                    vr.status = VariationRequest.REJECTED
+                    vr.status = VariationRequest.Statuses.REJECTED
                     vr.reject_cancellation_reason = application.variation_refuse_reason
                     vr.closed_datetime = timezone.now()
-
                     vr.save()
-
+                    send_variation_request_email(
```
MattHolmes123, line 470: This should probably be moved to the bottom of the function.
Maybe after `send_refused_email`.
I just think triggering tasks to execute after everything else has successfully completed is a good idea.

web/notify/email.py
```diff
@@ -219,6 +226,45 @@ def send_withdrawal_email(withdrawal: WithdrawApplication) -> None:
...
+    if description == VariationRequestDescription.CANCELLED:
+        return "Variation Request Cancelled"
+    if description == VariationRequestDescription.UPDATE_REQUIRED:
+        return "Variation Update Required"
+    if description == VariationRequestDescription.UPDATE_CANCELLED:
+        return "Variation Update No Longer Required"
+    if description == VariationRequestDescription.UPDATE_RECEIVED:
+        return "Variation Update Received"
+    if description == VariationRequestDescription.REFUSED:
+        return f"Variation on application reference {application.reference} has been refused by ILB"
```
MattHolmes123, line 241: Not required but I think match statements get better the more if statements you have.
also raise a `ValueError` for unsupported description values

MattHolmes123 (changes requested):

web/notify/email.py
marcuspp: Updated to case owner

web/notify/email.py
```diff
@@ -216,7 +229,48 @@ def send_withdrawal_email(withdrawal: WithdrawApplication) -> None:
...
+
+
+def send_variation_request_email(
+    variation_request: VariationRequest,
+    description: VariationRequestDescription,
+    application: ImpOrExp,
+) -> None:
+    subject = get_variation_request_email_subject(description, application)
+
+    if description in [VariationRequestDescription.CANCELLED]:
```
MattHolmes123, line 260: change to ==

MattHolmes123 (approved):
Added a minor comment.

PR #1096, opened 2023-06-06
ICMSLST-1963 - Application Update emails

marcuspp:


web/notify/email.py
```diff
@@ -273,6 +273,33 @@ def send_variation_request_email(
     send_html_email(template_name, context, contacts)
 
 
+def get_application_update_request_contents(
+    application: ImpOrExp, case_type: str
+) -> tuple[str, str]:
+    match case_type:
+        case "import":
```
mikeylockwood, line 280: Could do this without the need for `case_type` as you already have the application object

```
match application:
    case ImportApplication():
     ...
```

mikeylockwood (approved):

web/tests/domains/case/views/test_views_update_request.py
```diff
@@ -1,18 +1,41 @@
-from typing import TYPE_CHECKING
-
+from django.test.client import Client
 from pytest_django.asserts import assertContains, assertTemplateUsed
 
-from web.tests.helpers import CaseURLS
+from web.domains.case.models import UpdateRequest
```
MattHolmes123, line 4: can you import this from `web.models`
All the app was changed to work this way to simplify imports so lets stick to it.

web/tests/domains/case/views/test_views_update_request.py
```diff
@@ -1,18 +1,41 @@
...
+from web.domains.case.types import ImpOrExp
+from web.models import ImportApplication
+from web.tests.application_utils import submit_app
+from web.tests.helpers import CaseURLS, check_email_was_sent
+
+
+def get_open_update_request(
+    client: Client,
+    app: ImpOrExp,
+    post: dict = {"request_subject": "subject", "request_detail": "detail"},
```
MattHolmes123, line 14: This is a mutable default argument.
better to be this
```python
def get_open_update_request(
    client: Client,
    app: ImpOrExp,
    post: dict | None = None
):
    if not post:
        post = {"request_subject": "subject", "request_detail": "detail"}
```

MattHolmes123: e.g.
image

web/tests/domains/case/views/test_views_update_request.py
```diff
@@ -23,6 +46,90 @@ def test_list_update_requests_get(
     assert resp.context["update_request"] is None
 
 
-# def test_manage_update_requests_post():
-#     # TODO: Add test for sending an update request
-#     ...
+def test_manage_update_requests(ilb_admin_client: Client, wood_app_submitted):
```
MattHolmes123, line 49: add missing type for consistency for other tests.
plus none of them have return types - But that's not really a problem with unittest functions.

MattHolmes123 (approved):
Approved but added a few comments to change things in the tests.
👍

web/tests/domains/case/views/test_views_update_request.py
MattHolmes123: I guess it could be `if post is None` if you ever want to supply an empty dict as a valid post value.

PR #1102, opened 2023-06-12
ICMSLST-1998 - Refactored sidebar menu for import applications

marcuspp:


web/domains/case/_import/models.py
```diff
@@ -98,6 +99,41 @@ def __str__(self) -> str:
...
+                return reverse("import:create-fa-dfl")
+            case self.SubTypes.SIL:
+                return reverse("import:create-fa-sil")
+            case _:
+                raise ValueError(
+                    f"Unknown Firearms Application Sub Type: {self.sub_type}"  # /PS-IGNORE
+                )
+
+    @property
+    def create_application_url(self) -> str:
```
mikeylockwood, line 116: We generally are trying to keep web logic separate from the models. I do see the value in having this as a property on the model being able to access it directly in the template. I guess an alternative could be having a a function that can be accessed in the templates in `web/jinja2.py`, but maybe it's too specific for a global function

marcuspp: Yeah happy to move it to a template tag if preferred. There was just no way for me to work out the url based on the data in the model. Something to consider in future is to maybe a slug field to model. It may simplify things.

web/domains/case/utils.py
```diff
@@ -12,7 +12,7 @@
 from web.domains.case.services import document_pack, reference
 from web.flow.models import ProcessTypes
 from web.models import ExportApplication, ImportApplication, Task, User
-from web.permissions import AppChecker
+from web.permissions.service import AppChecker
```
MattHolmes123, line 15: Can this be reverted to `from web.permissions import AppChecker`

web/domains/case/_import/models.py
mikeylockwood: I was thinking a slug field also when I was looking at it, would make things much easier

web/domains/case/_import/views.py
```diff
@@ -57,21 +57,9 @@
...
-        "show_textiles": ImportApplicationType.objects.get(
-            type=ImportApplicationType.Types.TEXTILES
-        ).is_active,
-        "show_sps": ImportApplicationType.objects.get(
-            type=ImportApplicationType.Types.SPS
-        ).is_active,
-        "show_ironsteel": ImportApplicationType.objects.get(
-            type=ImportApplicationType.Types.IRON_STEEL
-        ).is_active,
+        "application_types": ImportApplicationType.objects.filter(is_active=True).exclude(type="FA")
```
MattHolmes123, line 62: I was looking in to this `.exclude(type="FA")`.
I followed it back to the template and I realised that this doesn't cover all app types.

We really just want a function that returns all active application types to show in the create application screen.

To simplify everything we want something like this.
```python
Def get_active_application_create_link() -> list[tuple[str, str]]:
     """Returns (app label, app_create_url) pairs for all active app types"""
     ...
```

MattHolmes123: Just looked at your loop in the template using ImportApplicationTemplate.name.
We might need to be careful and decide how we want to get the labels.

The name attribute differs to the labels in the template before your change.
```python
>>> from web.models import ImportApplicationType
>>> for iat in ImportApplicationType.objects.all():
...     print(iat.name)
...
Derogation from Sanctions Import Ban
Firearms and Ammunition (Deactivated Firearms Licence)
Firearms and Ammunition (Open Individual Import Licence)
Firearms and Ammunition (Specific Individual Import Licence)
Iron and Steel (Quota)
Outward Processing Trade
Prior Surveillance
Sanctions and Adhoc Licence Application
Textiles (Quota)
Wood (Quota)
```

MattHolmes123: image

web/domains/case/_import/models.py
MattHolmes123: There are a few places in the models that return URLs for different scenarios, edit, create etc.

You can leave it here for me.

web/templates/partial/case/import/sidebar-create.html
```diff
@@ -1,16 +1,15 @@
...
-    {{ icms_link(request, icms_url('import:create-fa-oil'), 'Open Individual Import Licence') }}
-
-    {{ icms_link(request, icms_url('import:create-fa-sil'), 'Specific Individual Import Licence') }}
-
+  {% if other_application_types %}
     <h5>Other</h5>
-
-    {% for application_type in application_types %}
+    {% for application_type in other_application_types %}
         {{ icms_link(request, application_type.create_application_url, application_type.name ) }}
```
MattHolmes123, line 12: what about `get_type_display`?

mikeylockwood (approved):

PR #1108, opened 2023-06-20
Fix password reset template to match other registration templates.

MattHolmes123:


web/templates/registration/password_reset_confirm.html
```diff
@@ -3,32 +3,36 @@
...
+      <h1>Password Reset</h1>
       <p class="subheading">Import Case Management System</p>
     </div>
     <div class="clearBoth"></div>
 
     <p>Please enter your new password twice so we can verify you typed it in correctly.</p>
     {% call forms.form(method='post', csrf_input=csrf_input) %}
       {% for field in form %}
-        {{ fields.field(field) }}
+        {% if field.name == "new_password1" %}
```
mikeylockwood, line 14: `new_password1` doesn't look right

mikeylockwood: It is part of django's password change form

mikeylockwood (approved):

web/templates/registration/password_reset_confirm.html
MattHolmes123: its the name of the form field :)

PR #1109, opened 2023-06-21
ICMSLST-1933 - Fix for View Prepare Response, Licence undefined

marcuspp:
ICMSLST-661
 - Introduces the ability to simply turn on jinja2.StrictUndefined  
 - `@override_settings(TEMPLATES=settings.STRICT_TEMPLATES)` or `TEMPLATES=STRICT_TEMPLATES` within a settings file
 

ICMSLST-1933
 - Fixes issue with view by not rendering licence (and endorsement) blocks for Export Applications
 - Separated the logic for selecting the correct licence within the view


web/jinja2.py
```diff
@@ -174,6 +174,7 @@ def get_user_obj_perms(user: "User") -> ObjectPermissionChecker:
 
 
 def environment(**options):
+    options.pop("string_if_invalid", None)
```
MattHolmes123, line 177: What's this for?

marcuspp: https://github.com/pytest-dev/pytest-django/issues/327 Can technically be removed if we don't use the `--fail-on-template-vars`

web/domains/case/views/views_prepare_response.py
```diff
@@ -93,32 +107,22 @@ def prepare_response(
...
-            and application.decision == application.APPROVE
-        ):
-            context["licence"] = document_pack.pack_active_get(application)
-        elif application.status == ImpExpStatus.REVOKED:
-            context["licence"] = document_pack.pack_revoked_get(application)
-        elif (
             application.status == ImpExpStatus.VARIATION_REQUESTED
             and application.variation_decision == application.REFUSE
         ):
-            context["licence"] = document_pack.pack_active_get(application)
```
MattHolmes123, line 110: Haven't we lost this logic in the new version?
From memory a refused variation request should show the active licence (as the new draft one is being refused).
I believe that's what V1 does but I will have to check after lunch

web/jinja2.py
MattHolmes123: Maybe add a link in a comment above the line for future referenec?

web/tests/domains/case/views/test_views_prepare_response.py
```diff
@@ -14,67 +18,110 @@
...
+        _license = get_licence_for_application(completed_dfl_app)
+        assert _license.status == DocumentPackBase.Status.REVOKED
+
+    def test_get_licence_for_rejected_application(self, complete_rejected_app):
+        _license = get_licence_for_application(complete_rejected_app)
+        assert _license is None
+
+    # def test_prepare_response_post():
+    #     # TODO: Add test for approving an application
+    #     # TODO: Add test for rejecting an application
```
MattHolmes123, line 126: Do you want to create a story in the tech debt epic for these todo's and the the one above (# TODO: Add test for each application type as response prep is different for each)

Just a story to fully test the prepare responce view.

And then reference the story in the TODO comments. 
(Sorry because I know you didn't add these comments - I bet it was me)

MattHolmes123 (approved):

PR #1127, opened 2023-07-11
ICMSLST-2099 - Import files to S3 periodically save progress

marcuspp:


data_migration/management/commands/import_v1_files_to_s3.py
```diff
@@ -24,6 +24,8 @@ def get_query_last_run_key(query_name: str) -> str:
 
 
 class Command(BaseCommand):
+    SAVE_RUN_DATA = 100  # Save run data to s3 periodically after processing this number of files
```
marcuspp, line 27: This could potential be an argument. I have excluded it for now as I'm aware there are a lot of arguments and it could get confusing. I will be reviewing all arguments in my next task.

data_migration/management/commands/import_v1_files_to_s3.py
```diff
@@ -14,16 +14,18 @@
...
-REQUIRED_QUERY_PARAMETERS = ["created_datetime"]
-
 
 def get_query_last_run_key(query_name: str) -> str:
     return f"{query_name}-last-run.json"
 
 
 class Command(BaseCommand):
+    SAVE_RUN_DATA: int = (
+        100  # Save run data to s3 periodically after processing this number of files
```
MattHolmes123, line 24: I would probably put this comment on the line above so it doesn't split to three lines like this.

MattHolmes123 (approved):

mikeylockwood (approved):

PR #1128, opened 2023-07-12
ICMSLST-2098 Refactor Import to s3 script options

marcuspp:
- Removed `--s3-file-prefix` option in favour of prefixing the query name to the file path within the queries themselves
- Renamed `--number-of-rows` to `--batchsize` to be more consistent with other data migration commands
- Created an option  `--run-data-batchsize` to replace `SAVE_RUN_DATA`
- Introduced query aliases `small` / `large` that groups queries based on total data (small below 1.5 GB)
- Updated Readme

### DIFF
--- a/data_migration/README.md
+++ b/data_migration/README.md
@@ -46,6 +46,31 @@ The following data_types are valid in the `--start` parameter
 * `f` or `file` for the file data type
 * `ia` or `import_application` for the import_application data type
 
+## File Import
+
+The following script is used to pull the file data from V1 and export into S3
+
+```
+./manage.py import_v1_files_to_s3
+```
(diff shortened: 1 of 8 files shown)


data_migration/README.md
```diff
@@ -46,6 +46,31 @@ The following data_types are valid in the `--start` parameter
...
 
+## File Import
+
+The following script is used to pull the file data from V1 and export into S3
+
+```
+./manage.py import_v1_files_to_s3
+```
+The command can be run with options
+* `--batchsize` to modify the size of the batches when running the migration (default: 1000)
```
mikeylockwood, line 57: Default is set to 500 in the command

PR #1130, opened 2023-07-14
ICMSLST-2061 Fetch Email Attachments in Celery Task

mikeylockwood:
Pass attachment ids instead of attachment files to send_email so emails with attachments can be called as a celery task. Fetch the files from S3 within send_email task instead

### DIFF
--- a/web/notify/email.py
+++ b/web/notify/email.py
@@ -28,16 +28,16 @@ from . import utils
 from .constants import DatabaseEmailTemplate, VariationRequestDescription
 
 
-# TODO ICMSLST-2061 Change attachments to be fetched from S3 within this task
 @app.task(name="web.notify.email.send_email")
 def send_email(
     subject: str,
     body: str,
     recipients: Collection[str],
     cc: Collection[str] = (),
-    attachments: Collection[tuple[str, bytes]] = (),
+    attachment_ids: tuple[int, ...] = (),
     html_message: str | None = None,
 ) -> None:
+    attachments = utils.get_attachments(attachment_ids)
     message = EmailMultiAlternatives(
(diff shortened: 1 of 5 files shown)


web/notify/notify.py
```diff
@@ -174,18 +173,15 @@ def send_fir_to_contacts(
...
-            # attachments=attachments, TODO ICMSLST-2061
+            attachment_ids=attachment_ids,
         )
 
 
 def send_further_information_request(process: Process, fir: FurtherInformationRequest) -> None:
     context = {"subject": fir.request_subject, "body": fir.request_detail}
-    fir_files = fir.files.filter(is_active=True).values("filename", "path").order_by("pk")
-    # TODO ICMSLST-2061 attachments need to be specific to email being sent not all attachments
-    # Different attachments sent at different point in email chain
```
mikeylockwood, line 185: Attachments are only sent in the initial FIR. In the response, the user just uploads the attachments to the FIR when the case officer can view them. Attachments are not included in the response email notification

web/notify/email.py
```diff
@@ -28,16 +28,16 @@
...
 
-# TODO ICMSLST-2061 Change attachments to be fetched from S3 within this task
 @app.task(name="web.notify.email.send_email")
 def send_email(
     subject: str,
     body: str,
     recipients: Collection[str],
     cc: Collection[str] = (),
-    attachments: Collection[tuple[str, bytes]] = (),
+    attachment_ids: Collection[int] = (),
```
MattHolmes123, line 37: You have a few collection types but always covert the incoming data to a tuple.
You can type an any length tuple like this if you wanted:
`attachment_ids: tuple[int, ...]`

MattHolmes123 (approved):

PR #1131, opened 2023-07-14
ICMSLST-2107 - Fix sanction admins only get relevant emails

marcuspp:


web/notify/notify.py
```diff
@@ -92,18 +83,12 @@ def register(user, password):
...
-    subject = f"Access Request {case_reference}"
-    send_case_officer_notification(
-        subject, "email/access/access_requested.html", context={"subject": subject}
+def send_access_requested_email(access_request):
+    context = {"subject": f"Access Request {access_request.reference}"}
+    email.send_html_email(
+        "email/access/access_requested.html",
+        context,
+        list(get_case_officers_for_process_type(access_request.process_type)),
     )
```
MattHolmes123, line 92: Right now all access requests are managed by ILB, the sanctions admins don't get involved until there has been an application submitted.

MattHolmes123: So the access request code should probably still use: `get_ilb_admin_users()`

web/permissions/service.py
```diff
@@ -372,6 +373,18 @@ def get_ilb_admin_users() -> QuerySet[User]:
...
+def get_case_officers_for_process_type(process_type: ProcessTypes) -> QuerySet[User]:
+    """Return the case officers for a given process type."""
+    sanction_case_officers = get_users_with_permission(Perms.sys.sanctions_case_officer)
+    match process_type:
+        case ProcessTypes.SANCTIONS:
+            return sanction_case_officers
+        case _:
+            return get_users_with_permission(Perms.sys.ilb_admin).exclude(
+                pk__in=sanction_case_officers.values("pk")
+            )
```
MattHolmes123, line 385: We need to check this with ILB as I don't think the logic is correct.
I think it should be this

- For Sanctions applications: ILB Admins & Sanctions Case Officers
- For everthing else: ILB Admins

MattHolmes123: @mikeylockwood what do you think?

mikeylockwood: Yeah I think `get_users_with_permission(Perms.sys.ilb_admin)` for sanctions

web/notify/email.py
```diff
@@ -116,20 +117,6 @@ def get_application_contacts(application: ImpOrExp) -> QuerySet[User]:
...
-) -> None:
-    if application.case_owner:
-        send_email.delay(
-            subject,
-            message,
-            utils.get_notification_emails(application.case_owner),
-            html_message=html_message,
-        )
-    else:
-        send_to_contacts(subject, message, get_ilb_admin_users(), html_message)
```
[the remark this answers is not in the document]
mikeylockwood: Withdraw Request, Variation Request Response, Further Information Request Response and Application Update Response emails should all go to the assigned case officer if there is one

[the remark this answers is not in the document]
MattHolmes123: This code has never been used though so is that something one of you is planning on doing?

web/permissions/service.py
MattHolmes123: Yep cool so we all agree.

web/notify/email.py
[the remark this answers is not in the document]
MattHolmes123: I have just seen other code doing things like this:
`contacts = [application.case_owner]` so I assume that's what you're referring to

web/notify/notify.py
marcuspp: The process type is `ExporterAccessRequest.PROCESS_TYPE` or `ImporterAccessRequest.PROCESS_TYPE` so the email will be sent to ilb_admins - sanction officers

web/notify/email.py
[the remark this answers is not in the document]
marcuspp: Yes removed function as was not used anywhere. Will raise another ticket to look at case owners and make the logic consistent everywhere

web/notify/email.py
```diff
@@ -215,7 +202,7 @@ def send_withdrawal_email(withdrawal: WithdrawApplication) -> None:
         WithdrawApplication.Statuses.OPEN,
         WithdrawApplication.Statuses.DELETED,
     ]:
-        contacts = get_ilb_admin_users()
+        contacts = get_case_officers_for_process_type(ProcessTypes.WITHDRAWAL)
```
MattHolmes123, line 205: use `application.process_type` and then we can get rid of `ProcessTypes.WITHDRAWAL`

web/permissions/service.py
```diff
@@ -372,6 +373,19 @@ def get_ilb_admin_users() -> QuerySet[User]:
     return get_users_with_permission(Perms.sys.ilb_admin)
 
 
+def get_case_officers_for_process_type(process_type: ProcessTypes) -> QuerySet[User]:
+    """Return the case officers for a given process type.
+
+    Sanctions case officers are a subset of ilb admin users
+    """
+    sanction_case_officers = get_users_with_permission(Perms.sys.sanctions_case_officer)
```
MattHolmes123, line 381: Move this to `case _:` block as it's the only place it's used?

web/tests/domains/case/access/test_views.py
```diff
@@ -9,7 +9,22 @@
...
+    exp_sent_to_list: list[str],
+    exp_subject: str,
+    exp_in_body: str | None = None,
+):
+    check_email_was_sent(exp_num_emails, exp_sent_to_list[0], exp_subject, exp_in_body)
+    outbox = mail.outbox
+    sent_to = []
+    for _email in outbox:
+        if _email.subject == exp_subject:
+            sent_to.append(_email.to[0])
```
MattHolmes123, line 26: Could be a list comp if you prefer one line over 4 

`send_to = [_email.to[0] for _email in outbox if _email.subject == exp_subject]`

MattHolmes123 (approved):

web/permissions/service.py
```diff
@@ -366,12 +367,34 @@ def get_users_with_permission(
...
+def get_all_case_officers() -> QuerySet[User]:
+    """Returns users who are case officers."""
 
     return get_users_with_permission(Perms.sys.ilb_admin)
 
 
+def get_ilb_case_officers() -> QuerySet[User]:
+    """Returns users who are ilb case officers."""
+
+    return get_all_case_officers().exclude(pk__in=get_sanctions_case_officers().values("pk"))
```
MattHolmes123, line 379: Does this work?
I would have thought it should be `.values_list("pk", flat=True)` as you're checking `pk_in`

MattHolmes123:
```
>>> all_users.exclude(pk__in=all_users.filter(username="ilb_admin").values("pk"))
<QuerySet [<User: Emilio Mcgee (SAN Case Officer)>, <User: Karen Bradley (SAN Case Officer)>, <User: Samantha Stevens (ilb_admin)>]>
>>> all_users.exclude(pk__in=all_users.filter(username="ilb_admin").values_list("pk", flat=True))
<QuerySet [<User: Emilio Mcgee (SAN Case Officer)>, <User: Karen Bradley (SAN Case Officer)>, <User: Samantha Stevens (ilb_admin)>]>
>>> 
```
Turns out you can use both... who knew.
Always supplied flat lists to calls to `xxx__in=` calls

MattHolmes123 (approved):

PR #1132, opened 2023-07-14
ICMSLST-2079 Data Migration Constabulary Contacts

mikeylockwood:


MattHolmes123:
Roy mentioned that nobody is currently using this feature.
So I wonder if we should in fact be migrating any of the actual users?
We do need to speak to ILB at some point to try and sort out getting constabulary contacts to use this feature again.

MattHolmes123 (approved):

data_migration/management/commands/post_migration.py
```diff
@@ -23,6 +27,9 @@ def apply_user_permissions(self) -> None:
             self.fetch_data(connection, "Importer User")
             self.fetch_data(connection, "Exporter User")
 
+            # TODO Constabulary contact permissions may not be migrated to V2
```
MattHolmes123, line 30: link to jira issue

PR #1138, opened 2023-07-21
ICMSLST-2134 Remove sidebar, radio and filter buttons from workbasket.

MattHolmes123:
Removed features that were copied from V1 that will not be implemented in the V2 workbasket.


MattHolmes123:
Before:
image

After:
image

marcuspp (approved):

PR #1139, opened 2023-07-24
ICMSLST-2125 Fix Clear action for view issued documents.

MattHolmes123:


web/domains/case/_import/models.py
```diff
@@ -424,6 +424,9 @@ class ImportApplicationLicence(DocumentPackBase):
     # Set when licence is marked active.
     case_completion_datetime = models.DateTimeField(verbose_name="Case Completion Date", null=True)
 
+    # Used to remove from workbasket when Clear action is performed.
+    cleared_by = models.ManyToManyField("web.User")
```
mikeylockwood, line 428: I'll add a ticket for the data migration to see if this can be populated

mikeylockwood (approved):

web/domains/case/_import/models.py
MattHolmes123: Yer good idea, lets look at it together.

PR #1140, opened 2023-07-24
ICMSLST-2076 Refactor PersonalEmail and AlternativeEmail into Email

mikeylockwood:
* Replace `PersonalEmail` and `AlternativeEmail` models with a combined `Email` model
* Currently in V1 there are 202 examples of Alternative Emails being used and 137 users with more than 1 Personal Email set


marcuspp (approved):

web/management/commands/add_test_data.py
```diff
@@ -269,7 +269,7 @@ def create_import_search_user(self, username):
 
 
 def add_personal_email(user):
```
MattHolmes123, line 271: Do you want to rename this to `def add_email`

MattHolmes123 (approved):

PR #1146, opened 2023-07-31
ICMSLST-2169 - Email Template model and admin screens

marcuspp:
- Introduces new `mail` app to help migrate away from `notify` app
- Adds EmailTemplate model and admin view
- Adds gov notify support (package and API key)
- Checks if Template ID is valid before saving admin model
- Adds initial templates as part of `load_data` process


web/mail/api.py
```diff
@@ -0,0 +1,29 @@
...
+def get_gov_notify_client() -> NotificationsAPIClient:
+    return NotificationsAPIClient(settings.GOV_NOTIFY_API_KEY)
+
+
+def get_template_by_id(template_id: UUID) -> dict:
+    client = get_gov_notify_client()
+    try:
+        return client.get_template(template_id)
+    except HTTPError as e:
+        return e.response.json()
```
MattHolmes123, line 17: Later on we could use something like pydantic to check this response and improve our code (so we don't have to use dicts everywhere.
This documents the keys coming back, their type and if they are required.
https://docs.notifications.service.gov.uk/python.html#get-a-template-by-id-response

MattHolmes123: It supports ignoring extra keys as well, if the API changes and they add stuff, our code wouldn't break.

MattHolmes123: Im thinking more when we have to use the templates, not here to check if a template id is valid.

web/mail/api.py
```diff
@@ -0,0 +1,29 @@
...
+def is_valid_template_id(template_id: UUID) -> bool:
+    response = get_template_by_id(template_id)
+    gov_notify_template_id = response.get("id", "")
+    try:
+        gov_notify_template_id = UUID(gov_notify_template_id)
+    except ValueError:
+        return False
+    if gov_notify_template_id == template_id:
+        return True
+    return False
```
MattHolmes123, line 29: `return gov_notify_template_id == template_id`

config/settings/base.py
```diff
@@ -129,8 +129,9 @@
 ]
 
 # Email
-# EMAIL_BACKEND = "django_ses.SESBackend" TODO ICMLST-1994
+GOV_NOTIFY_API_KEY = env.str("GOV_NOTIFY_API_KEY", default="")
```
MattHolmes123, line 132: Can you document `GOV_NOTIFY_API_KEY` in `.env.example`, it's the only reference a new dev would have when setting up the app for the first time.
e.g.
`GOV_NOTIFY_API_KEY=<FROM_VAULT>`

MattHolmes123 (approved):

mikeylockwood (approved):

PR #1149, opened 2023-08-02
ICMSLST-2135 - Gov Notify

marcuspp:
Initial implementation of Gov Notify
 - Gov notify backend added
 - Access requested email updated to use gov notify
 - Backend not set by default to GOV Notify this will be done once Gov Notify is 'Live'
 - Support added to override recipients of emails on dev/local to aid with manual testing
 - Logging handled by `django-celery-results`
 - Removed `django-ses`


requirements-base.txt
```diff
@@ -1,5 +1,5 @@
 boto3==1.21.29
-celery[redis]==5.2.7
+celery[redis]==5.3.1
```
marcuspp, line 2: Came across this issue during testing https://github.com/celery/billiard/issues/377 Upgrading celery fixed the issue

requirements-base.txt
```diff
@@ -26,6 +25,7 @@ notifications-python-client==8.0.1
 openpyxl==3.0.7
 oracledb==1.2.0
 phonenumbers==8.12.12
+pytz==2023.3
```
marcuspp, line 28: Was a dependancy of celery 5.2.7 and not 5.3.1. Is used by `add_commodity_data` script

web/mail/backends.py
```diff
@@ -0,0 +1,19 @@
...
+from web.mail.api import send_email
+
+logger = logging.getLogger(__name__)
+
+
+class GovNotifyEmailBackend(BaseEmailBackend):
+    def send_messages(self, email_messages: list) -> None:
+        for message in email_messages:
+            for recipient in message.recipients():
+                logger.info(f"Sending {message.name.label} email to {recipient}")
```
MattHolmes123, line 15: Logging should us the old %-style formatting.
It's for backward compatability mainly, but the formatting doesn't take place when logging is disabled if you do it the %-style way.
(I know a few f-string formatting lines have snuck in to ICMS).

web/mail/backends.py
```diff
@@ -0,0 +1,19 @@
...
+import structlog as logging
+from django.core.mail.backends.base import BaseEmailBackend
+
+from web.mail.api import send_email
+
+logger = logging.getLogger(__name__)
+
+
+class GovNotifyEmailBackend(BaseEmailBackend):
+    def send_messages(self, email_messages: list) -> None:
```
MattHolmes123, line 12:
```
from web.mail.messages import GOVNotifyEmailMessage
def send_messages(self, email_messages: list[GOVNotifyEmailMessage]) -> None:
```

web/mail/messages.py
```diff
@@ -0,0 +1,48 @@
...
+        } | self.get_context()
+
+    def get_template_id(self) -> UUID:
+        return EmailTemplate.objects.get(name=self.name).gov_notify_template_id
+
+
+class AccessRequestEmail(GOVNotifyEmailMessage):
+    name = EmailTypes.ACCESS_REQUEST
+
+    def __init__(self, access_request, *args, **kwargs):
```
MattHolmes123, line 43: Any reason you've moved away from doing it as a named kwarg?

e.g.
```python
def __init__(self, *args, access_request: AccessRequest, **kwargs):
```

web/mail/messages.py
```diff
@@ -0,0 +1,48 @@
...
+
+from django.conf import settings
+from django.core.mail import EmailMessage, SafeMIMEMultipart
+
+from .constants import EmailTypes
+from .models import EmailTemplate
+
+
+class GOVNotifyEmailMessage(EmailMessage):
+    name = None | EmailTypes
```
MattHolmes123, line 11:
```python
from typing import ClassVar
name: ClassVar[EmailTypes | None] = None
```
In your version name get assigned a value of `<class 'types.UnionType'>`

web/mail/messages.py
```diff
@@ -0,0 +1,48 @@
...
+        return {
+            "icms_url": settings.DEFAULT_DOMAIN,
+            "icms_contact_email": settings.ILB_CONTACT_EMAIL,
+            "icms_contact_phone": settings.ILB_CONTACT_PHONE,
+            "subject": self.subject,
+            "body": self.body,
+        } | self.get_context()
+
+    def get_template_id(self) -> UUID:
+        return EmailTemplate.objects.get(name=self.name).gov_notify_template_id
```
MattHolmes123, line 37: Don't think it will be a problem but this will hit the db everytime any derived class is initialised.

MattHolmes123 (approved):

PR #1150, opened 2023-08-08
Data migration fix licence and certificate document filenames

mikeylockwood:
- Add a check for files with duplicate paths
- Fix queries for data migration licence and certificates so pdf file extension is included in filenames

### DIFF
--- a/data_migration/management/commands/_types.py
+++ b/data_migration/management/commands/_types.py
@@ -9,6 +9,8 @@ from data_migration import models as dm
 from web import models as web
 
 ModelT = Union[T[Model], list[T[Model]]]
+Anno = dict[str, Any] | None
+Val = list[str] | None
 Params = dict[str, int | str | bool | tuple]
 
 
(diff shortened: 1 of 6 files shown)


data_migration/management/commands/_types.py
```diff
@@ -9,6 +9,8 @@
 from web import models as web
 
 ModelT = Union[T[Model], list[T[Model]]]
+AnnoT = dict[str, Any] | None
+ValT = list[str] | None
```
MattHolmes123, line 13: I would drop t `T` on your types here.
They document instances instead of types I think right?
`ModelT` is correct

MattHolmes123: Yep just checked annotation and values are instance values where model is a class 
```
    CheckCount(
        name="Duplicate File Paths",
        expected_count=0,
        model=web.File,
        filter_params={"path_count__gt": 1},
        annotation={"path_count": Count("pk")},
        values=["path"],
    ),
```

PR #1154, opened 2023-08-11
ICMSLST-2065 Add staff-sso-client to ICMS.

MattHolmes123:
Changes:
  - Add staff-sso-client.
  - Added environment variable to enable / disable staff-sso integration.
  - Set correct backend when calling User.objects.with_perm to fix ValueError exception being raised when multiple authentication backends are set.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -21,10 +21,6 @@ env = environ.Env()
 
 VCAP_SERVICES = env.json("VCAP_SERVICES", default={})
 
-LOGIN_URL = "accounts:login"
-LOGIN_REDIRECT_URL = "workbasket"
-LOGOUT_REDIRECT_URL = "accounts:login"
-
 # Application definition
 DEBUG = env.bool("ICMS_DEBUG", False)
 WSGI_APPLICATION = "config.wsgi.application"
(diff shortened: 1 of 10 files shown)


config/settings/base.py
```diff
@@ -126,8 +124,31 @@
...
+LOGIN_REDIRECT_URL = "workbasket"
+
+#
+# STAFF-SSO client app settings
+AUTHBROKER_URL = env.str("STAFF_SSO_AUTHBROKER_URL", default="")
+AUTHBROKER_CLIENT_ID = env.str("STAFF_SSO_AUTHBROKER_CLIENT_ID", default="")
+AUTHBROKER_CLIENT_SECRET = env.str("STAFF_SSO_AUTHBROKER_CLIENT_SECRET", default="")
+AUTHBROKER_STAFF_SSO_SCOPE = env.str("STAFF_SSO_AUTHBROKER_STAFF_SSO_SCOPE", default="")
+AUTHBROKER_ANONYMOUS_PATHS = env.list("STAFF_SSO_AUTHBROKER_ANONYMOUS_PATHS", default=[])
+AUTHBROKER_ANONYMOUS_URL_NAMES = env.list("STAFF_SSO_AUTHBROKER_ANONYMOUS_URL_NAMES", default=[])
```
MattHolmes123, line 149: These probably don't need to come from vault but I don't think we will use this feature anyway.

mikeylockwood (approved):

marcuspp (approved):

PR #1155, opened 2023-08-11
ICMSLST-2165 - Access Request Closed (Gov Notify) Email

marcuspp:
- Converted Access Request Closed email to use Gov Notify
 - Add a command to resend an email for a given task id


web/domains/case/access/models.py
```diff
@@ -121,6 +121,10 @@ class ImporterAccessRequest(AccessRequest):
     def is_agent_request(self):
         return self.request_type == self.AGENT_ACCESS
 
+    @property
+    def is_importer_request(self):
+        return True
```
mikeylockwood, line 126: Is this just to set `request_type` in the context? Could just set a `REQUEST_TYPE = "importer"` constant on the model

web/mail/recipients.py
```diff
@@ -12,7 +12,7 @@ def get_ilb_case_officers_email_addresses() -> list[str]:
 
 
 @override_recipients
-def get_email_addresses_for_users(users: QuerySet[User]) -> list[str]:
+def get_email_addresses_for_users(users: QuerySet[User] | list[User]) -> list[str]:
```
mikeylockwood, line 15: `Collection[User]`

mikeylockwood (approved):

web/management/commands/resend_email.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--task_id",
+            type=UUID,
+            help="",
+            required=True,
+        )
+
+    def get_task(self, task_id) -> TaskResult:
```
MattHolmes123, line 23: `task_id: int`

PR #1159, opened 2023-08-16
ICMSLST-2211 Add ICMSStaffSSOBackend class.

MattHolmes123:
Changes:
  - Migrates legacy users after successful staff-sso login.
  - ICMSLST-2212 Fix missing Email record for staff-sso users.

### DIFF
--- a/web/auth/backends.py
+++ b/web/auth/backends.py
@@ -1,10 +1,14 @@
+from authbroker_client.backends import AuthbrokerBackend
 from django.contrib.auth.backends import ModelBackend
 from guardian.backends import check_support
 from guardian.conf import settings as guardian_settings
 from guardian.ctypes import get_content_type
 from guardian.exceptions import WrongAppError
 
+from web.models import Email as UserEmail
 from web.models import User
+
+from .types import STAFF_SSO_ID, StaffSSOProfile
 
 
 class ModelAndObjectPermissionBackend(ModelBackend):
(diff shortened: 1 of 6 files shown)


web/auth/backends.py
```diff
@@ -87,6 +91,53 @@ def get_user_permissions(self, user_obj, obj=None):
...
+        id_key: STAFF_SSO_ID = self.get_profile_id_name()
+        id_value = profile[id_key]
+
+        user_data = self.user_create_mapping(profile)
+        staff_sso_email = user_data["email"]
+        update_email = False
+
+        try:
+            # A legacy user is a user who has an email address as a username.
+            user = User.objects.get(**{User.USERNAME_FIELD: staff_sso_email})
```
MattHolmes123, line 112: Legacy users who plan to use a different email for staff-sso compared to their icms email will find they get a new user created rather than updating their existing user account with a different email.

web/auth/backends.py
```diff
@@ -87,6 +91,53 @@ def get_user_permissions(self, user_obj, obj=None):
...
+            if created:
+                user.set_unusable_password()
+                user.save()
+                update_email = True
+
+        if update_email:
+            setattr(user, User.EMAIL_FIELD, staff_sso_email)
+            user.save()
+
+            UserEmail.objects.get_or_create(user=user, email=staff_sso_email)
```
marcuspp, line 136: Left a comment on slack should this set the email address to primary?

marcuspp (approved):

web/auth/backends.py
```diff
@@ -87,6 +91,53 @@ def get_user_permissions(self, user_obj, obj=None):
...
+
+    def get_or_create_user(self, profile: StaffSSOProfile) -> User:
+        id_key: STAFF_SSO_ID = self.get_profile_id_name()
+        id_value = profile[id_key]
+
+        user_data = self.user_create_mapping(profile)
+        staff_sso_email = user_data["email"]
+        update_email = False
+
+        try:
```
marcuspp, line 110: Is it worth creating a reusable function to create a user? So that it can be use regardless of which Auth Backend is used?

MattHolmes123: I will no doubt refactor this backend when I add the one login version.
Didn't want to refactor before though

PR #1161, opened 2023-08-17
ICMSLST-2197 Add login-start/ and logout/ endpoints.

MattHolmes123:
Changes:
  - Add login start view to direct users to correct login page.
  - Add logout view to logout using correct authentication backend.
  - Add context_processor for authentication template context.
  - Add placeholder gov.uk one login authentication code.
  - Refactor templates to use new global auth context.
  - Refactor logout links to be post only.


config/settings/base.py
```diff
@@ -135,22 +135,20 @@
 AUTHBROKER_ANONYMOUS_URL_NAMES = env.list("STAFF_SSO_AUTHBROKER_ANONYMOUS_URL_NAMES", default=[])
 
 
-# TODO: ICMSLST-2197 This logic will be replaced by ICMS login view.
 STAFF_SSO_ENABLED = env.bool("STAFF_SSO_ENABLED", True)
+GOV_UK_ONE_LOGIN_ENABLED = env.bool("GOV_UK_ONE_LOGIN_ENABLED", True)
```
MattHolmes123, line 139: REMINDER FOR ME: Update vault and set to false.

web/views/views.py
```diff
@@ -1,17 +1,93 @@
...
+    backend = request.session.get(auth.BACKEND_SESSION_KEY, "")
+
+    # 2. Clear session and user
+    auth.logout(request)
+
+    # 3. Using user backend, redirect to correct auth logout view.
+    login_start = reverse("login-start")
+    match backend:
+        case "web.auth.backends.ICMSStaffSSOBackend":
+            url = urljoin(settings.AUTHBROKER_URL, "logout/")
```
MattHolmes123, line 76: There are questions being asked about whether we shoud be logging the user out of staff-sso when logging out of ICMS.
However gov.uk one login insists we should log the users out of both the application and one login: 
https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/log-your-users-out/#log-your-users-out-of-gov-uk-one-login

So for now I have done the same thing for staff-sso (redirect to the staff sso logout endpoint after logging the user out of ICMS)

web/views/views.py
```diff
@@ -1,17 +1,93 @@
...
+
+    context = {
+        "staff_sso_login_url": staff_sso_login_url,
+        "one_login_login_url": one_login_login_url,
+    }
+
+    return render(request, "login_start.html", context)
+
+
+@require_POST
```
MattHolmes123, line 63: https://docs.djangoproject.com/en/4.1/releases/4.1/#log-out-via-get

web/views/views.py
```diff
@@ -1,17 +1,93 @@
...
+            self.url = reverse("login-start")
+
+        return super().get_redirect_url(*args, **kwargs)
+
+
+def login_start_view(request: HttpRequest) -> HttpResponse:
+    """Custom login view to handle redirecting to correct authentication backend."""
+
+    # Default login for environments where staff-sso or one-login is disabled.
+    django_model_auth_login = reverse("accounts:login")
```
marcuspp, line 42: Its this line that raises the error if both AUTH backends are enabled

MattHolmes123: yer thats a bug! Thanks :)

web/views/views.py
```diff
@@ -1,17 +1,92 @@
...
+
+    if settings.STAFF_SSO_ENABLED:
+        staff_sso_login_url = reverse("authbroker_client:login")
+    else:
+        # Default login for environments where staff-sso is disabled.
+        staff_sso_login_url = reverse("accounts:login")
+
+    if settings.GOV_UK_ONE_LOGIN_ENABLED:
+        # TODO: ICMSLST-2196 Replace with gov.uk one login url
+        one_login_login_url = reverse("authbroker_client:login")
```
MattHolmes123, line 49: This is still a bug if `GOV_UK_ONE_LOGIN_ENABLED` is set to `True` and `STAFF_SSO_ENABLED` is set to `False` but it's only a placeholder value until we actually implement the one login backend.

marcuspp (approved):

mikeylockwood (approved):

PR #1176, opened 2023-09-07
ICMSLST-1406 Generate CFS Schedule PDFs

mikeylockwood:
- Generate a document for each country being exported to
- Create a schedule in the document for each schedule in the application
- Still todo digital signatures, QR verification, translations
- Schedule paragraphs to be taken from templates (ICMSLST-1913)
- Applicant cannot view final case documents due to permissions (ICMSLST-2297)


marcuspp (approved):

web/domains/case/views/views_misc.py
```diff
@@ -754,25 +754,23 @@ def get_document_context(
...
-            "document_reference": document_reference,
+            "certificate_docs": certificate_docs,
+            "certificate_pk": certificate.pk,
+            "is_import": False,
             # TODO: Revisit when we can generate an export certificate
-            # https://uktrade.atlassian.net/browse/ICMSLST-1406
             # https://uktrade.atlassian.net/browse/ICMSLST-1407
             # https://uktrade.atlassian.net/browse/ICMSLST-1408
-            "certificate_links": [],
         }
```
MattHolmes123, line 774: This needs reverting to use `certificate_links` and not hard code urls in the template.
Look at the import block, this is used for both the pre-sign and the final view (so the links change).
An exception is being thrown currenntly when viewing the "pre-sign" version of the screen as it's trying to fetch documents from S3 that dont exist yet.

MattHolmes123 (changes requested):

web/domains/case/tasks.py
```diff
@@ -91,11 +91,42 @@ def create_import_application_document(
 def create_export_application_document(
     application_pk: int, certificate_pk: int, casedocumentreference_pk: int, user_pk: int
 ) -> None:
-    # TODO: Revisit when we can generate an export certificate:
-    # https://uktrade.atlassian.net/browse/ICMSLST-1406
-    # https://uktrade.atlassian.net/browse/ICMSLST-1407
-    # https://uktrade.atlassian.net/browse/ICMSLST-1408
-    print("*************************** SKIPPING FOR NOW")
+    with transaction.atomic():
```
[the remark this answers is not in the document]
MattHolmes123: Suppose you'd move your comment below in to the if block as well.
# TODO: Revisit when we can generate an export certificate:
# https://uktrade.atlassian.net/browse/ICMSLST-1407
# https://uktrade.atlassian.net/browse/ICMSLST-1408

web/static/web/css/pdfs/export-certificate-style.css
```diff
@@ -0,0 +1,168 @@
...
+}
+
+.product-header {
+    margin-bottom: 2px;
+}
+
+.product-name {
+    margin-top: 2px;
+    margin-bottom: 2px;
+}
```
MattHolmes123, line 168: Missing newline at end of the file.

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #1179, opened 2023-09-12
ICMSLST-1407 Generate COM Application PDF

mikeylockwood:


web/tests/utils/pdf/test_generator.py
```diff
@@ -318,6 +329,19 @@ def test_get_preview_cfs_certificate_context(cfs_app_submitted):
     assert context["reference"] == "[[CERTIFICATE_REFERENCE]]"
 
 
+def test_get_preview_com_certificate_context(cfs_app_submitted):
```
mikeylockwood, line 332: Test should use a com app not a cfs app

MattHolmes123 (approved):

PR #1180, opened 2023-09-14
ICMSLST-2196 Add django.contrib.sites package to support several sites.

MattHolmes123:
Features can now be enabled / disabled depending on the URL.
Importer and Exporter access requests have been used as an example.

Other changes:
  - Add data migration to create sites
  - Add script to set ICMS site urls.


web/sites/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+            return f(request, *args, **kwargs)
+
+        return _wrapped_view
+
+    return decorator
+
+
+def is_exporter_site(site: Site) -> bool:
+    return site.name == EXPORTER_SITE_NAME
```
marcuspp, line 53: The site name can be changed in django admin. Maybe consider using domain and make it uneditable in django admin.

marcuspp (approved):

web/sites/__init__.py
MattHolmes123: Domain is unique, we may want multiple domains to mean the same thing "An Exporter Site".

MattHolmes123: I considered updating the admin interface to check name was one of the 3 supported values in a custom form.

MattHolmes123: https://uktrade.atlassian.net/browse/ICMSLST-2318

mikeylockwood (approved):

PR #1183, opened 2023-09-19
ICMSLST-1913 Fetch schedule paragraphs from CFSScheduleParagraph

mikeylockwood:


web/utils/string.py
```diff
@@ -0,0 +1,43 @@
+import datetime as dt
+import re
+
+
```
MattHolmes123, line 4: I don't think we need this file. 
`day_ordinal_date` and `split_text_field_newlines` are both only used in pdf/utils.py so should be fine to live there.
Also `clean_address` should be called something like `strip_extra_spaces` or something if it was to live in a generic file called `string.py`.
In either case I would be tempted to just move it back to pdf/utils.py

mikeylockwood: I use clean_address in both `template/context.py`  and  `pdf/utils.py` so would like a generic place to import it from rather than defining it twice

web/domains/template/utils.py
```diff
@@ -194,3 +198,65 @@ def get_fir_template_data(process: Process, current_user: User) -> tuple[str, st
...
+            paragraph += content(paragraph_names.GOOD_MANUFACTURING_PRACTICE)
+
+    if schedule.manufacturer_name and schedule.manufacturer_address:
+        paragraph += content(paragraph_names.COUNTRY_OF_MAN_STATEMENT_WITH_NAME_AND_ADDRESS, False)
+    elif schedule.manufacturer_name:
+        paragraph += content(paragraph_names.COUNTRY_OF_MAN_STATEMENT_WITH_NAME, False)
+    else:
+        paragraph += content(paragraph_names.COUNTRY_OF_MAN_STATEMENT, False)
+
+    return paragraph
```
MattHolmes123, line 256: It feels like this should go the following.
1. Get a list of strings 
2. Return " ".join(values)

The space param is just a bit odd when the join function is good at adding values between every element in a list.
Also get_schedule_paragraph_content and content are essentially the same function with a different default value.
Just feels a bit busy this version.

MattHolmes123 (approved):

web/utils/string.py
MattHolmes123: The name string.py is my problem then.
Just put it in web/utils/__init__.py

then your import is `from web.utils import clean_address`

marcuspp: Just to add something, I'm not a fan of files/libraries being called the same thing as existing inbuilt python libraries.

PR #1184, opened 2023-09-20
ICMSLST-2318 Only allow valid name values when configuring site records.

MattHolmes123:


MattHolmes123:
image
Changed to a select list to only allow valid values.

mikeylockwood (approved):

PR #1185, opened 2023-09-20
ICMSLST-2145 - Constabulary Deactivated Firearms Email

marcuspp:
- Email includes a link to the documents (not attachments)
- Includes removal of notify app


web/mail/messages.py
```diff
@@ -284,6 +285,19 @@ class FirearmsSupplementaryReportEmail(BaseApplicationEmail):
     name = EmailTypes.FIREARMS_SUPPLEMENTARY_REPORT
 
 
+@final
+class ConstabularyDeactivatedFirearmsEmail(BaseApplicationEmail):
+    name = EmailTypes.CONSTABULARY_DEACTIVATED_FIREARMS
+
+    def get_context(self) -> dict:
+        context = super().get_context()
+        context["documents_url"] = get_document_view_url(self.application, full_url=True)
```
MattHolmes123, line 294: I don't think a constabulary contact will be able to view the case based on their permissions.
In V1 they can't access the view application page in ICMS.
This function is what is used to check if a user can view that case history view:
icms/web/permissions/service.py -> def can_view

If a constabulary contact clicks the documents_url it will raise a PermissionDenied error.

MattHolmes123 (changes requested):

MattHolmes123 (approved):
Add the todo for the new story and this can be merged.

mikeylockwood (approved):

PR #1189, opened 2023-09-22
ICMSLST-2314 Remove invalid DEFAULT_DOMAIN setting.

MattHolmes123:
All emails use the caseworker domain by default, this will change in ICMSLST-2313.


MattHolmes123:
The ICMS_DEFAULT_DOMAIN key will be removed from vault prior to merging.

marcuspp (approved):

PR #1193, opened 2023-09-26
ICMSLST-2313 - ICMS correct domain within emails

marcuspp:
- Groups site names into SiteName typed choice collection
- Sets url domains for each email type
- Sets static (validate pdf doc) to use importer domain


web/permissions/service.py
```diff
@@ -430,3 +430,7 @@ def filter_users_with_org_access(org: ORGANISATION, users: QuerySet[User]) -> Qu
 
     # Note: We are ignoring user permissions by design
     return users.filter(groups__permissions__codename=org_access).distinct()
+
+
+def is_user_a_case_officer(user: User) -> bool:
+    return user.has_perm(Perms.sys.ilb_admin)
```
MattHolmes123, line 436: Can we remove this please and just call has_perm directly.

web/tests/domains/case/views/test_views_variation_request.py
```diff
@@ -166,14 +172,14 @@ def set_client(self, ilb_admin_client):
...
-    def set_app(self, set_client, com_app_submitted, ilb_admin_user):
+    def set_app(self, set_client, com_app_submitted, exporter_one_contact):
         self.app = com_app_submitted
         self.client.post(CaseURLS.take_ownership(self.app.pk))
 
         self.app.refresh_from_db()
         self.app.status = ImpExpStatus.VARIATION_REQUESTED
         add_variation_request_to_app(
-            self.app, ilb_admin_user, status=VariationRequest.Statuses.OPEN
+            self.app, exporter_one_contact, status=VariationRequest.Statuses.OPEN
```
MattHolmes123, line 182: This doesn't seem valid, An export variation request can only be opened by a caseworker?
In this test calling `add_variation_request_to_app` with the `exporter_one_contact` sets that user in the `requested_by` field.

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #1201, opened 2023-10-02
ICMSLST-1779 Fix RadioSelectInline css to render elements inline.

MattHolmes123:
The default render changed from Form.as_ul to Form.as_div in django 4.1.


MattHolmes123:
Examples before fix:
image
image

Examples after fix:
image
image

mikeylockwood (approved):

PR #1208, opened 2023-10-05
ICMSLST-1640 Remove Search Mailshots feature from ILB Admin group.

MattHolmes123:


web/domains/mailshot/views.py
```diff
@@ -256,7 +259,7 @@ def has_permission(self) -> bool:
         if mailshot.is_to_importers and user.has_perm(Perms.sys.importer_access):
             return True
 
-        if mailshot.is_to_exporters and user.has_perm("web.export_access"):
+        if mailshot.is_to_exporters and user.has_perm(Perms.sys.exporter_access):
```
MattHolmes123, line 262: Unrelated to feature but it was a bug so I fixed it.
`"web.export_access"` doesn't mean anything

mikeylockwood (approved):

PR #1225, opened 2023-10-17
ICMSLST-1831 CFS Schedule PDF Translations

mikeylockwood:
Adding translations to the CFS PDF certificates using the partial implementation that has already been added as well as the data in V1.

CFS_DECLARTION_TRANSLATION templates are used to fetch the translation for the declaration

CFS_SCHEDULE_TRANSLATION templates are used to fetch the paragraphs that are translated within the cfs schedule

### DIFF
--- a/web/domains/template/utils.py
+++ b/web/domains/template/utils.py
@@ -1,5 +1,6 @@
 import dataclasses
 import re
+from collections.abc import Iterator
 from typing import TYPE_CHECKING
 
 from web.domains.case.types import ImpOrExp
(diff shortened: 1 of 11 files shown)


web/management/commands/utils/add_template_data.py
```diff
@@ -109,55 +110,113 @@ def remove_templates():
...
         application_domain="CA",
-    )
-    Template.objects.get_or_create(
+        template_content=(
+            "Pelo presente documento certifica-se, para os fins das leis do pais acima referido, "
+            "que os preparados designados no anexo, que faz parte integrante do presente certificado, "
+            "podem ser vendidos legalmente no Reino Unido desde que sejam satisfeitos os requisitos regulamentários."
+        ),
+    )[0]
+    portuguese.countries.add(*Country.objects.filter(name__in=["Brazil"]))
```
marcuspp, line 183: Portugal?

web/management/commands/utils/add_template_data.py
```diff
@@ -167,6 +226,100 @@ def add_schedule_translation_templates():
         template_type="CFS_SCHEDULE_TRANSLATION",
         application_domain="CA",
         country_translation_set=CountryTranslationSet.objects.get(name="Spanish"),
+    )[0]
+
+    t.countries.add(
+        *Country.objects.filter(
+            name__in=[
```
marcuspp, line 233: Maybe these groups of countries should be a setting or a constant

web/management/commands/utils/add_template_data.py
mikeylockwood: The export certificates aren't for countries within the EU

web/management/commands/utils/add_template_data.py
mikeylockwood: I'm just going to leave this as it as for now as it is just to create test data. In V1, the country lists for the cfs declaration translation and the cfs schedule paragraph translation are independant of each other

web/domains/template/utils.py
```diff
@@ -200,56 +202,122 @@ def get_fir_template_data(process: Process, current_user: User) -> tuple[str, st
...
+    product: str | None = None
+
+    def populate(self) -> "ScheduleParagraphs":
+        context = ScheduleParagraphContext(self.schedule, self.template.country_translation_set)
+        names = CFSScheduleParagraph.ParagraphName
+        self.header = self.content(names.SCHEDULE_HEADER, context)
+        self.introduction = self.content(names.SCHEDULE_INTRODUCTION, context)
+        self.create_paragraph(context)
+        self.product = self.content(names.PRODUCTS, context)
+        return self
```
MattHolmes123, line 221: I found this a bit strange when I saw it, I was initially going to just say "return Self from the typing modlue".
But what you really need is some post initialisation code, with dataclasses that is handled in this method:
```
def __post_init__(self) -> None:
    # Do stuff
    ...
```

Then later where you create your classes they can just be defined like this:
```
self.english_paragraphs = ScheduleParagraphs(self.schedule, template)
...
self.translation_paragraphs = ScheduleParagraphs(self.schedule, template)
```

web/domains/template/utils.py
```diff
@@ -200,56 +202,122 @@ def get_fir_template_data(process: Process, current_user: User) -> tuple[str, st
...
 
-    if schedule.schedule_statements_accordance_with_standards:
-        if is_ni:
-            paragraphs.append(content(paragraph_names.GOOD_MANUFACTURING_PRACTICE_NI))
-        else:
-            paragraphs.append(content(paragraph_names.GOOD_MANUFACTURING_PRACTICE))
+    def populate(self):
+        self.get_english_paragraphs()
+        self.get_translation_paragraphs()
+        return self
```
MattHolmes123, line 288: same here

web/domains/template/utils.py
```diff
@@ -200,56 +202,122 @@ def get_fir_template_data(process: Process, current_user: User) -> tuple[str, st
...
+            )
+            self.translation_paragraphs = ScheduleParagraphs(self.schedule, template).populate()
+        except Template.DoesNotExist:
+            pass
 
 
-def fetch_schedule_paragraphs(application: CertificateOfFreeSaleApplication) -> dict[int, str]:
+def fetch_schedule_text(
+    application: CertificateOfFreeSaleApplication, country: Country
+) -> dict[int, str]:
```
MattHolmes123, line 308: this return type is wrong.

web/management/commands/utils/add_template_data.py
```diff
@@ -109,55 +110,113 @@ def remove_templates():
 
 
 def add_cfs_declaration_templates():
-    Template.objects.get_or_create(
+    spanish = Template.objects.get_or_create(
```
MattHolmes123, line 113: `spanish, _` is more obvious than doing `[0]` later

MattHolmes123 (approved):

PR #1227, opened 2023-10-18
ICMSLST-1268 - Remove CC list from FIR

marcuspp:


MattHolmes123 (approved):
Lets see what Michael says about the field in the migration.

mikeylockwood (approved):

PR #1230, opened 2023-10-20
ICMSLST-2376 - Authority archived email

marcuspp:


mikeylockwood (approved):

web/mail/messages.py
```diff
@@ -455,3 +468,33 @@ def get_context(self) -> dict:
...
+        context = super().get_context()
+        context["authority_name"] = self.authority.reference
+        context["authority_type"] = self.authority.AUTHORITY_TYPE
+        context["authority_url"] = get_authority_view_url(self.authority, full_url=True)
+        context["date"] = timezone.now().strftime("%-d %B %Y")
+        context["importer_url"] = get_importer_view_url(self.authority.importer, full_url=True)
+        context["importer_name"] = self.authority.importer.name
+        context["reason"] = self.get_reason()
+        context["reason_other"] = self.authority.other_archive_reason or ""
+        return context
```
MattHolmes123, line 491: I'd probably do this to reduce the amount of times I wrote `context`:
```python
def get_context(self) -> dict:
    context = super().get_context()
    
    return context | {
        "foo": bar",
    }
```

PR #1239, opened 2023-10-30
ICMSLST-2389 - Authority Expiring emails

marcuspp:


config/celery.py
```diff
@@ -13,12 +15,12 @@
 
 app.conf.beat_schedule = {
     # Only enable when constabulary contacts manage firearms authorities
-    # "firearms_authority_expiring": {
-    #     "task": "web.notify.notify.send_firearms_authority_expiry_notification",
+    # "authority_expiring_firearms_email": {
```
marcuspp, line 18: Left commented out as comment above still true

web/mail/emails.py
```diff
@@ -1,3 +1,5 @@
+from datetime import date
```
MattHolmes123, line 1: Can you import this as `import datetime` or `import datetime as dt` please.
I have had problems in the past importing datetime / date directly from the datetime module.

MattHolmes123: there isn't 100% consistency in ICMS but the majority of imports import the whole module.

MattHolmes123 (approved):

web/mail/messages.py
```diff
@@ -503,6 +507,62 @@ def get_site_domain(self) -> str:
...
+        self.authority_type = authority_type
+        self.expiry_date = expiry_date
+
+    def get_context(self) -> dict:
+        context = super().get_context()
+        return context | {
+            "importers_count": len(self.importers_details),
+            "authority_type": self.authority_type,
+            "summary_text": self.get_summary_text(),
+            "expiry_date": self.expiry_date.strftime("%-d %B %Y"),
```
mikeylockwood, line 530: We use this `"%-d %B %Y"` format in multiple places. That string should probably be a constant

mikeylockwood (approved):

PR #1247, opened 2023-11-08
ICMSLST-1935 Add postcode form validation to importer office views.

MattHolmes123:
CHIEF/CDS requires trader postcodes to be valid and will error if not. Add form validation to reduce the number of errors raised in this way.

### DIFF
--- a/web/domains/chief/serializers.py
+++ b/web/domains/chief/serializers.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Any, Literal, Protocol
 
 from web.domains.case.services import document_pack
 from web.flow.models import ProcessTypes
+from web.forms.utils import clean_postcode
 from web.utils.commodity import annotate_commodity_unit
 
 from . import types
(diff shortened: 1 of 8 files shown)


web/domains/chief/serializers.py
```diff
@@ -305,7 +306,7 @@ def _get_organisation(application: "ImportApplication") -> types.OrganisationDat
         address=types.AddressData(
             # max address lines is 5
             **address_lines,
-            postcode=office.postcode,
+            postcode=clean_postcode(office.postcode),
```
MattHolmes123, line 309: Added in case a V1 importer is sending data to chief (their postcode might not be uppercase & without spaces)

web/domains/exporter/forms.py
```diff
@@ -63,6 +64,9 @@ def save(self, commit=True):
             locality = office_address.get("locality")
             postcode = office_address.get("postal_code")
 
+            if postcode:
+                postcode = clean_postcode(postcode)
```
MattHolmes123, line 68: Companies house postcode so no need to do regex checking for valid postcode (we only want to make sure its formatted correctly)

web/domains/importer/forms.py
```diff
@@ -98,6 +99,9 @@ def save(self, commit=True):
             locality = office_address.get("locality")
             postcode = office_address.get("postal_code")
 
+            if postcode:
+                postcode = clean_postcode(postcode)
```
MattHolmes123, line 103: Companies house postcode so no need to do regex checking for valid postcode (we only want to make sure its formatted correctly)

mikeylockwood (approved):

PR #1249, opened 2023-11-09
ICMSLST-697 Digital Signature for Licence and Certificate PDFs

mikeylockwood:
* Add digital signatures to licence and certificate pdfs in document signing step
* Currently the signing certificate comes from a base64 encoded string stored in the env. Can consider other options for storing the certificate.


requirements-base.txt
```diff
@@ -21,12 +21,13 @@ gunicorn==20.1.0
...
 Jinja2==3.0.3
-lxml==4.9.1
+lxml==4.9.3
 mohawk==1.1.0
 notifications-python-client==8.0.1
 openpyxl==3.0.7
-oracledb==1.2.0
+oracledb==1.4.2
 phonenumbers==8.12.12
+pyhanko==0.20.1
```
kevbradwick-dbt, line 30: Reading the docs, it says the library is not production ready. It's good to prove it work with this lib though. We probably need to investigate endesive for an alternative implementation

marcuspp (approved):

PR #1251, opened 2023-11-13
ICMSLST-2393 - View for constabularies to see issued documents

marcuspp:
To test the view
 - Create a DFL application setting constabulary as Nottingham
 - Approve application
 - Click link in  Automatic Notification email (login as con_user)
 - View and download documents

### DIFF
--- a/web/domains/case/urls.py
+++ b/web/domains/case/urls.py
@@ -4,6 +4,7 @@ from web.types import DocumentTypes
 
 from .views import (
     views_case_history,
+    views_documents,
     views_email,
     views_fir,
     views_misc,
@@ -427,6 +428,16 @@ urlpatterns = [
                                 views_case_history.CaseHistoryView.as_view(),
                                 name="ilb-case-history",
                                 kwargs={"mode": "ilb"},
+                            ),
+                            path(
+                                "documents/<int:doc_pack_pk>/",
+                                views_documents.ConstabularyDocumentView.as_view(),
+                                name="constabulary-doc",
+                            ),
+                            path(
+                                "documents/<int:doc_pack_pk>/download/<int:cdr_pk>/",
+                                views_documents.ConstabularyDocumentDownloadView.as_view(),
+                                name="constabulary-doc-download",
                             ),
                             path(
                                 "applicant-case-history/",
--- a/web/domains/case/views/views_documents.py
+++ b/web/domains/case/views/views_documents.py
@@ -1,0 +1,124 @@
+from typing import Any
+
+from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
+from django.db.models import QuerySet
+from django.http import HttpResponse
+from django.shortcuts import get_object_or_404
+from django.views.generic import DetailView
+from guardian.shortcuts import get_objects_for_user
+
+from web.domains.case.types import DocumentPack
+from web.domains.case.utils import get_case_page_title
+from web.flow.models import ProcessTypes as pt
+from web.mail.url_helpers import get_constabulary_document_download_view_url
+from web.models import (
+    CaseDocumentReference,
+    Constabulary,
+    ImportApplication,
+    ImportApplicationLicence,
+)
+from web.permissions import Perms
+from web.utils.s3 import get_file_from_s3
+
+
+class BaseConstabularyDocumentView(PermissionRequiredMixin, LoginRequiredMixin):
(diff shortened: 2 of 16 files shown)


web/domains/case/views/views_documents.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def is_application_linked_to_constabularies(
+        self, constabularies: QuerySet[Constabulary]
+    ) -> bool:
+        application = self.get_object().get_specific_model()
+        match application.process_type:
+            case pt.FA_DFL:
+                if application.constabulary not in constabularies:
+                    return False
+                return True
```
MattHolmes123, line 49: This could be one line
`return application.constabulary in constabularies`

web/domains/case/views/views_documents.py
```diff
@@ -0,0 +1,124 @@
...
+            issued_document.document_references,
+            document_type=CaseDocumentReference.Type.COVER_LETTER,
+        )
+
+    def get_document_context_data(self, application: ImportApplication) -> dict[str, Any]:
+        at = application.application_type
+        all_issued_documents = document_pack.pack_issued_get_all(application)
+        issued_document = all_issued_documents.first()
+        if not issued_document:
+            raise Http404()
```
MattHolmes123, line 87: `pack_issued_get_all` returns all the document packs in order.
I don't think we should just return the first one because what happens when there is a variation request?
They will see the old documents.
We should include an `issued_document_pk` in the url in the email we send out.

```python
all_issued_documents = document_pack.pack_issued_get_all(application)
pack = issued_documents.get(pk=self.kwargs["issued_document_pk"])

licence_cdr = document_pack.doc_ref_licence_get(pack)
cover_letter_cdr = document_pack.doc_ref_cover_letter_get(pack)
```

web/domains/case/views/views_documents.py
```diff
@@ -0,0 +1,124 @@
...
+        has_required_permissions = super().has_permission()
+        if not has_required_permissions:
+            return False
+        application = self.get_object().get_specific_model()
+        doc_pack = document_pack.pack_active_get(application)
+        if doc_pack and self.kwargs["cdr_pk"] in doc_pack.document_references.values_list(
+            "pk", flat=True
+        ):
+            return True
+        return False
```
MattHolmes123, line 117: this will only work for the active document pack, if we include the `issued_document_pk` this would work for any pack rather than the most recent one.

web/domains/case/urls.py
```diff
@@ -428,6 +429,16 @@
                                 name="ilb-case-history",
                                 kwargs={"mode": "ilb"},
                             ),
+                            path(
+                                "documents/",
+                                views_documents.ConstabularyDocumentView.as_view(),
+                                name="constabulary-doc",
+                            ),
```
MattHolmes123, line 436: As discussed in the view I would add an `issued_document_pk` / `pack_pk` to this url.

web/domains/case/urls.py
```diff
@@ -428,6 +429,16 @@
...
+                            path(
+                                "documents/<int:cdr_pk>/",
+                                views_documents.ConstabularyDocumentDownloadView.as_view(),
+                                name="constabulary-doc-download",
+                            ),
```
MattHolmes123, line 441: As discussed in the view I would add an `issued_document_pk` / `pack_pk` to this url.

MattHolmes123 (changes requested):

web/domains/case/views/views_documents.py
MattHolmes123: Actually they might always see the most recent if we always get the first issued pack but I think adding a pack_pk url kwarg would make it easier to work with.

web/domains/case/views/views_documents.py
```diff
@@ -0,0 +1,124 @@
...
+    def get_document_context_data(self, application: ImportApplication) -> dict[str, Any]:
+        at = application.application_type
+        doc_pack = self.get_document_pack(application)
+        licence_cdr = self.get_licence_cdr(doc_pack)
+        cover_letter_cdr = self.get_cover_letter_cdr(doc_pack)
+        return {
+            "issue_date": doc_pack.case_completion_datetime,
+            "is_import": True,
+            "document_reference": application.reference,
+            "type_label": at.Types(at.type).label,
```
MattHolmes123, line 92: Is this the same as
`at.get_type_display()`

MattHolmes123 (approved):

PR #1260, opened 2023-11-23
ICMSLST-1974 Add New User Welcome Email.

MattHolmes123:
Changes:
  - Add NewUserWelcomeEmail class.
  - Update ICMSGovUKOneLoginBackend to send new user welcome message.
  
  
  Example email:
  
image


pii-ner-exclude.txt
```diff
@@ -3513,3 +3513,8 @@ addopts=--reuse-db
 Document Set Title
 Authorise
 INFO
+PDB
+ps
+Ctrl + Q
```
mikeylockwood, line 3518: Where's this string come from?

MattHolmes123: I have this in my local docker-compose, 
image
I normally stash it but it picked up the change and I stopped caring :D

mikeylockwood (approved):

PR #1265, opened 2023-12-04
ICMSLST-1398 Add certificate checker

mikeylockwood:
- Certificate checker added to http://export-a-certificate/check
- Query args are `CERTIFICATE_REFERENCE` and `CERTIFICATE_CODE`, mimicking V1. Also COUNTRY and ORGANISATION have been added
- Access does not require log in
- Add check_code field to CaseDocumentReference model, generation mimics V1

image
image

Helpers have been added for manually finding the information for new fields

image


web/static/web/js/pages/certificate-checker.js
```diff
@@ -0,0 +1,17 @@
...
+        arrow.innerHTML = "▼") : (
+        panel.style.display = "none",
+        arrow.innerHTML = "►"
+    )
+}
+
+window.addEventListener('load', (event) => {
+    const whereToFindElements = document.querySelectorAll(".whereToFind");
+    whereToFindElements.forEach((el) => el.addEventListener("click", toggleHelper));
+});
```
mikeylockwood, line 17: Thoughts on whether this should be rewritten in stone age javascript

mikeylockwood:
Revoked certificates show a revoked message with no download link

image

web/domains/case/services/document_pack.py
```diff
@@ -390,7 +391,11 @@ def doc_ref_documents_all(doc_pack: DocumentPack) -> QuerySet[CaseDocumentRefere
 def _create_document(
     doc_pack: DocumentPack, doc_type: str, doc_reference: str | None = None
 ) -> CaseDocumentReference:
-    return doc_pack.document_references.create(document_type=doc_type, reference=doc_reference)
+    # TODO ICMSLST-2406 check_code generation is currently mimicing how V1 works
+    check_code = randint(10000000, 99999999)
+    return doc_pack.document_references.create(
+        document_type=doc_type, reference=doc_reference, check_code=str(check_code)
+    )
```
mikeylockwood, line 398: Creating the check_code value is currently the same as V1 works. However in V2, this is stored as a string so we have the flexibility of changing to a hash if desired

web/utils/s3.py
```diff
@@ -71,3 +72,27 @@ def put_object_in_s3(file_data: str, key: str, client: Optional["S3Client"] = No
...
+    # Generate a presigned URL for the S3 object
+    s3_client = get_s3_client()
+    try:
+        response = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": settings.AWS_STORAGE_BUCKET_NAME, "Key": key},
+            ExpiresIn=expiration,
+        )
+    except ClientError as e:
+        logging.error(e)
```
chris-pettinga, line 94: Do we not want to be alerted of an error and for it to fail gracefully? If this function returns None, then the resulting webpage will have a broken download link which will be frustrating

mikeylockwood: Good point, yes. Would probably want some messaging to the user that the generating the download link has failed and they should try again.

web/static/web/js/pages/certificate-checker.js
chris-pettinga: I like it currently, easy to understand and short

web/domains/checker/views.py
```diff
@@ -0,0 +1,109 @@
...
+                    )
+
+                context["document"] = document
+                context["country"] = country_name
+
+                app = cert.export_application
+                context["exporter"] = app.exporter.name
+
+                context["issue_date"] = cert.case_completion_datetime.strftime("%d %B %Y")
+                context["goods"] = _get_export_application_goods(app)
```
chris-pettinga, line 74: As this function is only called once, my preference would be to just have it straight in this block of code for easier readability.

chris-pettinga (changes requested):

web/domains/checker/views.py
mikeylockwood: Initially it was extracted to properly unit test which I forgot to add. I would prefer to leave the definition outside of the view and to add unit tests for it

web/domains/case/services/document_pack.py
MattHolmes123: Do we care that the check code isn't unique?

web/domains/checker/views.py
```diff
@@ -0,0 +1,109 @@
...
+def _get_export_application_goods(app: ExportApplication) -> str:
+    match app.process_type:
+        case ProcessTypes.GMP:
+            return "N/A"
+        case ProcessTypes.CFS:
+            cfs = app.certificateoffreesaleapplication
+            product_names = cfs.schedules.values_list("products__product_name", flat=True)
+            return ", ".join(goods for goods in product_names)
+        case ProcessTypes.COM:
+            return app.certificateofmanufactureapplication.product_name
```
MattHolmes123, line 26: same her

web/domains/case/services/document_pack.py
mikeylockwood: It currently isn't unique in v1 and there are some that are duplicated so I'll say not for backward compatibility but we could make it unique

web/domains/checker/views.py
```diff
@@ -0,0 +1,109 @@
...
+        if form.is_valid():
+            check = form.clean()
+            reference = check["certificate_reference"]
+            code = check["certificate_code"]
+            country_name = check["country"]
+            org_name = check["organisation_name"]
+
+            try:
+                document = CaseDocumentReference.objects.get(
+                    reference=reference, check_code=code, reference_data__country__name=country_name
```
MattHolmes123, line 52: should we use case insensitive filtering for country name?
```
>>> Country.objects.filter(name="germany")
<QuerySet []>
>>> Country.objects.filter(name__iexact="germany")
<QuerySet [<Country: Germany>]>
```

MattHolmes123: Would make the form easier to use.

MattHolmes123: Just noticed you use iexact further down so might as well here?

web/domains/checker/views.py
```diff
@@ -0,0 +1,109 @@
...
+                )
+
+                # GMP Certificates have the manufacturer name on the certificate rather than the exporter
+                if reference and reference.startswith("GMP"):
```
MattHolmes123, line 56: The form is valid in this block so can't this just be:
`if reference.startswith("GMP"):`

web/domains/checker/views.py
mikeylockwood: The country field on the form is a django select2 widget look up so it can only be the exact country name that is entered

web/templates/web/domains/checker/certificate-checker.html
```diff
@@ -0,0 +1,137 @@
...
+      <strong>Enter the certificate reference and code below to check if the certificate is valid.
+      The certificate reference can be found on the certificate in the format 'CFS/2000/00001',
+      'COM/2000/00001', or 'GMP/2000/00001'. The certificate code is found under the QR code on
+      the certificate and is 8 digits long. Note that this service can only be used to validate
+      certificates with a certificate code.</strong>
+  </p>
+  {% call forms.form(action='', method='post', csrf_input=csrf_input) -%}
+    {% block extra_main_content_form %}{% endblock %}
+    {% for field in form %}
+      {{ fields.field(field, show_errors=False) }}
```
MattHolmes123, line 31: Why don't you want the errors?

web/utils/s3.py
MattHolmes123: Use this to send the exception to sentry
`from web.utils.sentry import capture_exception`

MattHolmes123 (approved):
Fix the few things and its good to go

web/templates/web/domains/checker/certificate-checker.html
mikeylockwood: Originally I was just copying the messaging in V1, but with the extra fields now it is probably better with the errors so I put them back

PR #1268, opened 2024-01-04
ICMSLST-2418 Use allow list when validating uploaded files.

MattHolmes123:


MattHolmes123:
image
Example usage in ICMS

web/domains/file/utils.py
```diff
@@ -9,26 +9,35 @@
...
+    "pdf",
+    "png",
+    "rtf",
+    "tif",
+    "tiff",
+    "txt",
+    "xls",
+    "xlsb",
+    "xlsx",
+    "xps",
```
MattHolmes123, line 35: Grabbed this list from ICMS V1 database for now.

MattHolmes123: Ideally we settle on a smaller list of extensions:
https://cheatsheetseries.owasp.org/cheatsheets/File_Upload_Cheat_Sheet.html#list-allowed-extensions

PR #1271, opened 2024-01-09
ICMSLST-2419 Upgrade jquery and jquery-ui to latest version.

MattHolmes123:
Changes:
  - Upgrade jquery and jquery-ui
  - Refactor code out of main.js and include in views that use it.
  - Remove lots of javascript that was copied over from ICMS V1.


web/templates/web/domains/country/groups/edit.html
```diff
@@ -1,6 +1,11 @@
 {% extends "model/edit.html" %}
 {% from "web/domains/country/country-selections.html" import country_selector %}
 
+{% block page_js %}
+  {{ super() }}
```
chris-pettinga, line 5: I would implement a separate block, called something like `extra_js` which doesn't require a `{{ super() }}` call, in case future developers forget to call super and it causes issues. 

Then again it extra_js could be implemented in parent templates on the chain to the base so a super() would still be required..

chris-pettinga (approved):
Perhaps not best placed to review/approve but it looks good to me. As long as all JS that is actually being used has been placed in a separate file and that file has been imported on all pages it is required.

web/templates/web/domains/country/groups/edit.html
MattHolmes123: Super wasn't needed tbh in all of these places where I've moved code out of main.js.
I don't think any of the base templates actually added any page_js content.

PR #1273, opened 2024-01-10
ICMSLST-2431 Create dummy active signature for non prod environments

mikeylockwood:
- Adds a dummy Signature object for non prod environments
- Fixes RevokeObjectUrl call in the preview when changing a file in Add Signature

### DIFF
--- a/web/management/commands/add_dummy_data.py
+++ b/web/management/commands/add_dummy_data.py
@@ -1,5 +1,6 @@
 import datetime
 from collections.abc import Collection
+from pathlib import Path
 from typing import Any
 
 from django.conf import settings
@@ -18,9 +19,11 @@ from web.models import (
     ObsoleteCalibre,
     ObsoleteCalibreGroup,
     Office,
+    Signature,
     User,
 )
 from web.permissions import constabulary_add_contact, organisation_add_contact
+from web.utils.s3 import upload_file_obj_to_s3
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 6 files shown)


web/management/commands/add_dummy_data.py
```diff
@@ -398,3 +403,25 @@ def create_certificate_application_templates(
...
+
+
+def create_dummy_signature(user: User) -> None:
+    """Creates a dummy active signature object to appear in licence and certificate documents"""
+    file_path = Path(__file__).parent.parent.parent / "static/web/img/dit-no-signature.png"
+    filename = "active_dummy_signature.png"
+    key = f"dummy_signature/{filename}"
+
+    with open(file_path, "rb") as f:
+        file_size = upload_file_obj_to_s3(f, key)
```
MattHolmes123, line 415:
```
file = Path(__file__).parent.parent.parent / "static/web/img/dit-no-signature.png"

if file.is_file():
    file.read_bytes()

You can use the Path object to read the bytes or string of the file

```

web/management/commands/add_dummy_data.py
```diff
@@ -398,3 +403,25 @@ def create_certificate_application_templates(
         )
 
     return CertificateApplicationTemplate.objects.bulk_create(objs)
+
+
+def create_dummy_signature(user: User) -> None:
+    """Creates a dummy active signature object to appear in licence and certificate documents"""
+    file_path = Path(__file__).parent.parent.parent / "static/web/img/dit-no-signature.png"
```
chris-pettinga, line 410: I would use `settings.ROOT_DIR` here in case the position of this file changes for whatever reason.

MattHolmes123 (approved):

web/management/commands/add_dummy_data.py
MattHolmes123: `settings.BASE_DIR` ?

marcuspp (approved):

PR #1279, opened 2024-01-10
ICMSLST-2422 - CSP headers

chris-pettinga:
Added CSP headers via django-csp-middleware

Added nonces to all script elements

Added x-permitted-cross-domain-policy header

Added tests

### DIFF
--- a/config/env.py
+++ b/config/env.py
@@ -160,5 +160,9 @@ class Environment(BaseSettings):
     # local site URL management
     local_site_url: str = "http://web:8080/"
 
+    # CSP settings
+    csp_report_only: bool = True
+    csp_report_uri: list[str] | None = None
+
 
 env: Environment = Environment()  # type:ignore[call-arg]
--- a/config/settings.py
+++ b/config/settings.py
@@ -69,6 +69,8 @@ MIDDLEWARE = [
     "htmlmin.middleware.MarkRequestMiddleware",
     "web.middleware.common.ICMSMiddleware",
     "web.middleware.one_login.UserFullyRegisteredMiddleware",
+    "csp.middleware.CSPMiddleware",
+    "web.middleware.common.SetPermittedCrossDomainPolicyHeaderMiddleware",
 ]
 
 ROOT_URLCONF = "config.urls"
@@ -163,7 +165,6 @@ LOGIN_URL = "login-start"
 LOGIN_REDIRECT_URL = "workbasket"
 # Do not use this (all handled in "logout-user")
 LOGOUT_REDIRECT_URL = None
-
 
 if STAFF_SSO_ENABLED:
     AUTHENTICATION_BACKENDS.append("web.auth.backends.ICMSStaffSSOBackend")
@@ -273,7 +274,6 @@ CELERY_ACCEPT_CONTENT = ["application/json"]
 CELERY_RESULT_SERIALIZER = "json"
 CELERY_TASK_SERIALIZER = "json"
 CELERY_RESULT_EXTENDED = True
-
 
 # Django cache with Redis
 CACHES = {
@@ -420,7 +420,6 @@ LOGGING = {
 if env.sentry_enabled:
     init_sentry(env.sentry_dsn, env.sentry_environment)
 
-
 # Settings for production environment
 if APP_ENV == "production":
     # TODO compression causes 50 error on server
@@ -459,3 +458,35 @@ else:
 
 # Local site URL management
 LOCAL_SITE_URL = env.local_site_url
+
+# CSP Settings
+
+# CSP policies
+CSP_DEFAULT_SRC = ("'self'",)
+
+# JS tags with a src attribute can only be loaded from ICMS itself or the DBT Sentry instance
+CSP_SCRIPT_SRC = (
+    "'self'",
+    "https://sentry.ci.uktrade.digital/",
+)
+# CSS elements with a src attribute can only be loaded from ICMS itself or inline, e.g. <style> tags
+CSP_STYLE_SRC = (
+    "'self'",
+    "'unsafe-inline'",
+)
+# Fonts can only be loaded from ICMS itself or data URIs
+CSP_FONT_SRC = (
+    "'self'",
+    "data:",
+)
+
+# CSP meta-settings
+
+# inline scripts without a src attribute must have a nonce attribute
+CSP_INCLUDE_NONCE_IN = ["script-src"]
+
+# if True, CSP violations are reported but not enforced
+CSP_REPORT_ONLY = env.csp_report_only
+
+# URL to send CSP violation reports to
+CSP_REPORT_URI = env.csp_report_uri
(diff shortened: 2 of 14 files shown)


web/templates/partial/js_constants.html
```diff
@@ -1,4 +1,4 @@
-<script type="application/javascript">
+<script nonce="{{request.csp_nonce}}" type="application/javascript">
```
mikeylockwood, line 1: Personally prefer template variables to have a space between the brackets e.g. `{{ request.csp_nonce }}` but that's  being very petty. Fine also to leave it

mikeylockwood (approved):

config/settings.py
```diff
@@ -459,3 +457,22 @@
...
+)
+CSP_FONT_SRC = (
+    "'self'",
+    "data:",
+)
+
+# CSP meta-settings
+CSP_INCLUDE_NONCE_IN = ["script-src"]
+CSP_REPORT_ONLY = True
+CSP_REPORT_URI = env.csp_report_uri
```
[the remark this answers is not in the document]
chris-pettinga: They are picked up by the django-csp-middleware package which then uses them to craft the correct response header

web/templates/partial/js_constants.html
chris-pettinga: Agreed, will change now

config/settings.py
```diff
@@ -459,3 +457,24 @@
...
+    "'self'",
+    "data:",
+)  # Fonts can only be loaded from ICMS itself or data URIs
+
+# CSP meta-settings
+CSP_INCLUDE_NONCE_IN = [
+    "script-src"
+]  # inline scripts without a src attribute must have a nonce attribute
+CSP_REPORT_ONLY = env.csp_report_only  # if True, CSP violations are reported but not enforced
+CSP_REPORT_URI = env.csp_report_uri  # URL to send CSP violation reports to
```
MattHolmes123, line 480: Could you write the documentation without using inline comments please.

They are harder to read and are not consistent with the rest of the comments in the settings file

web/templates/base.html
```diff
@@ -36,7 +36,7 @@
             <li><a href="#main">Main content</a></li>
         </ul>
     </div>
-    <iframe id="iframe-wrapper" src="javascript:'';" style="border: none; overflow: hidden; visibility: hidden; display: none;">&nbsp;</iframe>
+    <iframe id="iframe-wrapper" src="about:blank" style="border: none; overflow: hidden; visibility: hidden; display: none;">&nbsp;</iframe>
```
MattHolmes123, line 39: I've just checked the src in V1 test and it looks like someone copied this over:
image
Have a look in the v2 codebase but I think we can get rid of this

web/templates/web/domains/chief/_base.html
```diff
@@ -10,7 +10,7 @@
 {% endblock main_content %}
 
 {% block page_js %}
-  <script type="text/javascript">
+  <script nonce="{{ request.csp_nonce }}" type="text/javascript">
```
MattHolmes123, line 13: Nothing to do with this line but I've checked through the codebase and you've missed this inline script:
`web/templates/web/search-companies.html`
Probably because it doesn't specify the type:
`<script>`

config/env.py
```diff
@@ -160,5 +160,9 @@ class Environment(BaseSettings):
     # local site URL management
     local_site_url: str = "http://web:8080/"
 
+    # CSP settings
+    csp_report_only: bool = True
+    csp_report_uri: str = ""
```
MattHolmes123, line 165: How come we aren't adding an endpoint to report csp breaches to?
I added this view locally to test that it would send requests to an endpoint and it worked.
Not suggesting it's finished.

```
from web.utils.sentry import capture_message

@require_POST
@csrf_exempt
def csp_violation_report(request):
    if request.content_type == "application/csp-report":
        # TODO: Do something with the report
        ...
        capture_message("We've got a naughty script")
```

MattHolmes123: I can create a PR to your PR if you want to see what I did.

MattHolmes123 (changes requested):

config/env.py
chris-pettinga: Sentry provides a CSP reporting URI out-of-the-box which automatically deals with parsing the report. My concern was introducing noise from local environment CSP violations so I wanted to only include the REPORT_URI on deployed environments. What I'll do is hardcode the URI (as it's the same for all environments) and override it on local to be None, what do you think?

MattHolmes123: We shouldn't get any noise because isn't the goal to fix violations?
Thats cool about sentry for the deployed environments though.

MattHolmes123: Leave it as you have it now (as an environment variable)

MattHolmes123: Actually the type is wrong though.
CSP_REPORT_URI
    Set the report-uri directive. A tuple or list. Each URI can be a full or relative URI. None Note: This doesn’t use default-src as a fall-back.
    
    `csp_report_uri: list[str] = ""`

web/templates/base.html
chris-pettinga: @MattHolmes123 - It's used in js.fox.core-footer.js:473. Seemingly as part of loading spinner functionality blocking double-clicks.

MattHolmes123: core-footer.js is only 307 lines long?

config/env.py
chris-pettinga: The goal is to get alerted of any violations via Sentry, and if someone is developing some JS quickly and locally, they may not bother with a nonce which will result in a (I think) noisy Sentry alert sent to all team members. I think it's cleaner to only bother with the alerts on deployed instances so we know that code is actually a bit more developed and perhaps ready to go into main

chris-pettinga: Also just learnt that you can in fact include environment in the Sentry reporting URI, so we can perhaps solve this problem that way

MattHolmes123: I agree don't bother reporting to sentry locally, the example I added was from yesterday when I didn't even know sentry supported CSP.
The local endpoint would have just been used as a tool locally to tell the developer "hey you've added something invalid" so lets forget about it.

web/templates/web/search-companies.html
```diff
@@ -1,4 +1,4 @@
-<script>
+<script nonce="{{ request.csp_nonce }}">
```
MattHolmes123, line 1: Sorry there are actually several empty script tags like this:
`<script>`
Can you change them all to 
`<script nonce="{{ request.csp_nonce }}" type="text/javascript">`

MattHolmes123: I find 5 on main:
image

web/templates/base.html
chris-pettinga: static/web/js/fox/core-footer.js? Maybe it was reduced when you removed old JS recently

marcuspp (approved):

web/templates/base.html
MattHolmes123: Yes @chris-pettinga it's one of the files I updated.

MattHolmes123 (approved):

PR #1288, opened 2024-01-15
ICMSLST-696 Adding QR code to certificate documents

mikeylockwood:
- Add QR Code to certificate documents.
- QR code only appears when not viewing in preview mode

### DIFF
--- a/web/templates/pdf/export/certificate-base.html
+++ b/web/templates/pdf/export/certificate-base.html
@@ -42,21 +42,29 @@
 
       <div class="title-footer">
         <div class="signature-block page-margin">
-          {# TODO ICMSLST-697 - Revist when implementing digital signatures #}
+          {{ document_signature(signature_file, signature.content_type, signature.signatory) }}
           {% if preview %}
-            {{ document_signature(signature_file, signature.content_type, signature.signatory) }}
-            <span class="qr-code">[[QR Code Placeholder]]</span>
+            <div class="qr-code">
+              <span class="qr-placeholder">[[QR_PLACEHOLDER]]</span>
+            </div>
+            <div class="qr-text-block">
+              <span class="qr-text">QR code will be genrated when documents are authorised.</span>
+            </div>
           {% else %}
-            {{ document_signature(signature_file, signature.content_type, signature.signatory) }}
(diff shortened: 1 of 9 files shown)


web/templates/pdf/export/certificate-base.html
```diff
@@ -42,21 +42,28 @@ <h4 class="country centre">{{ country }}</h4>
...
+              <div class="qr-text-block">
+              <p class="qr-text">
+                Scan this QR code to check this certificate's validity.
+                <strong>Certificate Code: {{ certificate_code }}</strong>
+              </p>
+              <p class="qr-text">
+                Alternatively you can go to <strong>{{ qr_check_url }}</strong>
+                and enter the certificate reference and code.
+              </p>
+            {% endif %}
```
MattHolmes123, line 65: Why do you have two if preview / else / endif blocks.
Could it be simplified to one?

web/utils/pdf/utils.py
```diff
@@ -320,33 +326,60 @@ def _get_importer_eori_numbers(application: "FaImportApplication") -> list[str]:
     return eori_numbers
 
 
+def _certificate_document_context(
+    application: "ExportApplication",
+    certificate: "ExportApplicationCertificate",
+    doc_type: DocumentTypes,
+    country: "Country",
+) -> "Context":
+    qr_check_url = get_exporter_site_domain() + reverse("checker:certificate-checker")
```
MattHolmes123, line 335: from urllib.parse import urljoin

MattHolmes123 (approved):

web/utils/pdf/utils.py
mikeylockwood: `urljoin` doesn't add query strings

```
>>> urljoin(url, urlencode(query_params))
'http://export-a-certificate:8080/check/CERTIFICATE_REFERENCE=ABCD&CERTIFICATE_CODE=1234'

>>> urljoin(url, "?", urlencode(query_params))
'http://export-a-certificate:8080/check/'
```

MattHolmes123: You aren't adding query params here, you do that later on.
But you can still join the base domain and endpoint together using urljoin

PR #1289, opened 2024-01-16
ICMSLST-2422 - CSP fixes

chris-pettinga:
ICMSLST-2422

Adding some missing CSP values, such as cloudflare and IMAGE-SRC


config/settings.py
```diff
@@ -468,7 +468,12 @@
 CSP_SCRIPT_SRC = (
     "'self'",
     "https://sentry.ci.uktrade.digital/",
+    "https://cdnjs.cloudflare.com",
 )
+
+# JS scripts can can import other scripts, following the same rules as above
```
MattHolmes123, line 474: can can

PR #1290, opened 2024-01-16
ICMSLST-2381 & ICMSLST-2413 Data migration for check_code and replace &apos;

mikeylockwood:
- Data migration for the check_code field for export certificates
- Replace `&apos;` with ' character in templates and template paragraphs


web/migrations/0002_data_add_sites.py
```diff
@@ -12,11 +12,18 @@ def add_icms_sites(apps, schema_editor):
     Site.objects.create(domain="import-a-licence-unset", name=SiteName.IMPORTER)
 
 
+def reverse_icms_sites(apps, schema_editor):
+    Site = apps.get_model("sites", "Site")
+    Site.objects.filter(
+        name__in=(SiteName.CASEWORKER, SiteName.EXPORTER, SiteName.IMPORTER)
+    ).delete()
```
MattHolmes123, line 19: Might as well delete them all

data_migration/models/template.py
```diff
@@ -50,6 +59,13 @@ class CFSScheduleParagraph(MigrationBase):
     name = models.CharField(max_length=100)
     content = models.TextField(null=True)
 
+    @classmethod
+    def data_export(cls, data: dict[str, Any]) -> dict[str, Any]:
+        content = data["content"]
+        data["content"] = content and replace_apos(content)
```
MattHolmes123, line 65: Not saying it's better but this would work:
```
if (c := data["content"]):
    data["content"] = replace_apos(c)
```

MattHolmes123: Just found those two lines a bit odd to look at

MattHolmes123 (approved):

PR #1294, opened 2024-01-19
Add support for running ICMS on DBT Platform. 

MattHolmes123:
Environment variables are loaded via CloudFoundryEnvironment or DBTPlatformEnvironment.
settings.py remains agnostic to the environment in which it is running.
When building the container image the fake environment variables in .env.circleci are used to ensure there are no validation errors.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -14,7 +14,6 @@ import os
 import ssl
 from pathlib import Path
 
-import dj_database_url
 import jinja2
 import structlog
 from django.forms import Field
@@ -30,7 +29,7 @@ DEBUG = env.icms_debug
 WSGI_APPLICATION = "config.wsgi.application"
 APP_ENV = env.app_env
 SECRET_KEY = env.icms_secret_key
-ALLOWED_HOSTS = env.icms_allowed_hosts
+ALLOWED_HOSTS = env.allowed_hosts
 
 INSTALLED_APPS = [
     "web",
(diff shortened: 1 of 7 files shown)


config/settings.py
```diff
@@ -200,11 +200,7 @@
 # File storage
 # for https://github.com/uktrade/django-chunk-s3-av-upload-handlers
 
-if env.vcap_services:
-    app_bucket_creds = env.vcap_services.aws_s3_bucket[0]["credentials"]
-else:
-    app_bucket_creds = {}
-
+app_bucket_creds = env.get_s3_bucket_config()
```
chris-pettinga, line 203: I know it's not open for review yet but I wonder if we should use @computed_value for the env.py file so we can access them as properties much like the other env vars

chris-pettinga (approved):
Need to get more comfortable with DBT PaaS but looks good to me

acodeninja:
Looking good, I might point some other teams to this as a good example of config management for their migrations. We will have a few other bits for you to get in, but waiting on a review of the migration docs before that is publicly available.

MattHolmes123:
> Looking good, I might point some other teams to this as a good example of config management for their migrations. We will have a few other bits for you to get in, but waiting on a review of the migration docs before that is publicly available.

Cool sounds good @acodeninja I will leave this unmerged until it's finished then.

MattHolmes123:
Can confirm it loads the correct environment for cloud foundry when deployed.
image

requirements-base.txt
```diff
@@ -1,6 +1,7 @@
 Authlib==1.2.1
 boto3==1.29.6
 celery[redis]==5.3.6
+dbt-copilot-python==0.1.3
```
MattHolmes123, line 4: Will upgrade this when I do the celery healthcheck change

PR #1295, opened 2024-01-22
ICMSLST-2233 Exclude specific users from data migration

mikeylockwood:
- Exclude users with username not an email address from the data migration
- Exclude users with specific email domain from the data migration
- Add `icms_v1_user` flag to migrated users


config/env.py
```diff
@@ -123,6 +123,7 @@ class Environment(BaseSettings):
     icms_v1_replica_dsn: str = ""
     icms_prod_user: str = ""
     icms_prod_password: str = ""
+    data_migration_email_domain_exclude: str = ""
```
MattHolmes123, line 126: Do you only ever plan to support ignoring a single domain or could the list grow in future?

data_migration/models/user/access.py
```diff
@@ -56,6 +59,19 @@ def data_export(cls, data: dict[str, Any]) -> dict[str, Any]:
...
         return data
 
+    @classmethod
+    def get_source_data(cls) -> Generator:
+        values = cls.get_values()
+        values_kwargs = cls.get_values_kwargs()
+        related = cls.get_related()
+        return (
+            cls.objects.select_related(*related)
+            .exclude(submitted_by__username__endswith=EXCLUDE_DOMAIN)
```
MattHolmes123, line 69: `iendswith` for case insensitive search (same for all places you've used it)

data_migration/queries/permissions.py
```diff
@@ -57,6 +62,8 @@
   END login_id
   FROM securemgr.web_user_account_histories uah
   WHERE uah.person_id_current IS NOT NULL
+    AND login_id LIKE '%@%'
+    AND login_id NOT LIKE '%{EXCLUDE_DOMAIN}'
```
MattHolmes123, line 66: `ILIKE`

scripts/reset-data-migrations
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+rm data_migration/migrations/*.py
+make data_migrations
+
+# Restore the data migrations
+git restore data_migration/migrations/0002_data_firearms_act.py
+
+make black_format
+make isort_format
```
MattHolmes123, line 12: Missing newline

MattHolmes123 (approved):

MattHolmes123:
Have you done the other aspects of ICSMLST-2234 already?
image

config/env.py
mikeylockwood: Just a single domain it looks like. If needed in future I can change it to a list

PR #1297, opened 2024-01-23
ICMSLST-2009 - Issued Certificate Report

marcuspp:
Minimal approach so not included in this PR are as follows:
 - Delete option
 - Re-run option
 - Object permissions (just view permissions)
 - Scheduling
 - Continent column (will be added in a separate ticket)
 - Form options of Monthly / Quarterly / Yearly (as these ranges can be set using the manual option)
 - Email notifications
 - Error handling / Retry failed task (will be added in separate ticket)

To test
- Reset DB
- Add some export applications
- Login using the `report_user`
- And generate a report


web/reports/interfaces.py
```diff
@@ -0,0 +1,176 @@
...
+            else ", ".join(cdr.manufacturer_countries),
+            hse_email_count=export_application.case_emails.count(),
+            beis_email_count=export_application.case_emails.count(),
+            application_update_count=export_application.update_requests.count(),
+            fir_count=export_application.further_information_requests.count(),
+            product_legislation=""
+            if not hasattr(export_application, "schedules")
+            else ", ".join(cdr.legislations),
+            case_processing_time=self.get_total_processing_time(cdr, export_application),
+            total_processing_time=self.get_total_processing_time(cdr, export_application),
```
marcuspp, line 134: These values are the same as there was no real understanding of what case processing time is, so therefore ignored. The value seems to be total minus any time taken for an application update. Column could be removed but left in so output is the same as V1 (plus business days)

web/reports/views.py
```diff
@@ -0,0 +1,112 @@
...
+from web.models import GeneratedReport, Report, ScheduleReport
+from web.permissions import Perms
+from web.utils.s3 import get_file_from_s3
+
+from .constants import ReportStatus
+from .forms import IssuedCertificatesForm
+from .tasks import generate_issued_certificate_report_task
+
+
+class ReportViewMixin(LoginRequiredMixin, PermissionRequiredMixin):
```
chris-pettinga, line 19: In my mind a Mixin shouldn't inherit from any classes, when it does so it becomes a Base/Abstract class.

web/reports/views.py
```diff
@@ -0,0 +1,112 @@
...
+        return get_object_or_404(Report, pk=self.kwargs["report_pk"])
+
+
+class ReportListView(ReportViewMixin, ListView):
+    template_name = "web/domains/reports/list-view.html"
+    model = Report
+
+    def get_context_data(self, *args, **kwargs) -> dict[str, Any]:
+        context = super().get_context_data(*args, **kwargs)
+        context["page_title"] = "Reports"
```
chris-pettinga, line 32: can this not be done in the .html template itself? Same for the page_title elements below

web/reports/views.py
```diff
@@ -0,0 +1,112 @@
...
+        scheduled_report.save()
+        generate_issued_certificate_report_task.delay(scheduled_report.pk)
+        return super().form_valid(form)
+
+    def get_success_url(self) -> str:
+        return reverse("run-history-view", kwargs={"report_pk": self.kwargs["report_pk"]})
+
+    def get_report_parameters(self, form) -> dict:
+        data = {}
+        for field, value in form.cleaned_data.items():
```
chris-pettinga, line 91: I would pop `title` and `notes` manually from the cleaned_data dict so it's more explicit what we don't want to return, rather than loop through and check every time

web/reports/views.py
chris-pettinga: this only makes sense if I'm correct in thinking that the `page_title` context value maps to the <title> tag on the page. Reason being that anything that is explicitly to do with the appearance of a page should rather be done in the page itself rather than in Python

web/reports/views.py
```diff
@@ -0,0 +1,112 @@
...
+
+from .constants import ReportStatus
+from .forms import IssuedCertificatesForm
+from .tasks import generate_issued_certificate_report_task
+
+
+class ReportViewMixin(LoginRequiredMixin, PermissionRequiredMixin):
+    permission_required = [Perms.page.view_reports]
+
+    def get_report(self) -> Report:
```
chris-pettinga, line 22: It looks like all views use this, so I would override the __dispatch__ method are assign it the view `self` as an attribute so it doesn't have to be run twice per view, i.e. once in `get_context_data` and again in `form_valid`

web/reports/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Report(models.Model):
+    name = models.CharField(max_length=400)
+    description = models.TextField()
+    report_type = models.CharField(max_length=120, choices=ReportType.choices, unique=True)
+
+    def last_run_completed_at(self) -> str:
+        last_run = self.schedules.order_by("-finished_at").first()
+        if last_run and last_run.finished_at:
+            return last_run.finished_at.strftime("%d %b %Y %H:%M:%S")
```
chris-pettinga, line 16: I think Python functions should always try and return flexible python objects, in this case a date object. If the only reason this is returning a string is so it can be rendered in HTML cleanly, I would move the .strftime to the .HTML file itself. 

This allows you to open up this function to be ran by other python functions which need to operate on a datetime object, rather than converting it back from a string

web/reports/generate.py
```diff
@@ -0,0 +1,78 @@
...
+    return f"{scheduled_report.report.name}-{scheduled_report.title}-{scheduled_report.pk}"
+
+
+def generate_issued_certificate_report(scheduled_report: ScheduleReport) -> None:
+    report_interface = IssuedCertificateReportInterface(scheduled_report)
+    scheduled_report.status = ReportStatus.IN_PROGRESS
+    scheduled_report.started_at = timezone.now()
+    scheduled_report.save()
+
+    write_files(report_interface)
```
chris-pettinga, line 24: I would move `write_files` to a method on the `ReportInterface` base_class as it seems like something that the ReportInterface object does to itself, rather than pass it to another function. Not a blocker however just personal preference

chris-pettinga: Then also you could move all the other sub_functions as methods on the BaseClass, could also help with overriding down the line if we want to implement different saving behaviour for different types of reports

chris-pettinga (changes requested):
Some comments but overall looks great! Hell of a lot of code so well done!

web/reports/views.py
marcuspp: I'm not sure this matters as `get_context_data` will only be called on a GET request or an invalid POST request whereas the call within `form_valid` is called on a valid POST request. So in both cases it should only be called once.

requirements-base.txt
```diff
@@ -24,6 +24,7 @@ Jinja2==3.1.3
 lxml==4.9.3
 mohawk==1.1.0
 notifications-python-client==8.1.0
+numpy==1.26.3
```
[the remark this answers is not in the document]
marcuspp: Happy to switch packages. Will give it a try and see if it is a drop in replacement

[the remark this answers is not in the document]
MattHolmes123: Do we need a package at all:

```
import datetime as dt


def main():
    start = dt.date(year=2020, month=1, day=1)
    end = dt.date(year=2020, month=3, day=31)

    date_diff = end - start
    business_days = 0

    for i in range(date_diff.days + 1):
        day = start + dt.timedelta(days=i)

        # "Return day of the week, where Monday == 0 ... Sunday == 6."
        if day.weekday() < 5:
            business_days += 1

    print(business_days)


if __name__ == "__main__":
    main()
```

web/reports/interfaces.py
```diff
@@ -0,0 +1,176 @@
...
+        hours = time_delta.seconds // 3600
+        minutes = (time_delta.seconds % 3600) // 60
+        return f"{time_delta.days}d {hours}h {minutes}m"
+
+    def get_business_days_to_process(
+        self, cdr: CaseDocumentReference, export_application: ExportApplication
+    ) -> int:
+        completed_date = cdr.content_object.case_completion_datetime.date()
+        submitted_date = export_application.submit_datetime.date()
+        return numpy.busday_count(submitted_date, completed_date) + 1
```
MattHolmes123, line 176:
```import datetime as dt

    def get_business_days_to_process(
        self, cdr: CaseDocumentReference, export_application: ExportApplication
    ) -> int:

        start = export_application.submit_datetime.date()
        end = cdr.content_object.case_completion_datetime.date()
        date_diff = end - start
        business_days = 0

        for i in range(date_diff.days + 1):
            day = start + dt.timedelta(days=i)

            # "Return day of the week, where Monday == 0 ... Sunday == 6."
            if day.weekday() < 5:
                business_days += 1

        return business_days

```

web/reports/serializers.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import Annotated
+
+import pydantic
+
+CustomDateTime = Annotated[
+    datetime,
+    pydantic.PlainSerializer(
+        lambda _datetime: _datetime.strftime("%d/%m/%Y %H:%M:%S"), return_type=str
+    ),
+]
```
MattHolmes123, line 11: Do you want to link to the jquery form field to indicate why we have to do this.

web/reports/views.py
MattHolmes123: ReportViewBase 👍

MattHolmes123 (changes requested):

web/reports/generate.py
marcuspp: I'm going to leave this as is for now and when i do the next report consider moving them.

web/reports/interfaces.py
```diff
@@ -0,0 +1,176 @@
...
+            responsible_person_statement=self.get_is_responsible_person(export_application),
+            countries_of_manufacture=""
+            if not hasattr(export_application, "schedules")
+            else ", ".join(cdr.manufacturer_countries),
+            hse_email_count=export_application.case_emails.count(),
+            beis_email_count=export_application.case_emails.count(),
+            application_update_count=export_application.update_requests.count(),
+            fir_count=export_application.further_information_requests.count(),
+            product_legislation=""
+            if not hasattr(export_application, "schedules")
```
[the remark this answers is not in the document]
marcuspp: Yeah good shout, i'll update

web/reports/interfaces.py
```diff
@@ -0,0 +1,176 @@
...
+        ):
+            return "Yes"
+        return "No"
+
+    def get_total_processing_time(
+        self, cdr: CaseDocumentReference, export_application: ExportApplication
+    ) -> str:
+        return self._format_time_delta(
+            cdr.content_object.case_completion_datetime, export_application.submit_datetime
+        )
```
MattHolmes123, line 163: This breaks as soon as an application has a variation request.
image
That exampe there shows me running the report twice.
Before the second run I did the following:

1. Create a variation request as an admin worker
2. Created an application update request
3. As the exporter user added two more countries to the app and completed the app update request.
4. Completed the case again as the caseworker
5. reran the report

web/reports/views.py
mikeylockwood: This is passed in as a context variable as it is set in the base template that the template inherits from

web/reports/interfaces.py
MattHolmes123: image

web/reports/views.py
chris-pettinga: @marcuspp this is true, my mistake.

web/reports/views.py
chris-pettinga: @mikeylockwood makes sense, thank you

chris-pettinga (approved):

web/reports/generate.py
```diff
@@ -0,0 +1,84 @@
...
+    config.header.styles = {"bold": True}
+    config.rows = [d.values() for d in data]
+    config.column_width = 25
+    config.sheet_name = "Sheet 1"
+    file_data = generate_xlsx_file(config, {"remove_timezone": True})
+    write_file_data(
+        scheduled_report,
+        file_data,
+        file_name,
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
```
mikeylockwood, line 53: Maybe mimetypes should be constants in `web/utils/spreadsheet.py`

```
MIMETYPE.XLSX
MIMETYPE.CSV
```

web/reports/serializers.py
mikeylockwood: I think it's clear that it's serilaising datetime objects to strings. What else needs to be known?

MattHolmes123: I believe he created the CustomDateTime because the format is different to how pydantic serialises datetime by default.
`icms/web/forms/fields.py -> JqueryDateField`

```
# Always match the format here: web/static/web/js/fox/core-footer.js
# dateFormat: "dd'-'M'-'yy"
JQUERY_DATE_FORMAT = "%d-%b-%Y"


class JqueryDateField(forms.DateField):
    widget = DateInput(format=JQUERY_DATE_FORMAT)
    input_formats = [JQUERY_DATE_FORMAT]
```

MattHolmes123: It's nothing to do with our JQuery form field :D 
My bad

web/reports/interfaces.py
MattHolmes123: Will be fixed in https://uktrade.atlassian.net/browse/ICMSLST-2474

MattHolmes123 (approved):

web/reports/generate.py
marcuspp: Yes good shout, i'll do this and then rebase

mikeylockwood (approved):

PR #1304, opened 2024-02-01
ICMSLST-2383 Migrate updated case note fields

mikeylockwood:
- Use `created_by_id` and `created_datetime` to populate `updated_by_id` and `updated_at` as those fields do not appear in V1
- Change `TIMESTAMP_UPDATES` list to be a list of queries to be run rather than a list of models containing queries as some models will require two queries.
- Add post migration checks for case notes


scripts/reset-data-migrations
marcuspp, line 1: Is this meant to be empty?

data_migration/queries/case.py
```diff
@@ -81,13 +81,21 @@
...
-case_note_timestamp_update = """
+case_note_created_timestamp_update = """
 UPDATE web_casenote SET create_datetime = data_migration_casenote.create_datetime
 FROM data_migration_casenote
 WHERE web_casenote.id = data_migration_casenote.id
 """
 
 
+# Case note has no updated timestamp in V1 so use created
+case_note_updated_timestamp_update = """
```
marcuspp, line 92: Can this not be done within `case_note_created_timestamp_update` ?

mikeylockwood: Yes it can, good point

scripts/reset-data-migrations
mikeylockwood: Nope, I think this has been caused by switches between branches

marcuspp (approved):

MattHolmes123 (approved):

scripts/reset-data-migrations
mikeylockwood: Oh no this was me just making it executable

PR #1305, opened 2024-02-01
ICMSLST-1428

chris-pettinga:
ICMSLST-1428 

Adding tests for the get_licence_endorsements function, no need to change it that much as it works as expected


web/tests/utils/pdf/test_utils.py
```diff
@@ -3,19 +3,26 @@
 
 import pytest
 
+from web.domains.case._import.models import EndorsementImportApplication
```
MattHolmes123, line 6: `from web.models import EndorsementImportApplication`

web/tests/utils/pdf/test_utils.py
```diff
@@ -3,19 +3,26 @@
...
 from web.models import SILGoodsSection582Obsolete  # /PS-IGNORE
 from web.models import SILGoodsSection582Other  # /PS-IGNORE
 from web.models import SILGoodsSection1, SILGoodsSection2, SILGoodsSection5, Template
 from web.types import DocumentTypes
 from web.utils.pdf import utils
+from web.utils.pdf.utils import get_licence_endorsements
```
MattHolmes123, line 12: `get_licence_endorsements` is already used in this file, please do this: `utils.get_licence_endorsements`

web/utils/pdf/utils.py
```diff
@@ -55,7 +55,15 @@
...
-    Context = dict[str, str | bool | Collection[str] | ImpOrExp | dict[int, ScheduleText]]
+    Context = dict[
+        str,
+        str
+        | bool
+        | Collection[str]
+        | Collection[Collection[str]]
+        | ImpOrExp
+        | dict[int, ScheduleText],
+    ]
```
MattHolmes123, line 66: At what point do we just change the Context type to just `dict[str, Any` 😄.
I believe all the places it's used we just merge it with other template context and then render a template.

MattHolmes123 (approved):

PR #1312, opened 2024-02-06
ICMSLST-1832 Data Migration - WithdrawApplication

mikeylockwood:
- Data migration for WithdrawApplication model
- Data fetched from xml in import and export applications queries
- DRAFT status withdraw applications excluded from V2
- Data checks added to ensure counts of withdraw applications match in V1 and V2


data_migration/tests/test_e2e_export.py
```diff
@@ -166,10 +168,12 @@ def test_import_export_data(mock_connect, dummy_dm_settings):
...
-    assert web.CertificateOfGoodManufacturingPracticeApplication.objects.count() == 3
-    ea1, ea2, ea3 = web.ExportApplication.objects.filter(
-        process_ptr__process_type=ProcessTypes.GMP
-    ).order_by("pk")
+    assert web.CertificateOfGoodManufacturingPracticeApplication.objects.count() == 4
+    ea1, ea2, ea3 = (
+        web.ExportApplication.objects.filter(process_ptr__process_type=ProcessTypes.GMP)
+        .exclude(status="WITHDRAWN")
+        .order_by("pk")
+    )
```
MattHolmes123, line 176: Does it only match three because the 4th is withdrawn?

MattHolmes123 (approved):

data_migration/tests/test_e2e_export.py
mikeylockwood: Yeah I added an extra withdrawn application to the tests but didn't want to include it in the assertions. I guess I should just use `_` instead

PR #1313, opened 2024-02-06
ICMSLST-2330 - Fix Sanctions PDF page breaks

chris-pettinga:
ICMSLST-2330

Fixing page breaks in sanctions licences. Now endorsements that are too long are manually split between multiple pages.

### DIFF
--- a/web/static/web/css/pdfs/sanctions-licence-style.css
+++ b/web/static/web/css/pdfs/sanctions-licence-style.css
@@ -55,9 +55,11 @@ p {
     left: -28px;
     text-align: center;
 }
+
 .holder-number-first {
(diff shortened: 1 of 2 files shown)


MattHolmes123:
Might be worth creating a management command we can run that creates the required data for each live licence type that populates each pdf with a large amount of data.
Would allow us to iterate quickly if it outputs PDFs to a local directory.

chris-pettinga:
@MattHolmes123 good idea, I've created https://uktrade.atlassian.net/browse/ICMSLST-2533 to deal with this

MattHolmes123 (approved):

PR #1327, opened 2024-02-12
ICMSLST-2010 - Import Licence Report

marcuspp:
- Removed setting `start_datetime` to now() from `CommodityGroup` and `Commodity` both auto add fields so django ignores any attempt to set them
 - Sets start date for commodity `Usage` to 2023-01-01 rather than time of Data Migration
 -  Added Import Licence Report
 - Test data improvements to simulate time deltas between submission and approval of applications


web/reports/interfaces.py
```diff
@@ -92,8 +114,10 @@ def get_queryset(self) -> QuerySet:
         queryset = CaseDocumentReference.objects.filter(
             document_type=CaseDocumentReference.Type.CERTIFICATE,
             export_application_certificates__status=ExportApplicationCertificate.Status.ACTIVE,
-            export_application_certificates__case_completion_datetime__date__gte=self.filters.date_from,
-            export_application_certificates__case_completion_datetime__date__lte=self.filters.date_to,
+            export_application_certificates__case_completion_datetime__date__range=(
+                self.filters.date_from,
+                self.filters.date_to,
+            ),
```
MattHolmes123, line 120: Nice!

web/reports/interfaces.py
```diff
@@ -279,3 +297,108 @@ def process_results(self) -> list[BaseModel]:
...
+                    self.filters.date_to,
+                )
+            )
+
+        queryset = (
+            ImportApplication.objects.filter(_filter)
+            .filter(
+                status__in=[ImpExpStatus.COMPLETED, ImpExpStatus.WITHDRAWN, ImpExpStatus.REVOKED]
+            )
+            .order_by("-reference", "-variation_no")
```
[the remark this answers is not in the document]
marcuspp: The ordering is copied from V1 and `variation_no` is populated in historical applications.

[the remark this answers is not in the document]
MattHolmes123: the reference contains the variation number in it though right?
e.g.
`IMA/2024/00001/1`
what does `variation_no` add?

mikeylockwood, line 324: `reference` is already unique and contains the `variation_no`

web/reports/interfaces.py
```diff
@@ -279,3 +297,108 @@ def process_results(self) -> list[BaseModel]:
...
+        if licence.licence_start_date and licence.licence_end_date:
+            return f"{licence.licence_start_date.strftime('%d %b %Y')} - {licence.licence_end_date.strftime('%d %b %Y')}"
+        return ""
+
+    def serialize_row(self, import_application: ImportApplication) -> ImportLicenceSerializer:
+        ia = import_application.get_specific_model()
+        is_adhoc = import_application.process_type == ProcessTypes.SANCTIONS
+        is_sps = import_application.process_type == ProcessTypes.SPS
+
+        licences = ia.licences.exclude(status=ImportApplicationLicence.Status.DRAFT)
```
mikeylockwood, line 341: Should `ARCHIVED` also be excluded?

web/reports/interfaces.py
```diff
@@ -279,3 +297,108 @@ def process_results(self) -> list[BaseModel]:
...
+
+        first_licence = licences.first()
+        latest_licence = licences.last()
+        doc_pack = latest_licence.document_references.filter(document_type="LICENCE").first()
```
mikeylockwood, line 345: I think there should only be one licence doc on the doc pack so `latest_licence.document_references.get(document_type="LICENCE")` should work

web/reports/interfaces.py
```diff
@@ -279,3 +297,108 @@ def process_results(self) -> list[BaseModel]:
...
+    ) -> str:
+        ia = import_application.get_specific_model()
+        if is_sps and ia.commodity:
+            return f"Code: {ia.commodity.commodity_code}"
+
+        if not is_adhoc:
+            return ""
+
+        details = []
+        for goods in ia.sanctions_goods.all().order_by("-commodity"):
```
mikeylockwood, line 401: This also works without the `.all()`

web/reports/interfaces.py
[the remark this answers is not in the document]
marcuspp: Oh got you. I'll remove 👍

web/reports/interfaces.py
MattHolmes123: I do it more for readability :) 
But you've been telling me to remove `.all()` for years at this point :D

web/reports/interfaces.py
```diff
@@ -279,3 +297,108 @@ def process_results(self) -> list[BaseModel]:
...
+
+        first_licence = licences.first()
+        latest_licence = licences.last()
+        doc_pack = latest_licence.document_references.filter(document_type="LICENCE").first()
+        licence_ref = doc_pack.reference if doc_pack else None
+        licence_type = "Paper" if latest_licence.issue_paper_licence_only else "Electronic"
+        sub_type = ia.application_type.sub_type
+
+        if is_adhoc or is_sps:
+            sub_type = f"{ia.application_type.type}1"
```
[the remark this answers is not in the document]
marcuspp: Left blank

web/reports/interfaces.py
```diff
@@ -279,3 +297,126 @@ def process_results(self) -> list[BaseModel]:
...
+                doc_pack = latest_licence.document_references.filter(
+                    document_type=CaseDocumentReference.Type.LICENCE
+                ).first()
+                licence_ref = doc_pack.reference
+
+        return self.ReportSerializer(
+            case_reference=ia.get_reference(),
+            licence_reference=licence_ref,
+            licence_type=licence_type,
+            under_appeal="",  # Unused field in V1
```
MattHolmes123, line 377: We never implemented the under_appeal feature as like you've said it's not used in V1.
Might as well delete it

web/reports/interfaces.py
```diff
@@ -279,3 +297,126 @@ def process_results(self) -> list[BaseModel]:
...
+            ima_type=ia.application_type.type,
+            ima_type_title=ia.application_type.get_type_display(),
+            ima_sub_type=ia.application_type.sub_type,
+            ima_sub_type_title=ia.application_type.get_sub_type_display(),
+            variation_number=ia.variation_no,
```
MattHolmes123, line 382: I would rather we use V2 logic to work this out instead of a field we don't use anywhere in V2.
e.g. split the reference on `/` and work out if it's a variation.

web/reports/models.py
```diff
@@ -38,6 +41,18 @@ class ScheduleReport(models.Model):
...
     )
 
+    @property
+    def parameters_str(self):
+        return ", ".join(
+            [
+                f"{key.replace('_', ' ')}: {value}" if value else ""
+                for key, value in self.parameters.items()
+            ]
+        )
```
MattHolmes123, line 51: I don't see this used anywhere?

MattHolmes123 (changes requested):

web/reports/models.py
marcuspp: Yes sorry i'll remove

web/reports/interfaces.py
```diff
@@ -279,3 +297,126 @@ def process_results(self) -> list[BaseModel]:
...
+            licence_start_date = None
+            licence_end_date = None
+            case_completion_datetime = None
+            latest_case_completion_datetime = None
+            time_to_initial_close = ""
+            licence_type = ""
+            licence_ref = ""
+        else:
+            # Archived licences included for revoked applications
+            licences = ia.licences.exclude(status__in=[ImportApplicationLicence.Status.DRAFT])
```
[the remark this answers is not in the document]
marcuspp: `pack_licence_history` doesn't include archived applications so they are different

web/reports/interfaces.py
marcuspp: I want the report to be same as V1 for easy comparison this column appears in the middle of the file. This could be an improvement after go live. (Where all fields in all reports should be reviewed).

web/reports/interfaces.py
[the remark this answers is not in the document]
MattHolmes123: It does include Archived licences that were once live:
Here is an example where I created an application and then completed a variation request.
image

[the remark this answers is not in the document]
MattHolmes123: The `pack_licence_history` code doesn't filter on status, it uses the case_reference field as thats what gets set when an application is marked as live:
image

Your code will include licences that are archived that I can't imagine we'd care about:
image

web/reports/interfaces.py
MattHolmes123: Fair enough, if you know the place you could just remove the column from your comparison code though.
But leave it in if it makes your life easier.

web/reports/interfaces.py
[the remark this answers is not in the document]
MattHolmes123: Unless we do care about everything that is :)

MattHolmes123 (approved):

PR #1329, opened 2024-02-12
ICMSLST-2500 Remove system packages not needed for docker deployment.

MattHolmes123:


MattHolmes123:
Unittests and end to end tests all pass locally when I removed those packages.

marcuspp (approved):

PR #1336, opened 2024-02-15
ICMSLST-1694 Create UniqueReference objects for licences and certificates

mikeylockwood:
Create UniqueReference objects from `ImportApplication.reference`, `ExportApplication.reference` in the `post_migration` script.
Create  UniqueReference objects from  `CaseDocumentReference.reference` in the queries that pull licence and certificate documents


data_migration/management/commands/check_data_counts.py
```diff
@@ -97,6 +105,34 @@ def run_counts(self) -> None:
...
+        with oracledb.connect(**CONNECTION_CONFIG) as connection:
+            for check in UNIQUE_REFERENCES:
+                ref = (
+                    UniqueReference.objects.filter(**check.filter_params)
+                    .order_by("-year", "-reference")
+                    .first()
+                )
+
+                result = self.run_query(connection, check.query, check.bind_vars)
+                self._log_result(check.name, result, ref.reference)
```
MattHolmes123, line 134: Is this used anywhere?

data_migration/management/commands/post_migration.py
```diff
@@ -15,23 +29,82 @@
...
+        formatted_reference = Func(
+            F("reference"),
+            Value("/"),
+            function="regexp_split_to_array",
+            output=ArrayField(TextField()),
+        )
+
+        if ref_type == "import":
+            model = ImportApplication
+            filter = {"reference__isnull": False, "legacy_case_flag": False}
```
MattHolmes123, line 72: I try to not shadow builtin names.
`filter` -> `filters`

data_migration/tests/test_e2e.py
```diff
@@ -609,6 +621,7 @@ def test_import_oil_data(mock_connect, dummy_dm_settings):
     assert oil2.section2 is False
 
     call_command("import_v1_data", "--skip_export")
+    call_command("post_migration", "--skip_perms")
```
chris-pettinga, line 624: I wonder if it doesn't make more sense to split the 2 bits of functionality into 2 management commands so it's easier to understand what's going on here, i.e. 1 creating_permissions command and 1 create_references command. 

The current approach also means that if we add a new 'chunk' of functionality to the post_migration management command, we'll have to update the tests here to also skip.

chris-pettinga: This comment also applies to all other calls to `call_command("post_migration", "--skip_XXX")`

data_migration/models/import_application/import_application.py
```diff
@@ -106,16 +104,33 @@ def get_values_kwargs(cls) -> dict[str, Any]:
...
 
+    @classmethod
+    def get_source_data(cls) -> Generator:
+        values = cls.get_values()
+        values_kwargs = cls.get_values_kwargs()
+        related = cls.get_related()
+        exclude_parameters = cls.get_exclude_parameters()
+
+        licence_reference = UniqueReference.objects.filter(
+            prefix="ILD", ur_imad_id=OuterRef("imad_id")
```
MattHolmes123, line 115: `UniqueReference.Prefix.IMPORT_LICENCE_DOCUMENT` ?

MattHolmes123 (approved):

PR #1352, opened 2024-02-26
ICMSLST-2521 - Stop SIGL header from appearing

chris-pettinga:


MattHolmes123 (approved):

web/templates/web/domains/case/manage/manage.html
```diff
@@ -37,7 +37,6 @@ <h4>Case Stopping</h4>
   {% endif %}
 
   {% if case_type == "import" %}
-    <h4>SIGL</h4>
```
mikeylockwood, line 40: Also remove the comment on the following line. It looks like this was missed when line 51 was implemented

mikeylockwood (approved):

PR #1355, opened 2024-02-27
Fix form rendering.

MattHolmes123:
Changes:
  - Add form to it's own row
  - Render as p to fix style issue
  - Set default cookie choice to No.


MattHolmes123:
Before:
image

After:
image

chris-pettinga (approved):

PR #1356, opened 2024-02-27
Use pydantic to serialize and deserialize the cookie.

MattHolmes123:


MattHolmes123:
You'll have to fix the tests though :)

chris-pettinga (approved):

PR #1360, opened 2024-02-27
ICMSLST-1190 Refactor COM application template code.

MattHolmes123:
Changes:
  - Create models to hold template data.
  - Remove old json column used to save template data
  - Add app specific sidebar links, will be used in CFS application template.


web/domains/cat/models.py
```diff
@@ -95,7 +58,34 @@ def user_can_edit(self, user: "User") -> bool:
...
+    case_emails = None
+    submitted_by = None
+    created_by = None
+    exporter = None
+    exporter_office = None
+    contact = None
+    agent = None
+    agent_office = None
+    case_owner = None
+    cleared_by = None
```
MattHolmes123, line 81: In future I might move these fields from `ExportApplicationABC` to `ExportApplication` so that these don't have to be removed for the template classes.

web/domains/cat/views.py
```diff
@@ -154,28 +196,33 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
...
+        ExportApplicationType.Types.FREE_SALE: CertificateOfManufactureTemplateForm,
+        # TODO: Use correct form
+        ExportApplicationType.Types.GMP: CertificateOfManufactureTemplateForm,
+        ExportApplicationType.Types.MANUFACTURE: CertificateOfManufactureTemplateForm,
+    }
+
+    try:
+        return types_forms[type_code]
+    except KeyError:
+        raise NotImplementedError(f"Type not supported: {type_code}")
```
MattHolmes123, line 225: Copied this from export/form.py.

Can change to a match statement.

web/domains/cat/models.py
mikeylockwood: I didn't think you could define FK relationships on an ABC

web/domains/cat/views.py
mikeylockwood: We tend to use process type rather than app type also

MattHolmes123: In this instance it's because the template model just stores the ExportApplication.Type string:
image

PR #1362, opened 2024-02-28
Fix for add report data when resetting local db

marcuspp:
Commands need to be separate as `add_reports_data` was interrupted as a argument to the other command.

### DIFF
--- a/scripts/reset-local-docker-db
+++ b/scripts/reset-local-docker-db
@@ -6,7 +6,8 @@ make down
 docker volume rm icms_pgdata
 make migrate
 make localstack
-make manage args="create_icms_groups add_reports_data"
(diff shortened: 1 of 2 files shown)


MattHolmes123 (approved):

scripts/reset-local-docker-db
```diff
@@ -6,7 +6,8 @@ make down
 docker volume rm icms_pgdata
 make migrate
 make localstack
-make manage args="create_icms_groups add_reports_data"
+make manage args="create_icms_groups"
+make manage args="add_reports_data"
```
MattHolmes123, line 10: Have you added this command to the jenkins jobs?

MattHolmes123 (approved):

PR #1369, opened 2024-03-01
Scanning the entire codebase for PII with pii-secret-check-hooks

chris-pettinga:
PII secrets checking currently runs as a pre-commit hook, meaning it only scans files that have been touched as part of the commit. This works well, however it means that casual changes to file X result in a full scan of the entire file meaning that lines of code that have remain unchanged are suddenly picked up as potentially containing sensitive data.

This PR runs the secret checker on the entire codebase such that previously 'missed' secrets have all been flagged and excluded, meaning that the only time the secret checker should flag a new bit of potentially dangerous information is when it has just been added to the codebase.

### DIFF
--- a/pii-ner-exclude.txt
+++ b/pii-ner-exclude.txt
@@ -3908,3 +3908,262 @@ Enter Manually
 No Firearms to Report
 Supplementary Report</h3
 Supplementary Report</h3
+Prepare Task
+AI
+AI 4</NAME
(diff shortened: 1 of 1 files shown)


MattHolmes123:
`pre-commit run --all-files` has been ran several times in the past.
I swear the pii tool changes its mind... It's really not that useful

marcuspp (approved):

PR #1380, opened 2024-03-06
ICMSLST-2105 - Firearms Licences Reports

marcuspp:


web/reports/interfaces.py
```diff
@@ -116,6 +110,54 @@ def get_queryset(self) -> QuerySet:
...
+            eori_number=ia.importer.eori_number,
+            importer_address=utils.get_importer_address(ia),
+            submitted_datetime=ia.submit_datetime.date(),
+            last_submitted_datetime=ia.last_submit_datetime.date(),
+            licence_start_date=licence.start_date,
+            licence_end_date=licence.end_date,
+            coo_country_name=getattr(ia.origin_country, "name", ""),
+            coc_country_name=getattr(ia.consignment_country, "name", ""),
+            endorsements="\n".join(ia.endorsements.values_list("content", flat=True)),
+            revoked=(True if ia.status == ImportApplication.Statuses.REVOKED else False),
```
MattHolmes123, line 152: `revoked=ia.status == ImportApplication.Statuses.REVOKED`

web/reports/interfaces.py
```diff
@@ -116,6 +110,54 @@ def get_queryset(self) -> QuerySet:
...
+            **self.extra_fields(ia),
+        )
+
+    def extra_fields(self, ia: FaImportApplication) -> dict:
+        raise NotImplementedError
```
MattHolmes123, line 157: consider typing as `dict[str, Any]` just to indicate the keys should be valid keyword arguments for the serializers.

web/reports/interfaces.py
```diff
@@ -116,6 +110,54 @@ def get_queryset(self) -> QuerySet:
         raise NotImplementedError
 
 
+class BaseFirearmsLicenceInterface(ReportInterface):
+    ReportFilter = BasicReportFilter
+    filters: BasicReportFilter
+
+    def get_application_filter(self):
```
MattHolmes123, line 117: missing type

web/reports/interfaces.py
```diff
@@ -563,3 +545,56 @@ def serialize_row(
...
             goods_description_with_sub_section=s.get_description(),
         )
+
+
+@final
+class DFLFirearmsLicenceInterface(BaseFirearmsLicenceInterface):
+    name = "Deactivated Firearms Licences"
+    ReportSerializer = DFLFirearmsLicenceSerializer
+
+    def get_application_filter(self):
```
MattHolmes123, line 555: missing type for all get_application_filter

web/reports/serializers.py
```diff
@@ -166,3 +167,68 @@ class SupplementaryFirearmsSerializer(pydantic.BaseModel):
     )
     transport: str_or_empty = pydantic.Field(serialization_alias="Means of Transport")  # /PS-IGNORE
     all_reported: yes_no = pydantic.Field(serialization_alias="Reported all firearms for licence")
+
+
+class BaseFirearmsLicenceSerializer(pydantic.BaseModel):
+    licence_reference: str_or_empty = pydantic.Field(serialization_alias="Licence Reference")
```
MattHolmes123, line 173: General comment that applies to all Serializers:
https://docs.pydantic.dev/latest/concepts/alias/#using-an-aliasgenerator
_This is particularly useful if you need to use different naming conventions for loading and saving data, but you don't want to specify the validation and serialization aliases for each field individually._

You could reduce a lot of the noise on this models

MattHolmes123: AliasGenerator was added in 2.6 so you'd need to upgrade pydantic

MattHolmes123: e.g. (on one of the other reports):
```
from pydantic import AliasGenerator, ConfigDict


def get_report_field_name(field_name: str) -> str:
    match field_name:
        case "certificate_reference":
            return "Certificate Reference"
        case "case_reference":
            return "Case Reference"

class IssuedCertificateReportSerializer(pydantic.BaseModel):
    model_config = ConfigDict(
        alias_generator=AliasGenerator(
            serialization_alias=get_report_field_name,
        )
    )
```

MattHolmes123 (approved):

mikeylockwood (approved):

PR #1384, opened 2024-03-06
ICMSLST-2553 - Corrupt file generated when P12_SIGNATURE envs not set.

chris-pettinga:
When the P12_SIGNATURE_BASE_64 / P12_SIGNATURE_PASSWORD env variables are not set, the object returned by the sign_pdf function is bytes, and not a stream at position 0 that can be read as the following code expects it to be.


web/domains/case/tasks.py
```diff
@@ -84,6 +84,7 @@ def create_import_application_document(
 
         # Add digital signature to the licence pdf
         file_obj = signer.sign_pdf(file_obj)
```
MattHolmes123, line 86: There is no need for `signed_pdf.seek(0)` now right?
Unless you need it for a test or something

MattHolmes123 (approved):

web/domains/case/tasks.py
chris-pettinga: Yeah I'm going to remove it and add it to other tests that call this function. In my mind the function only has a responsibility to return a BytesIO object and whoever calls it has the responsibility to set the pointer to where they want it.

marcuspp (approved):

PR #1386, opened 2024-03-07
ICMSLST-1641 - show extra question when selected Legislation makes Biocidal claim.

chris-pettinga:
Before:
image

After:
image


mikeylockwood (approved):

web/static/web/js/pages/cfs-schedule-edit.js
```diff
@@ -153,3 +154,32 @@ function showResponsiblePersonStatement(legislations, exportOnly, config) {
...
+        break;
+      }
+    }
+  }
+  const isBiocidalClaimQuestion = document.querySelector("#biocidal-claim-question-wrapper");
+  if (isBiocidalClaim) {
+    isBiocidalClaimQuestion.style.display = "block";
+  } else {
+    isBiocidalClaimQuestion.style.display = "none";
+  }
```
MattHolmes123, line 184: We do have a utility function to show / hide an element that is used in several places.
```javascript
const showBiocidalClaim = UTILS.getShowElementFunc("#biocidal-claim-question-wrapper")
showBiocidalClaim(isBiocidalClaim)
```

MattHolmes123: Also this 🤷 
```javascript
isBiocidalClaimQuestion.style.display = isBiocidalClaim ? "block" : "none";
```

chris-pettinga: Thanks for this, I've gone for the 2nd option as I think it's clearer.

chris-pettinga: @MattHolmes123  - I've also added some stuff at the top of this file to deal with show/hide logic on page load - not sure if it's the standard approach so let me know.

web/static/web/js/pages/cfs-schedule-edit.js
```diff
@@ -9,6 +9,9 @@ window.addEventListener('load', (event) => {
 
   const events = new EditScheduleEventHandler(legislationSelect2);
 
+  // Show-hiding the is-biocidal-claim question on page load
+  showIsBiocidalClaim(events.legislationSelect2.val(), events.legislationConfig);
+
```
MattHolmes123, line 14: Do you need to do this on page load?
You do it in the template already

``` 
  <div id="biocidal-claim-question-wrapper" {% if not is_biocidal %} style="display: none;"{% endif %}>
    {{ fields.field(form.biocidal_claim, mark_safe=True, show_optional_indicator=False) }}
  </div>
```

web/static/web/js/pages/cfs-schedule-edit.js
MattHolmes123: > his, I've gone for the 2nd option as I think it's clearer.

Fair enough @chris-pettinga .
The utility function is more suited to when you need to do something with callbacks as well.

MattHolmes123: > @MattHolmes123 - I've also added some stuff at the top of this file to deal with show/hide logic on page load - not sure if it's the standard approach so let me know.

I've replied @chris-pettinga

web/static/web/js/pages/cfs-schedule-edit.js
chris-pettinga: @MattHolmes123 the hidden field should be required when they've added biocidal legislation, and so when the form is invalid and the page re-renders the `is_biocidal` variable is False (because the changes haven't been saved to the db) but we want the field to be shown in order to display the validation errors.

We could in theory check for `form.cleaned_data.legislations.filter(is_biocidal_claim=True)` in the template but that seems out of step and a bit too complex, seems easier to just re-run the JS logic on page load. However this unfortunately makes the page inaccessible to those without JS

web/domains/case/export/forms.py
```diff
@@ -342,6 +345,23 @@ def __init__(self, *args, **kwargs):
...
+        self.fields["biocidal_claim"].help_text = render_to_string(
+            "web/domains/case/export/partials/cfs/biocidal_claim_help_text.html"
+        )
+
+    def clean_biocidal_claim(self):
+        biocidal_claim = self.cleaned_data["biocidal_claim"]
+
+        if (
+            not biocidal_claim
+            and self.cleaned_data["legislations"].filter(is_biocidal_claim=True).exists()
```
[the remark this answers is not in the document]
chris-pettinga: good point, will change

MattHolmes123 (approved):

PR #1387, opened 2024-03-07
ICMSLST-1624 - allow update of historical record without EORI Number present.

chris-pettinga:
As per the ILB Team feedback, older Importer records may not have an EORI Number, as previously is was not a mandatory value.

The following changes should be implemented:

Check if an Importer has an EORI Number entered when the Admin initiates editing of the Importer record

When [Save] is used, if the EORI Number was populated on entry, maintain mandatory value checks on this field

When [Save] is used, if there was no EORI Number on the record, skip the mandatory value checks on this specific field only

### DIFF
--- a/web/domains/importer/forms.py
+++ b/web/domains/importer/forms.py
@@ -25,8 +25,9 @@ class ImporterIndividualForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        self.fields["eori_number"].required = True
+        # EORI number is required if it's already in the DB, or if it's a new importer
+        if self.instance.eori_number or not self.instance.pk:
+            self.fields["eori_number"].required = True
 
     def clean(self):
         """Set type as individual as Importer can be an organisation too."""
@@ -35,11 +36,11 @@ class ImporterIndividualForm(forms.ModelForm):
 
     def clean_eori_number(self):
         """Make sure eori number starts with GBPR."""
-        eori_number = self.cleaned_data["eori_number"]
-        prefix = "GBPR"
-        if eori_number.startswith(prefix):
+        if eori_number := self.cleaned_data.get("eori_number"):
+            prefix = "GBPR"
+            if not eori_number.lower().startswith(prefix.lower()):
+                raise ValidationError(f"'{eori_number}' doesn't start with {prefix}")
             return eori_number
-        raise ValidationError(f"'{eori_number}' doesn't start with {prefix}")
 
 
 class ImporterIndividualNonILBForm(ImporterIndividualForm):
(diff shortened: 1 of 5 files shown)


web/domains/importer/forms.py
```diff
@@ -35,11 +35,11 @@ def clean(self):
 
     def clean_eori_number(self):
         """Make sure eori number starts with GBPR."""
-        eori_number = self.cleaned_data["eori_number"]
-        prefix = "GBPR"
-        if eori_number.startswith(prefix):
+        if eori_number := self.cleaned_data.get("eori_number"):
+            prefix = "GBPR"
+            if not eori_number.startswith(prefix):
+                raise ValidationError(f"'{eori_number}' doesn't start with {prefix}")
```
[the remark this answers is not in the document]
chris-pettinga: Same for below

MattHolmes123:
I do wonder about this one.
EORI is a required field and sending data to HMRC will fail without it.

Questions I have:
1. Why are the editing old importer records?
2 If they are, are these importers still using ICMS?

If the answer to 2 is yes then they will need to enter the EORI number.

I also don't think we should allow new importer records to be created without an EORI for the same reason.

web/tests/domains/importer/test_forms.py
```diff
@@ -106,3 +102,69 @@ def test_type_importer_organisation_form(api_get_company):
...
+            "locality": "Bruxelles",
+        }
+    }
+
+    importer.eori_number = None
+    importer.save()
+    data = {"name": "hello", "eori_number": "", "registered_number": "42"}
+    form = ImporterOrganisationForm(instance=importer, data=data)
+
+    assert form.is_valid()
```
MattHolmes123, line 170: We need a "assert an EORI number is required when it's a new record" as a minumim for this change to be merged.
I'd rather we kick this back to ILB and say we aren't doing it though.

chris-pettinga:
> I do wonder about this one. EORI is a required field and sending data to HMRC will fail without it.
> 
> Questions I have:
> 
> 1. Why are the editing old importer records?
>    2 If they are, are these importers still using ICMS?
> 
> If the answer to 2 is yes then they will need to enter the EORI number.
> 
> I also don't think we should allow new importer records to be created without an EORI for the same reason.

Didn't know that it was necessary for HMRC. So essentially, no importer can fully complete an application without their EORI number? What happens if an importer without an EORI number submits an application, are there checks for the EORI number before it gets sent to CHIEF, and if so, how is this displayed to ILB?

If it's all handled gracefully, I would say that this change is alright, from a UX perspective it's annoying to force ILB to find and enter the EORI number if they don't actually have to at that point in time (i.e. if they don't have any active applications). We should only force them to do something when it's strictly necessary IMO

web/domains/importer/forms.py
```diff
@@ -35,11 +36,11 @@ def clean(self):
...
```
MattHolmes123, line 42: I know the old version didn't but maybe we should make the comparison be case insensitive.
e.g. `eori_number.casefold().startswith("gbpr")`

web/domains/importer/forms.py
```diff
@@ -90,13 +94,13 @@ def clean_registered_number(self):
...
     def clean_eori_number(self):
         """Make sure eori number starts with GB."""
-        eori_number = self.cleaned_data["eori_number"]
-        prefix = "GB"
+        if eori_number := self.cleaned_data.get("eori_number"):
+            prefix = "GB"
 
-        if eori_number.startswith(prefix):
-            return eori_number
+            if not eori_number.startswith(prefix):
```
MattHolmes123, line 100: same casefold here.

MattHolmes123 (approved):

PR #1392, opened 2024-03-13
DPM-159 - DBT Platform build configuration

acodeninja:


acodeninja:
@MattHolmes123 and @chris-pettinga feel free to review and merge at your leisure.

.copilot/image_build_run.sh
```diff
@@ -0,0 +1,17 @@
...
+export COPILOT_ENVIRONMENT_NAME='build'
+
+echo "Running python manage.py collectstatic --noinput --traceback"
+python manage.py collectstatic --noinput --traceback
+
+echo "Running compress python manage.py compress --force --engine jinja2"
+python manage.py compress --force --engine jinja2
+
+echo 'Running PLAYWRIGHT_BROWSERS_PATH="/workspace/playwright-deps" python -m playwright install'
+PLAYWRIGHT_BROWSERS_PATH="/workspace/playwright-deps" python -m playwright install
```
MattHolmes123, line 17: How come you needed to change the browser path?
Is the default not suitable:
`~/.cache/ms-playwright`
https://playwright.dev/python/docs/browsers#managing-browser-binaries

acodeninja: At the time this script runs `~/.cache` is not writeable, all changes made by these scripts can only update files in the `/workspace` directory.

MattHolmes123: This will break PDF generation without another code change then but it's something we can fix after this is merged.
(we will need to update the library code to point to where the browsers are installed).

acodeninja: Doesn't playwright work with the lines in the celery entrypoint and the main entry script where we export the env var so playwright knows where the browsers are cached?

```shell
export PLAYWRIGHT_BROWSERS_PATH="/workspace/playwright-deps"
```

MattHolmes123: Yer I was just wondering this after I said the above.
We will find out when we test it I guess :)

MattHolmes123 (approved):

PR #1396, opened 2024-03-14
ICMSLST-1622 - Importer/Exporter - create new without need to match Reg No. on Companies House.

chris-pettinga:
image

Create entity

_If the entered Registration Number is not on Companies House, do not show an error message and do not prevent creation of the new Exporter or Importer

Instead, if the entered Registration Number is not on Companies House, show a pop-up message when [Create] is used that informs the ILB Admin that there is no company with that number and to confirm they want to continue and create the Exporter or Importer._

Save an Edited entity

_If the amended Registration Number is not on Companies House, show a pop-up message when [Save] is used that informs the ILB Admin that there is no company with that number and to confirm they want to continue and save the Exporter or Importer_


web/domains/exporter/forms.py
```diff
@@ -40,6 +40,9 @@ class Meta:
         fields = ["name", "registered_number", "comments"]
         widgets = {"name": forms.Textarea(attrs={"rows": 1})}
 
+    class Media:
+        js = ("web/js/api/check_company_number.js", "web/js/pages/edit-importer-exporter.js")
```
MattHolmes123, line 44: `edit-importer-exporter.js` -> `edit-organisation.js`?
We refer to importers and exporters as orgs in a lot of places in ICMS

web/static/web/js/pages/edit-importer-exporter.js
```diff
@@ -0,0 +1,48 @@
...
+
+    const company_number = $("#id_registered_number").val()
+    GetCompanyFromCompaniesHouse(company_number).then(response => response.json()).then(data => {
+        if (data.status === "not-found") {
+            const d = Dialogue();
+            const message = "Registered number not found on Companies House, are you sure you want to continue?";
+            d.show(message, submitEditForm);
+        } else {
+            submitEditForm()
+        }
```
MattHolmes123, line 46: Can't we use the status code instead of checking a text value?
An example using the postcode lookup:
```javascript
        postcodeLookup(postcode).then((response) => {
          response.json().then(results => {
            clearErrors(postcodeInputDiv);

            if (response.ok) {
              setAddressSelectDisplay(selectAddressWrapper, "block")
              processAddressOptions(selectAddress, results);
            } else {
              setAddressSelectDisplay(selectAddressWrapper, "none")
              processError(postcodeInputDiv, results);
            }
          });
        });
```

MattHolmes123: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

web/misc/views.py
```diff
@@ -46,3 +46,25 @@ def company_lookup(request: AuthenticatedHttpRequest) -> JsonResponse:
...
+def company_number_lookup(request: AuthenticatedHttpRequest) -> JsonResponse:
+    """Fetches a single company from CompaniesHouse from its company number"""
+    company_number = request.POST["company_number"]
+
+    try:
+        company_data = api_get_company(company_number)
+        status = "found"
+    except CompanyNotFound:
+        company_data = {}
+        status = "not-found"
```
MattHolmes123, line 62: This doesn't handle every outcome gracefully, see `def postcode_lookup` for error handling.
We should return one of three status codes, 200, 400, or 404

web/static/web/js/pages/edit-importer-exporter.js
MattHolmes123: I added a comment about the company_number_lookup view that relates to this

web/static/web/js/pages/edit-importer-exporter.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * Event handler for the submit button on the edit importer/exporter page
+ */
+(function ($) {
```
MattHolmes123, line 4: General comment about the file.
I don't really like that we are adding more jquery, we can do all of this with vanilla JS (As we've done in a lot of places).
Most of the jquery code is the code that was copied over from fivium.

web/domains/exporter/forms.py
MattHolmes123: Also I've never really linked JS to a form this way (despite having typed `{{ form.media.js }}` many times) :)

MattHolmes123 (changes requested):

web/static/web/js/pages/edit-importer-exporter.js
chris-pettinga: @MattHolmes123 I've never understood jQuery's bad reputation, everything written with jQuery can be done in Vanilla JS as it's just a wrapper around it, but isn't the whole point is that you can achieve the same result with a lot fewer lines and less complexity? I would say this is a lot easier to read/understand than the equivalent in vanilla JS. 

Happy to convert if required but I would be remiss not to bat for jQuery here.

MattHolmes123: Search for `$(document).ready` compared to `window.addEventListener('load'` and you can see which is more consistent

web/static/web/js/pages/edit-organisation.js
```diff
@@ -0,0 +1,48 @@
...
+
+    const company_number = $("#id_registered_number").val()
+    GetCompanyFromCompaniesHouse(company_number).then(response => {
+        if (response.status === 404) {
+            const d = Dialogue();
+            const message = "Registered number not found on Companies House, are you sure you want to continue?";
+            d.show(message, submitEditForm);
+        } else {
+            submitEditForm()
+        }
```
MattHolmes123, line 46: Is this the logic want?
e.g. if the server returns any other error code apart from a 404 you want to ignore it and submit the form?

web/static/web/js/api/check_company_number.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Gets company information from Companies House using the company number
+ * @param companyNumber
+ * @returns {Promise<Response>}
+ * @constructor
+ */
+async function GetCompanyFromCompaniesHouse(companyNumber) {
```
MattHolmes123, line 7: `GetCompanyFromCompaniesHouse` -> `getCompanyFromCompaniesHouse` ?

web/misc/views.py
```diff
@@ -46,3 +46,22 @@ def company_lookup(request: AuthenticatedHttpRequest) -> JsonResponse:
...
+@login_required
+@require_POST
+def company_number_lookup(request: AuthenticatedHttpRequest) -> JsonResponse:
+    """Fetches a single company from CompaniesHouse from its company number"""
+
+    if company_number := request.POST.get("company_number"):
+        try:
+            company_data = api_get_company(company_number)
+        except CompanyNotFound:
+            raise Http404()
```
MattHolmes123, line 60: This still doesn't handle unknown errors gracefully.

web/static/web/js/pages/edit-importer-exporter.js
chris-pettinga: @MattHolmes123 - yes I know where you're coming from, I'm just so jaded and burnt from vanilla JS in the past that I have a gut reaction to it, I have converted it to vanilla JS and it wasn't nearly as bad as I remember.

web/static/web/js/pages/edit-organisation.js
chris-pettinga: @MattHolmes123 - I  figure that if the CompaniesHouse API is down, we don't want to stop the user from submitting the form. If the company_number is malformed (e.g. if they've entered N/A) then I would rather leave that to the form's `clean` method then here.

This functionality is dealing purely with whether or not a company number can be found on CompaniesHouse and I think checking for a 404 is the most appropriate way to deal with that.

web/misc/views.py
chris-pettinga: a 404 is a valid response? We could return a JsonResponse with a dict like `{"error_msg": "not_found"}` but I feel that would be superfluous, a 404 status code conveys to the user all they need to know and the response body doesn't add anything IMO.

chris-pettinga: If an APIError is raised from the `api_get_company` function, it'll bubble up, get sent to Sentry, and cause a 500 which is kind of exactly what we want I think? An error has occurred, show an error page.

web/static/web/js/pages/edit-organisation.js
chris-pettinga: Basically this is a 'nice bit of info to know' and not a 'you must do this before saving changes'

web/misc/views.py
MattHolmes123: But this is a view that returns json and the user won't see an error page as it isn't a full page reload.
There will just be a message in the console saying their was a 500 error.

MattHolmes123 (approved):

PR #1402, opened 2024-03-19
ICMSLST-2487 All Countries and Overseas regions

marcuspp:
- Add countries on this [government list](https://www.gov.uk/government/publications/uk-trade-tariff-country-and-currency-codes/uk-trade-tariff-country-and-currency-codes) that aren't already on the platform but sets them to inactive
- Add overseas regions to all countries using [regions list](https://data.trade.gov.uk/datasets/240d5034-6a83-451b-8307-5755672f881b)
- Add continent column to issued certificates report
- Refactors country test data to use django fixtures
- Changed view to "Add Country" to activate an inactive country. 
- All commission codes / hmrc codes now preset and are not editable

To test
- `git checkout ICMSLST-2487`
- `./scripts/reset-local-docker-db`
 - In a shell `Country.objects.filter(is_active=False).delete()`
 - In a shell `Country.objects.update(overseas_region=None)`
 - `python manage.py add_v2_reference_data`
 - `python manage.py check_data_counts` - Command will fail as cannot talk to v1 but will print `PASS` for the countries checks prior to the error


web/migrations/0001_initial.py
```diff
@@ -23,8 +23,8 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ("auth", "0012_alter_user_first_name_max_length"),
         ("contenttypes", "0002_remove_content_type_name"),
+        ("auth", "0012_alter_user_first_name_max_length"),
```
chris-pettinga, line 27: interesting that the order of these keeps on switching around whenever anyone regenerates the migration file

MattHolmes123: Yer it kind of blows my mind that it's nondeterministic

web/templates/web/domains/country/add.html
```diff
@@ -0,0 +1,19 @@
...
+    {% include "web/domains/country/groups-tooltip.html" %}
+  {% endset %}
+  {{fields.field(form.name, tooltip=tooltip)}}
+{% endblock %}
+
+{% block actions %}
+<li>
+  <button id="btn-submit" type="submit" class="primary-button button">Add Country</button>
+</li>
+{% endblock %}
```
MattHolmes123, line 19: Missing newline

web/domains/country/views.py
```diff
@@ -59,14 +59,18 @@ class CountryEditView(ModelUpdateView):
...
+    template_name = "web/domains/country/add.html"
     form_class = CountryCreateForm
-    success_url = reverse_lazy("country:list")
-    cancel_url = success_url
-    page_title = "New Country"
+    page_title = "Add Country"
     permission_required = Perms.sys.ilb_admin
+    cancel_url = reverse_lazy("country:list")
+
+    def form_valid(self, form):
```
MattHolmes123, line 69: missing types

MattHolmes123 (approved):
I've approved this but I think @mikeylockwood is better placed to give it a real review.

web/templates/web/domains/country/add.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block sidebar %}
+  {% include "web/domains/country/sidebar.html" %}
+{% endblock %}
+
+{% block fields %}
+  {% set tooltip %}
+    {% include "web/domains/country/groups-tooltip.html" %}
+  {% endset %}
+  {{fields.field(form.name, tooltip=tooltip)}}
```
mikeylockwood, line 12: Minor but prefer whitespace within template tags
`{{ fields.field(form.name, tooltip=tooltip) }}`

web/templates/web/domains/country/edit.html
```diff
@@ -14,6 +14,9 @@
 
   {{fields.field(form.name, tooltip=tooltip)}}
   {{fields.field(form.type)}}
+  {{fields.field(form.overseas_region)}}
+  {{fields.field(form.is_active)}}
```
mikeylockwood, line 18: Template tag whitespace for this entire file

data_migration/management/commands/config/data_counts.py
```diff
@@ -49,6 +49,23 @@
...
+        model=web.Country,
+        filter_params={"overseas_region__isnull": True, "type": web.Country.SOVEREIGN_TERRITORY},
+    ),
+    CheckCount(
+        name="Countries",
+        expected_count=269,
+        model=web.Country,
+    ),
+    CheckCount(
+        name="Countries that are active",
```
mikeylockwood, line 64: Inactive countries are ones created by add_v2_reference_data and not the data migration?

mikeylockwood (approved):

data_migration/management/commands/config/data_counts.py
marcuspp: Correct. So check for total count and a check for the data migration part that you are left with the correct amount of active countries.

PR #1418, opened 2024-03-27
ICMSLST-591 - CFS Schedule shouldn't be editable

chris-pettinga:
Before:
image

After:
image


web/views/actions.py
```diff
@@ -113,14 +113,19 @@ class EditTemplate(LinkAction):
...
     def href(self, instance):
-        # TODO: add CFS_SCHEDULE and point it its special URL
         if instance.template_type == Template.CFS_DECLARATION_TRANSLATION:
             return reverse("template-cfs-declaration-translation-edit", kwargs={"pk": instance.pk})
         elif instance.template_type == Template.CFS_SCHEDULE_TRANSLATION:
             return reverse("template-cfs-schedule-translation-edit", kwargs={"pk": instance.pk})
 
         return f"{instance.id}/edit/"
 
+    def display(self, object):
```
chris-pettinga, line 123: Doing it in display() as the alternative - taking them to the 'create new cfs schedule translation' page - felt pretty unintuitive. People clicking 'edit' want to edit the english template which is not and should not be possible, they don't want to create a new translation. That was my thought process anyway.

MattHolmes123, line 123: Missing types

web/views/actions.py
```diff
@@ -113,14 +113,19 @@ class EditTemplate(LinkAction):
...
+        # ICMSLST-591 - CFS Schedule templates are not editable
+        if object.template_type == Template.CFS_SCHEDULE:
+            return False
+        return True
```
MattHolmes123, line 127: I would remove `ICMSLST-591` from the comment.
Regarding this comment in the story:
"CFS Schedule defines the reference text in English and it should not be editable"
Should nobody be able to edit the template then?

MattHolmes123 (approved):

PR #1422, opened 2024-03-27
ICMSLST-2556 Fix end_to_end tests and improve reliability

mikeylockwood:


chris-pettinga (approved):

web/end_to_end/utils.py
```diff
@@ -36,22 +38,48 @@ def get_future_datetime() -> dt.datetime:
...
 
-        else:
-            page.wait_for_timeout(5_000)
-            page.reload()
+        page.wait_for_timeout(timeout)
+        page.reload()
 
-    raise TimeoutError("Max retries waiting for Bypass CHIEF link.")
+    el = page.locator(locator_str).get_by_role(**role_data)
+    el.click()
```
MattHolmes123, line 82: I guess this will raise an error if it doesn't find it the final time?

MattHolmes123 (approved):

web/end_to_end/utils.py
mikeylockwood: Yes just the standard error when it can't find an element

PR #1427, opened 2024-04-02
ICMSLST-2570 - Using transparent signature image causes clashing signature/placeholder PDF

chris-pettinga:
Stopping users from uploading transparent PNGs to the signature field as transparency can cause issues when actually signing the PDF


chris-pettinga:
Initially thought about using Pillow to automatically convert the transparent image to a flat jpeg but decided against it for the following:

- The file is already in S3, so we would have to download it, convert it, then re-upload it - quite costly
- The conversion itself is not very intuitive and involves creating a separate white background PNG and then merging the 2
- It is prone to mistake, what if the user wanted a black background for their signature?

Therefore I just decided to prevent people from uploading transparent images in the first place

web/domains/file/utils.py
```diff
@@ -60,15 +57,34 @@ def __init__(self, *, validators=(), help_text=HELP_TEXT, **kwargs) -> None:
...
+    def __init__(
+        self,
+        *,
+        validators=(),
+        allowed_extensions: Iterable[str] = IMAGE_EXTENSION_ALLOW_LIST,
+        help_text: str | None = None,
+        **kwargs,
+    ) -> None:
+        self.allowed_extensions = allowed_extensions
+        help_text = help_text or IMAGE_HELP_TEXT + ", ".join(allowed_extensions)
```
MattHolmes123, line 69: Don't understand why you did this chagne?

MattHolmes123:
```
have a default value for the init method argument is cleaner?
def __init__(self, *, validators=(), help_text: str = IMAGE_HELP_TEXT, **kwargs) -> None:

File: web/domains/signature/forms.py [hunk 2]
```diff
@@ -10,5 +10,5 @@ class Meta:
         model = Signature
         fields = ("name", "signatory")
 
-    file = ImageFileField(required=True)
+    file = ImageFileField(required=True, allowed_extensions=["jpeg", "jpg"])
```
MattHolmes123, line 13: Maybe a comment here to say why we are excluding png files.

web/domains/file/utils.py
```diff
@@ -60,15 +57,34 @@ def __init__(self, *, validators=(), help_text=HELP_TEXT, **kwargs) -> None:
...
         super().__init__(
-            # order is important: validate_file_extension can delete the file
-            # from S3, so has to be after the virus check
-            validators=[validate_virus_check_result, validate_image_file_extension, *validators],
+            validators=[validate_virus_check_result, *validators],
             help_text=help_text,
             **kwargs,
         )
 
+    def clean(
```
MattHolmes123, line 76: As an alternative to changing the behaviour, by calling the validator in clean, we could use either a class that implements the `__call__` method or a closure.

`__call__` class:
```python
class ImageFileFieldValidator:
    def __init__(self, allowed_extensions: Iterable[str] = IMAGE_EXTENSION_ALLOW_LIST) -> None:
        self.allowed_extensions = allowed_extensions

    def __call__(self, file: S3Boto3StorageFile) -> None:
        """Django validator that only allows specific image file extensions."""

        _, file_extension = os.path.splitext(file.name)

        if file_extension.lstrip(".").lower() not in self.allowed_extensions:
            # by the time custom validations are called, file upload handlers have
            # already done their job and the file is in S3, so we have to delete it
            delete_file_from_s3(file.name)

            raise forms.ValidationError(
                "Invalid file extension. Only these extensions are allowed: "
                + ", ".join(self.allowed_extensions)
            )


class ImageFileField(forms.FileField):
    def __init__(
            self,
            *,
            validators=(),
            allowed_extensions: Iterable[str] = IMAGE_EXTENSION_ALLOW_LIST,
            help_text: str | None = None,
            **kwargs,
    ) -> None:
        validate_image_file_extension = ImageFileFieldValidator(allowed_extensions)
        help_text = help_text or IMAGE_HELP_TEXT + ", ".join(allowed_extensions)

        super().__init__(
            # order is important: validate_file_extension can delete the file
            # from S3, so has to be after the virus check
            validators=[validate_virus_check_result, validate_image_file_extension, *validators],
            help_text=help_text,
            **kwargs,
        )
```

Closure:
```python
def validate_image_file_extension_closure(allowed_extensions: Iterable[str] = IMAGE_EXTENSION_ALLOW_LIST):
    def validator(file: S3Boto3StorageFile) -> None:
        """Django validator that only allows specific image file extensions."""

        _, file_extension = os.path.splitext(file.name)

        if file_extension.lstrip(".").lower() not in allowed_extensions:
            # by the time custom validations are called, file upload handlers have
            # already done their job and the file is in S3, so we have to delete it
            delete_file_from_s3(file.name)

            raise forms.ValidationError(
                "Invalid file extension. Only these extensions are allowed: "
                + ", ".join(allowed_extensions)
            )

    return validator


class ImageFileField(forms.FileField):
    def __init__(
            self,
            *,
            validators=(),
            allowed_extensions: Iterable[str] = IMAGE_EXTENSION_ALLOW_LIST,
            help_text: str | None = None,
            **kwargs,
    ) -> None:
        validate_image_file_extension = validate_image_file_extension_closure(allowed_extensions)
        help_text = help_text or IMAGE_HELP_TEXT + ", ".join(allowed_extensions)

        super().__init__(
            # order is important: validate_file_extension can delete the file
            # from S3, so has to be after the virus check
            validators=[validate_virus_check_result, validate_image_file_extension, *validators],
            help_text=help_text,
            **kwargs,
        )
```

MattHolmes123: My preference is the closure (as I believe they are quicker)

MattHolmes123 (changes requested):
Added a comment.

web/domains/file/utils.py
chris-pettinga: @MattHolmes123 great spot and a much cleaner way to do it, thank you. Personally I think __call__ is a bit more intuitive to read in the spirit of OOP.

mikeylockwood (approved):

PR #1432, opened 2024-04-04
ICMSLST-2473 - Report Error handling

marcuspp:
Added Error handing to reports
 - If a pydantic validation error is raised this is caught and output as part of an error csv. This row is then skipped so the rest of the report can still be generated
 - If errors are found (in the data) then a message box is shown on the run output view
 - If any other processing occurs with the generator, this is also added to the error csv but is also flagged in sentry for further investigation
 - Added ability to delete a run (v1 feature)
 - Added ability to show/hide deleted runs from the run history (v1 feature)
 - Separate celery queue added (to give the ability to separate report generation to its own worker if required)

*History List View*
Screenshot 2024-04-08 at 15 26 28

*Run Output View*
Screenshot 2024-04-08 at 15 50 47


web/reports/views.py
```diff
@@ -40,13 +41,17 @@ class RunHistoryListView(BaseReportView, ListView):
...
+        if not self.request.GET.get("deleted"):
+            queryset = queryset.exclude(status=ReportStatus.DELETED)
+        return queryset.order_by("-finished_at")
 
     def get_context_data(self, *args, **kwargs) -> dict[str, Any]:
         context = super().get_context_data(*args, **kwargs)
         report = self.get_report()
         context["report"] = report
         context["page_title"] = report.name
+        context["showing_deleted"] = True if self.request.GET.get("deleted") else False
```
MattHolmes123, line 54: if you want a this would also work: 
`bool(self.request.GET.get("deleted"))`

web/reports/views.py
```diff
@@ -128,3 +133,20 @@ def get(self, *args, **kwargs) -> HttpResponse:
...
+
+    def post(self, request, *args, **kwargs) -> HttpResponseRedirect:
+        schedule_report = self.get_schedule_report()
+        schedule_report.status = ReportStatus.DELETED
+        schedule_report.deleted_by = request.user
+        schedule_report.deleted_at = timezone.now()
+        schedule_report.save(update_fields=["status", "deleted_by", "deleted_at"])
+        return HttpResponseRedirect(
+            reverse("run-history-view", kwargs={"report_pk": schedule_report.report.pk})
+        )
```
MattHolmes123, line 152: Do we care about atomic transactions? 
It's not as important as the import / export application views

web/reports/views.py
```diff
@@ -128,3 +133,20 @@ def get(self, *args, **kwargs) -> HttpResponse:
...
         return response
+
+
+class DeleteReportView(BaseReportView, RedirectView):
+    http_method_names = ["post"]
+
+    def get_schedule_report(self) -> ScheduleReport:
+        return get_object_or_404(ScheduleReport, pk=self.kwargs["schedule_pk"])
+
+    def post(self, request, *args, **kwargs) -> HttpResponseRedirect:
```
MattHolmes123, line 144: Can you add complete annotations here please.

web/reports/views.py
```diff
@@ -128,3 +133,20 @@ def get(self, *args, **kwargs) -> HttpResponse:
             f'attachment; filename="{generated_report.document.filename}"'
         )
         return response
+
+
+class DeleteReportView(BaseReportView, RedirectView):
+    http_method_names = ["post"]
+
+    def get_schedule_report(self) -> ScheduleReport:
+        return get_object_or_404(ScheduleReport, pk=self.kwargs["schedule_pk"])
```
MattHolmes123, line 142: Same thing can be achieved using SingleObjectMIxin but the lines you'd save in not implementing your own method would be taken up by the two exta config lines (`model` & `pk_url_kwarg`)

MattHolmes123 (approved):

PR #1434, opened 2024-04-08
ICMSLST-2550 Data Migration - Constabulary email attachments

mikeylockwood:


data_migration/models/case.py
```diff
@@ -319,6 +320,19 @@ def get_m2m_data(cls, target: models.Model) -> Generator:
         )
 
 
+class ConstabularyEmailAttachments(MigrationBase):
+    file_target = models.ForeignKey(FileTarget, on_delete=models.CASCADE)
+    caseemail = models.ForeignKey(CaseEmail, on_delete=models.CASCADE)
```
marcuspp, line 325: Convention is this should be `case_email`

marcuspp (approved):

data_migration/models/case.py
mikeylockwood: Just named as such because it mimics an automatically generated through table for a M2M relationship. Means I don't have to rename the field as I populate the data in that table

PR #1435, opened 2024-04-08
ICMSLST-1354 - SIIL Firearms Report - Include missing Upload Document option.

chris-pettinga:
Before:
image

After:
image

### DIFF
--- a/web/domains/case/_import/fa_sil/forms.py
+++ b/web/domains/case/_import/fa_sil/forms.py
@@ -614,39 +614,3 @@ class SILSupplementaryReportForm(forms.ModelForm):
             self.add_error(None, "You must enter this item.")
 
         return cleaned_data
-
-
-class SILSupplementaryReportFirearmSection1Form(forms.ModelForm):
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSection1
-        fields = ("serial_number", "calibre", "model", "proofing")
-
-
-class SILSupplementaryReportFirearmSection2Form(forms.ModelForm):
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSection2
-        fields = ("serial_number", "calibre", "model", "proofing")
-
-
-class SILSupplementaryReportFirearmSection5Form(forms.ModelForm):
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSection5
-        fields = ("serial_number", "calibre", "model", "proofing")
-
-
-class SILSupplementaryReportFirearmSection582ObsoleteForm(forms.ModelForm):  # /PS-IGNORE
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSection582Obsolete  # /PS-IGNORE
-        fields = ("serial_number", "calibre", "model", "proofing")
-
-
-class SILSupplementaryReportFirearmSection582OtherForm(forms.ModelForm):  # /PS-IGNORE
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSection582Other  # /PS-IGNORE
-        fields = ("serial_number", "calibre", "model", "proofing")
-
-
-class SILSupplementaryReportFirearmSectionLegacyForm(forms.ModelForm):
-    class Meta:
-        model = models.SILSupplementaryReportFirearmSectionLegacy
-        fields = ("serial_number", "calibre", "model", "proofing")
--- a/web/domains/case/_import/fa_sil/types.py
+++ b/web/domains/case/_import/fa_sil/types.py
@@ -41,14 +41,3 @@ SILReportFirearmModel = Union[
 ]
 
 SILReportFirearmModelT = type[SILReportFirearmModel]
-
-SILReportFirearmFormT = type[
-    Union[
-        forms.SILSupplementaryReportFirearmSection1Form,
-        forms.SILSupplementaryReportFirearmSection2Form,
-        forms.SILSupplementaryReportFirearmSection5Form,
-        forms.SILSupplementaryReportFirearmSection582ObsoleteForm,  # /PS-IGNORE
-        forms.SILSupplementaryReportFirearmSection582OtherForm,  # /PS-IGNORE
-        forms.SILSupplementaryReportFirearmSectionLegacyForm,
-    ]
-]
--- a/web/domains/case/_import/fa_sil/utils.py
+++ b/web/domains/case/_import/fa_sil/utils.py
@@ -1,0 +1,378 @@
+from typing import NamedTuple
+
+from django import forms as django_forms
+from django.forms.models import model_to_dict, modelform_factory
+from django.urls import reverse
+
+from web.domains.case._import.fa.forms import (
+    ImportContactKnowBoughtFromForm,
+    UserImportCertificateForm,
+)
+from web.domains.case.app_checks import get_org_update_request_errors
+from web.domains.file.utils import ICMSFileField
+from web.utils.validation import (
+    ApplicationErrors,
+    FieldError,
+    PageErrors,
+    create_page_errors,
+)
+
+from . import forms, models, types
+
+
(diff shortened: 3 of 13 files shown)


web/domains/case/_import/fa_sil/types.py
```diff
@@ -52,3 +52,14 @@
         forms.SILSupplementaryReportFirearmSectionLegacyForm,
     ]
 ]
+
+SILReportUploadFirearmFormT = type[
+    Union[
+        forms.SILSupplementaryReportUploadFirearmSection1Form,
+        forms.SILSupplementaryReportUploadFirearmSection2Form,
```
MattHolmes123, line 59: You can define unions with `|`
```
SILReportUploadFirearmFormT = type[
        forms.SILSupplementaryReportUploadFirearmSection1Form
        | forms.SILSupplementaryReportUploadFirearmSection2Form
        | forms.SILSupplementaryReportUploadFirearmSection5Form
        | forms.SILSupplementaryReportUploadFirearmSection582ObsoleteForm  # /PS-IGNORE
        | forms.SILSupplementaryReportUploadFirearmSection582OtherForm  # /PS-IGNORE
        | forms.SILSupplementaryReportUploadFirearmSectionLegacyForm
```

web/domains/case/_import/fa_sil/views.py
```diff
@@ -165,6 +166,30 @@ def _get_report_firearm_form_class(sil_section_type: str) -> types.SILReportFire
...
+    elif sil_section_type == "section582-obsolete":
+        return forms.SILSupplementaryReportUploadFirearmSection582ObsoleteForm  # /PS-IGNORE
+
+    elif sil_section_type == "section582-other":
+        return forms.SILSupplementaryReportUploadFirearmSection582OtherForm  # /PS-IGNORE
+
+    elif sil_section_type == "section_legacy":
+        return forms.SILSupplementaryReportUploadFirearmSectionLegacyForm
+
+    raise NotImplementedError(f"sil_section_type is not supported: {sil_section_type}")
```
MattHolmes123, line 190: Does it not format very nicely when using a `match` statement?

web/domains/case/_import/fa_sil/views.py
```diff
@@ -951,6 +976,91 @@ def add_report_firearm_manual(
...
+@login_required
+def view_upload_document(
+    request: AuthenticatedHttpRequest,
+    *,
+    application_pk: int,
+    report_firearm_pk: int,
+    sil_section_type: str,
+    **kwargs: dict[str, Any],
+) -> HttpResponse:
+    application: models.SILApplication = get_object_or_404(models.SILApplication, pk=application_pk)
```
MattHolmes123, line 1054: Do we need any permission checking (the same as the upload document view)?

web/domains/case/_import/fa_sil/views.py
mikeylockwood: I guess it was s copy of the rest of the code which predates `match` statements

MattHolmes123 (changes requested):
Needs the permission checking code.

web/domains/case/_import/fa_sil/views.py
chris-pettinga: Copied it from the other view_upload_document methods, the permissions checking is done in the return function - `view_application_file`:

image

MattHolmes123: I knew this... :D

chris-pettinga: @MattHolmes123  Happy to approve?

web/domains/case/_import/fa_sil/forms.py
```diff
@@ -650,3 +651,52 @@ class SILSupplementaryReportFirearmSectionLegacyForm(forms.ModelForm):
...
+        model = models.SILSupplementaryReportFirearmSection582Other  # /PS-IGNORE
+        fields = ("file",)
+
+
+class SILSupplementaryReportUploadFirearmSectionLegacyForm(forms.ModelForm):
+    file = ICMSFileField(required=True)
+
+    class Meta:
+        model = models.SILSupplementaryReportFirearmSectionLegacy
+        fields = ("file",)
```
MattHolmes123, line 702: Slighty wild idea to replace all these forms as well as the other `_get_report_upload_firearm_form_class` that simply returns one of the above based on a string.

```python
from django import forms
from django.forms import modelform_factory


def get_report_upload_firearm_form(sil_section_type) -> type[forms.ModelForm]:
    match sil_section_type:
        case "section1":
            model_cls = models.SILSupplementaryReportFirearmSection1
        case "section2":
            model_cls = models.SILSupplementaryReportFirearmSection2
        case "section5":
            model_cls = models.SILSupplementaryReportFirearmSection5
        case "section582-obsolete":
            model_cls = models.SILSupplementaryReportFirearmSection582Obsolete
        case "section582-other":
            model_cls = models.SILSupplementaryReportFirearmSection582Other
        case "section_legacy":
            model_cls = models.SILSupplementaryReportFirearmSectionLegacy
        case _:
            raise NotImplementedError(f"sil_section_type is not supported: {sil_section_type}")

    class FileForm(forms.ModelForm):
        document = ICMSFileField(required=True)

    upload_firearms_form = modelform_factory(model_cls, form=FileForm, fields=["document"])

    return upload_firearms_form
```

mikeylockwood: Nice, yeah the same could be done for the ReportFirearm forms for manually entering the data. Should change both if you're going to change one. Also be wary of changing the field name from `file` to `document`. There might be shared templates that are also using forms that expect the field to be `file` so would have to also be changed in DFL and OIL as well

MattHolmes123: I just did the example using the name of the model field.

MattHolmes123 (approved):
approved. 
Happy to go with whatever you decide with the forms.

web/domains/case/_import/fa_sil/utils.py
```diff
@@ -0,0 +1,379 @@
...
+    class SILSupplementaryReportFirearmSection1Form(forms.ModelForm):
+        class Meta:
+            model = models.SILSupplementaryReportFirearmSection1
+            fields = ("serial_number", "calibre", "model", "proofing")
+    """
+    model_class = _get_report_firearm_model(sil_section_type)
+
+    return modelform_factory(
+        model_class,
+        form=django_forms.ModelForm,
```
MattHolmes123, line 70: You don't need to specify a start form if you aren't doing anything fancy.
A model_class and fields list is enough.

web/domains/case/_import/fa_sil/utils.py
```diff
@@ -0,0 +1,379 @@
...
+    """Dynamically creates and returns a Form class for uploading a SIIL firearm report.
+
+    The resulting class looks something like the below:
+
+    class SILSupplementaryReportUploadFirearmSection1Form(forms.ModelForm):
+        file = ICMSFileField(required=True)
+
+        class Meta:
+            model = models.SILSupplementaryReportFirearmSection1
+            fields = ("file")
```
mikeylockwood, line 87: Mismatch field name in the docstring to field name in the form

chris-pettinga: @mikeylockwood - yes good spot

web/domains/case/_import/fa_sil/forms.py
chris-pettinga: There is an odd discrepancy between the form's field name and the attribute on the model, form calls it `file` but the model calls it `document`, I wonder if that's because the upload views use the `create_file_model` function to actually save the file, then assign it to the supp. report:

image

Going to leave the discrepancy for now and update the _get_report_upload_firearm_form_class method for now.

PR #1449, opened 2024-04-10
ICMSLST-2573 - Report Permissions

marcuspp:
Report Permissions updated
 - Each report type given its own page permission
 - ILB Admin group assigned all reports
 - NCA Admin group assigned just 2 of the reports
 - Removed report user


web/tests/domains/importer/test_forms.py
```diff
@@ -131,9 +131,10 @@ def test_eori_number_not_required_importer_individual_form(importer, importer_on
 
 
 @pytest.mark.django_db()
-def test_eori_numer_required_importer_organisation_form(importer):
+@patch("web.domains.importer.forms.api_get_company")
```
marcuspp, line 134: Not related, Fixed tests that fail when no wifi

web/permissions/service.py
```diff
@@ -18,7 +18,16 @@
...
+from web.permissions import (
+    ExporterObjectPermissions,
+    ImporterObjectPermissions,
+    PagePermissions,
+    Perms,
+)
+from web.reports.constants import ReportType
+
+if TYPE_CHECKING:
+    from web.models import Report
```
mikeylockwood, line 30: This is fine but we generally just import the model now with the other model imports

web/reports/views.py
```diff
@@ -41,6 +48,13 @@ def get_context_data(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
         context["page_title"] = "Reports"
         return context
 
+    def get_queryset(self) -> list[Report]:
+        reports = []
+        for report in Report.objects.all():
+            if can_user_view_report(report, self.request.user):
+                reports.append(report)
+        return reports
```
mikeylockwood, line 56: Could also be a list comprehension

```
return [
    report for report in Report.objects.all()
    if can_user_view_report(report, self.request.user)
]
```

web/permissions/service.py
marcuspp: Circular dependency if i do so

MattHolmes123: yep save the `if TYPE_CHECKING` clock for circular dependencies and expensive imports

web/permissions/service.py
```diff
@@ -442,3 +451,24 @@ def filter_users_with_org_access(org: ORGANISATION, users: QuerySet[User]) -> Qu
 
     # Note: We are ignoring user permissions by design
     return users.filter(groups__permissions__codename=org_access).distinct()
+
+
+def get_report_permission(report: "Report") -> PagePermissions:
```
MattHolmes123, line 456: `def get_report_permission(report: "Report") -> tuple[str, str]:`

web/permissions/service.py
```diff
@@ -442,3 +451,24 @@ def filter_users_with_org_access(org: ORGANISATION, users: QuerySet[User]) -> Qu
...
+            return Perms.page.view_report_access_requests
+        case ReportType.SUPPLEMENTARY_FIREARMS:
+            return Perms.page.view_report_supplementary_firearms
+        case ReportType.FIREARMS_LICENCES:
+            return Perms.page.view_report_firearms_licences
+        case _:
+            raise ValueError(f"Unknown Report Type {report.report_type}")
+
+
+def can_user_view_report(report: "Report", user: User) -> bool:
```
MattHolmes123, line 472: All of the `def can_user_xxx(` functions have `user` as the first paramater

web/reports/views.py
```diff
@@ -26,13 +26,20 @@
...
 
 class BaseReportView(LoginRequiredMixin, PermissionRequiredMixin):
-    permission_required = [Perms.page.view_reports]
+    permission_required = [Perms.sys.access_reports]
+
+    def has_permission(self) -> bool:
+        has_permission = super().has_permission()
+        if not has_permission:
+            return has_permission
+        return can_user_view_report(self.get_report(), self.request.user)
```
chris-pettinga, line 35: would just turn this into:

```suggestion
        return can_user_view_report(self.get_report(), self.request.user) and super().has_permission()
```

chris-pettinga (approved):

mikeylockwood (approved):

web/reports/views.py
MattHolmes123, line 56: I see what you mean about not checking in the templates 👍

MattHolmes123: Kind of misleading doing this though as `get_queryset` doesn't return a queryset, it returns a list (that has already evaluated the queryset

web/reports/views.py
mikeylockwood, line 35: Just because everyone else commented
```
return super().has_permission() and can_user_view_report(self.get_report(), self.request.user)
```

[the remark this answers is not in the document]
MattHolmes123: Exposing `can_user_view_report` as a jinja function would have meant you didn't need to do this and the reports only get iterated once.

web/reports/views.py
MattHolmes123: Exposing can_user_view_report as a jinja function would have meant you could easily check report access in templates (useful for later) and the list of reports would only be iterated once.

mikeylockwood: I was thinking about this when I looked at it, could store the page permission on the `Report` model as it is one per report, and the filter on it in `get_queryset`

MattHolmes123: Spoke to Marcus about this, `report_type` is already in the db so using `user.get_all_permissions()` could be used to generate a list of report_type's the user should be able to access.
That can then be easily used to filter the Report model and then `get_queryset` will return a QuerySet instance

web/reports/views.py
```diff
@@ -41,6 +47,9 @@ def get_context_data(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
         context["page_title"] = "Reports"
         return context
 
+    def get_queryset(self) -> QuerySet:
```
MattHolmes123, line 50: QuerySet[Report]

web/permissions/service.py
```diff
@@ -442,3 +452,40 @@ def filter_users_with_org_access(org: ORGANISATION, users: QuerySet[User]) -> Qu
...
+        case _:
+            raise ValueError(f"Unknown Report Type {report.report_type}")
+
+
+def can_user_view_report(user: User, report: "Report") -> bool:
+    permission = get_report_permission(report)
+    return user.has_perm(permission)
+
+
+def get_report_type_for_permission(perm: PagePermissions | SysPerms) -> ReportType | None:
```
MattHolmes123, line 478: Wherever you've done PagePermissions, SysPermission we should probably just use the actual type
right `tuple[str, str]`?

MattHolmes123: I don't mind whatever you do 🤸

PR #1452, opened 2024-04-11
ICMSLST-2584 Add Product template download / upload spreadsheet views.

MattHolmes123:


web/domains/cat/views.py
```diff
@@ -770,6 +782,74 @@ def post(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRespon
...
+        products_file = form.cleaned_data["file"]
+
+        try:
+            product_count = process_products_file(products_file, schedule_template)
+            messages.success(
+                self.request, f"Upload Success: {product_count} products uploaded successfully"
+            )
+
+        except CustomError as custom_error:
+            messages.warning(self.request, f"Upload failed: {custom_error}")
```
chris-pettinga, line 818: can this not be handled as a ValidationError on the form?

web/domains/cat/views.py
```diff
@@ -770,6 +782,74 @@ def post(self, request: AuthenticatedHttpRequest, *args, **kwargs) -> HttpRespon
...
+        finally:
+            products_file = self.request.FILES.get("file")
+
+            if products_file:
+                delete_file_from_s3(products_file.name)
+
+        return super().form_valid(form)
+
+    def form_invalid(self, form: ProductsFileUploadForm) -> HttpResponseRedirect:
+        if form.errors and "file" in form.errors:
```
chris-pettinga, line 833: same as above

web/tests/domains/case/export/test_utils.py
```diff
@@ -56,27 +57,33 @@ def create_dummy_config(is_biocidal: bool = False) -> XlsxSheetConfig:
...
-    legislation.save()
+    legislation = create_legislation(is_biocidal)
 
     schedule = CFSSchedule.objects.create(application=app, created_by=app.last_updated_by)
     schedule.legislations.add(legislation)
 
     return schedule
 
 
+def create_legislation(is_biocidal: bool) -> ProductLegislation:
```
chris-pettinga, line 76: this feels like a fixture to me

web/domains/cat/views.py
MattHolmes123: Its existing code that I was reusing (copying) but sure I can refactor them all to raise a ValidationError.

MattHolmes123: The form itself just cares that it's a file, there is too much logic specific to validating the template to move it all in to the form.
I have replaced the 20+ references to CustomError with ValidationError though

web/tests/domains/case/export/test_utils.py
MattHolmes123: if it were a fixture wouldn't there be two of them? 
`legislation` & `biocidal_legislation`
There are plenty of utility functions like this in the tests that create data based on a few function parameters.

I am going to remove this actually, as well as ProductLegislationFactory, I just refactored it as part of these tests.

Just going to add a real Legislation instead.
e.g.
```python
schedule.legislations.add(
    ProductLegislation.objects.filter(is_biocidal=True).first()
)
```

PR #1459, opened 2024-04-15
ICMSLST-1280 - CoM - Application - convert Y/N questions to radio buttons.

chris-pettinga:
Before:
image

After:
image

### DIFF
--- a/web/domains/case/export/forms.py
+++ b/web/domains/case/export/forms.py
@@ -227,6 +227,8 @@ class PrepareCertManufactureFormBase(forms.ModelForm):
 
         widgets = {
             "countries": Select2MultipleWidget,
+            "is_pesticide_on_free_sale_uk": icms_widgets.YesNoRadioSelectInline,
+            "is_manufacturer": icms_widgets.YesNoRadioSelectInline,
         }
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


web/domains/case/export/forms.py
```diff
@@ -227,6 +227,10 @@ class Meta:
 
         widgets = {
             "countries": Select2MultipleWidget,
+            "is_pesticide_on_free_sale_uk": icms_widgets.RadioSelectInline(
+                choices=BooleanYesNoChoices.choices
+            ),
+            "is_manufacturer": icms_widgets.RadioSelectInline(choices=BooleanYesNoChoices.choices),
```
MattHolmes123, line 233: Do you need `BooleanYesNoChoices`?
We have a widget for this already in web/form/widgets.py:
```
# Used for boolean fields
YesNoRadioSelectInline = RadioSelectInline(choices=((True, "Yes"), (False, "No")))
```

web/models/shared.py
```diff
@@ -16,6 +16,11 @@ class YesNoChoices(TypedTextChoices):
     no = ("no", "No")
 
 
+class BooleanYesNoChoices(TypedTextChoices):
+    yes = (True, "Yes")
+    no = (False, "No")
```
MattHolmes123, line 21: I don't think we need this

web/domains/case/export/forms.py
chris-pettinga: Sorry didn't see that

MattHolmes123 (approved):

PR #1463, opened 2024-04-15
ICMSLST-1833 Data Migration - Certificate Application Templates

mikeylockwood:


data_migration/queries/export_application/cat.py
```diff
@@ -0,0 +1,98 @@
...
+
+com_template_timestamp_update = """
+UPDATE web_certificateofmanufactureapplicationtemplate
+SET last_update_datetime = dm_cat.last_updated_datetime
+FROM data_migration_certificateapplicationtemplate dm_cat
+WHERE web_certificateofmanufactureapplicationtemplate.template_id = dm_cat.id
+"""
+
+
+cfs_schedule_template_timesheet_update = """
```
MattHolmes123, line 90: timesheet ?

web/domains/cat/models.py
```diff
@@ -17,9 +17,9 @@
 
 class CertificateApplicationTemplate(models.Model):
     class SharingStatuses(TypedTextChoices):
-        PRIVATE = ("private", "Private (do not share)")
-        VIEW = ("view", "Share (view only)")
-        EDIT = ("edit", "Share (allow edit)")
+        PRIVATE = ("PRIVATE", "Private (do not share)")
+        VIEW = ("VIEW", "Share (view only)")
+        EDIT = ("EDIT", "Share (allow edit)")
```
MattHolmes123, line 23: These values are used in a few places that use the lowercase versions of these strings.
an end to end test
several templates 
a unittest.

Can we change them back to the lowercase version please so these places don't break

MattHolmes123 (dismissed):
Added a few comments.

web/domains/cat/models.py
MattHolmes123: In an ideal work all the places would use the enum values but in this instance they don't everywhere

MattHolmes123: I'm going to sumbit a PR to fix the app code, can you fix the broken end to end test as that code doesn't import any of the web code

MattHolmes123: https://github.com/uktrade/icms/pull/1465

data_migration/queries/export_application/cat.py
mikeylockwood: Timestamp. Was filling out my timesheet when I wrote this...

web/domains/cat/models.py
mikeylockwood: Merged it in a squashed commits

data_migration/queries/export_application/cat.py
MattHolmes123: haha that's anazing :D

data_migration/management/commands/config/data_counts.py
```diff
@@ -529,6 +529,21 @@
...
+    CheckQuery(
+        name="COM Application Templates",
+        query=queries.com_application_templates_count,
+        model=web.CertificateOfManufactureApplicationTemplate,
+    ),
+    CheckQuery(
+        name="GMP Application Templates",
+        query=queries.gmp_application_templates_count,
+        model=web.CertificateOfGoodManufacturingPracticeApplicationTemplate,
+    ),
```
mikeylockwood, line 546: These checks are run after the migration in prod and check the number of app templates in V1 and V2 are the same for each application type. For GMP, with no data migration written for GMP app templates so if one gets added, this check should catch it and fail.

data_migration/models/export_application/cat.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+
+class CFSTemplateLegislation(MigrationBase):
+    cfsschedule = models.ForeignKey(CFSScheduleTemplate, on_delete=models.CASCADE)
+    productlegislation = models.ForeignKey(ProductLegislation, on_delete=models.CASCADE)
+
+    @classmethod
+    def get_excludes(cls) -> list[str]:
+        return ["cfsschedule_id"]
```
marcuspp, line 127: This doesn't call super where as `CertificateOfManufactureApplicationTemplate` does. Is it required?

marcuspp (approved):

data_migration/models/export_application/cat.py
mikeylockwood: It's not required for this one as there are only 2 fields on the model and I just want to exclude one of them. The `super()` call will  automatically exclude all the `_xml` fields which is useful on the models where they are used but not needed here

PR #1466, opened 2024-04-16
ICMSLST-2614 - Store site last login date time

marcuspp:


MattHolmes123 (approved):

web/auth/backends.py
```diff
@@ -96,6 +96,14 @@ def get_user_permissions(self, user_obj, obj=None):
 
         return set()
 
+    def authenticate(self, request: HttpRequest, **kwargs: Any) -> User | None:
+        """Authenticates the user and if the user is using an exporter/importer site updates site last login date time."""
+        user = super().authenticate(request, **kwargs)
+        site = getattr(request, "site", None)
```
marcuspp, line 102: Tests pass a None for request when authenticating so this tries to get the site attribute from request.

MattHolmes123: Yep for the auth backend we use in non prod envs this looks right 👍

PR #1467, opened 2024-04-16
ICMSLST-2541 Add Download DFL Case Documents feature.

MattHolmes123:
Constabulary contacts need to view DFL case documents without logging in to ICMS.
Uploading case documents to GOV.Notify isn't acceptable for IMCS so this feature has been added to ensure all documents are still downloaded from ICMS directly.

**Changes:**
  - Add DownloadDFLCaseDocumentsFormView view.
  - Add RegenerateDFLCaseDocumentsDownloadLinkView view.
  - Update ConstabularyDeactivatedFirearmsEmail to generate new link.
  - Add unittests.

**Screenshots:**
  - Gov Notify email template:
image
  - Download Case Documents form:
caseworker_8080_case_download-dfl-case-documents_d6b940e5-89e4-49de-b329-991f527ffb85__email=matthew holmes%40digital trade gov uk constabulary=15 check_code=37117148
  - Download Case Documents form validation error (showing regen link):
caseworker_8080_case_download-dfl-case-documents_d6b940e5-89e4-49de-b329-991f527ffb85__email=matthew holmes%40digital trade gov uk constabulary=15 check_code=37117148 (1)
  - Download Case Documents form (showing licence documents):
  
caseworker_8080_case_download-dfl-case-documents_d6b940e5-89e4-49de-b329-991f527ffb85__email=matthew holmes%40digital trade gov uk constabulary=15 check_code=37117148 (2)


web/domains/case/views/views_documents.py
```diff
@@ -1,27 +1,161 @@
...
+
+    def form_invalid(self, form: DownloadDFLCaseDocumentsForm) -> HttpResponse:
+        if form.link:
+            link = form.link
+
+            if link.failure_count < ImportApplicationDownloadLink.FAILURE_LIMIT:
+                link.failure_count = F("failure_count") + 1
+                link.save()
+
+            link.refresh_from_db()
```
chris-pettinga, line 119: Don't think this is necessary as the form passed here will be _just_ retrieved from the DB. Re. the bit above, setting the attribute directly then calling `save()` means the `link` object will have the new, correct value without needing to fetch again from DB

web/domains/case/forms.py
```diff
@@ -303,3 +316,73 @@ class ReassignOwnershipExport(ReassignOwnershipBaseForm):
...
+
+    def clean(self):
+        # Load and store the link record if valid, display error if not.
+        cleaned = super().clean()
+
+        check_code = cleaned.get("check_code")
+        email = cleaned.get("email")
+        constabulary = cleaned.get("constabulary")
+
+        if not all((check_code, email, constabulary)):
```
chris-pettinga, line 356: personal preference is multiple ands is more intuitive and doesn't require another function call

```suggestion
        if not (check_code and email and constabulary):
```

web/domains/case/forms.py
```diff
@@ -303,3 +316,73 @@ class ReassignOwnershipExport(ReassignOwnershipBaseForm):
...
+            return cleaned
```
chris-pettinga, line 357: Do we not want to raise a ValidationError here? Can we also not accomplish this checking by adding `required=True` to these 3 fields

web/domains/case/forms.py
```diff
@@ -303,3 +316,73 @@ class ReassignOwnershipExport(ReassignOwnershipBaseForm):
...
+        try:
+            # Store the link using just the code
+            link_qs = ImportApplicationDownloadLink.objects.filter(code=self.code, expired=False)
+            self.link = link_qs.get()
+
+            # Check the form data matches the record by fetching the record again.
+            link_qs.get(check_code=check_code, email=email, constabulary=constabulary)
+
+        except ObjectDoesNotExist:
+            if self.link:
```
chris-pettinga, line 368: looks like you're checking how far down the try: function we were able to get, I would wrap this in 2 separate try blocks to make it a bit more intuitive to understand. I think a ModelForm is better suited to this because the View will raise a 404 if the UUID passed in the URL doesn't correspond to a `ImportApplicationDownloadLink` object which I think makes sense. Then you could just manually check equivalence of all 3 of the user-supplied values against the `self.instance`.

web/domains/case/views/views_documents.py
```diff
@@ -1,27 +1,161 @@
...
+
+    # FormView config
+    form_class = DownloadDFLCaseDocumentsForm
+    template_name = "web/domains/case/download-case-documents.html"
+    http_method_names = ["get", "post"]
+
+    licence_date_format = "%d %b %Y"
+
+    # We don't want to see the cookie banner / enable the JS
+    extra_context = {"gtm_enabled": False}
```
chris-pettinga, line 54: related to the above comment, I would make this view inherit from `ModelFormMixin` then override the `get_object` method to get the corresponding `ImportApplicationDownloadLink` object.

chris-pettinga: Then this object is automatically passed to the ModelForm and you can access it in `self.instance` rather than querying the DB there.

web/domains/case/forms.py
MattHolmes123: We do not want to raise a 404 when a UUID isn't found, because that implies if a 404 isn't raised then we are telling whoever is trying to find valid UUID that they have got one.
I intentionally made this feature work the same with and without a valid UUID to give nothing away.

Thats why the same validation error message is always raised.

web/domains/case/forms.py
MattHolmes123: Agreed, it was four fields initially so thats why I probaby used `all`, but I removed it before submitting the PR to review.

web/domains/case/views/views_documents.py
MattHolmes123: My bad! debug code!

MattHolmes123: Actually it is needed (thanks unit tests).
We are updating the failure_count in SQL (as we used an F object), therefore after saving the updated value needs fetching.

MattHolmes123: I have moved it to the if block and added a comment

web/domains/case/forms.py
MattHolmes123: Required is True by default.
`def clean` is always called (even if the form is completely empty).
https://docs.djangoproject.com/en/5.0/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

web/domains/case/forms.py
chris-pettinga: @MattHolmes123 - I understand what you mean, we don't necessarily need to raise a 404, if no object is found with that ID we can pass `instance=None` to the form and check the individual fields in their corresponding `clean_xxx()` method only if `self.instance` exists. That way we can still use a ModelForm and the checking logic is split up nicely between functions. e.g.


```
if not self.instance or self.instance.check_code != self.cleaned_data["check_code"]:
    raise ValidationError()

```

MattHolmes123: Ooops! I actually meant to try multiple try/excepts to see if that improved the clarity of the code (so will try that now).
Regarding a ModelForm we'd still need a `def clean` method as we do the date checking.
So I'd define 4 cleaning functions instead of one, not sure if changing it to model form will add any value.

MattHolmes123: Going to try the model form out now

MattHolmes123: Discussed offline... Tried and failed

marcuspp (approved):

chris-pettinga (approved):

PR #1471, opened 2024-04-18
Adding .envrc to help with dbt-platform commands

chris-pettinga:


MattHolmes123 (changes requested):
These values are defined in the icms-deploy repo and when I spoke to @acodeninja  he said we should be in that directory when doing all dbt-platform commands apart (apart from one command and I can't remember what he said) :D

chris-pettinga:
> These values are defined in the icms-deploy repo

They are indeed, but we all work in this repo so it would be nice if we could do it all from one place, rather than `cd`ing in an out

MattHolmes123:
> > These values are defined in the icms-deploy repo
> 
> They are indeed, but we all work in this repo so it would be nice if we could do it all from one place, rather than `cd`ing in an out

icms-deploy also defines the version of platform-helper we should use for deployments as well:
https://github.com/uktrade/icms-deploy/blob/main/.copilot-version

MattHolmes123:
Rather than adding this duplication, if it's benificial to you why don't you just create symlinks for the files on your machine?

PR #1480, opened 2024-04-22
ICMSLST-1646 - COM application - pesticide validation emails to be updated

chris-pettinga:
Removing the `.gsi` from all email addresses. 

Also removed the ApplicationGuidance.pdf as it was not being used anywhere


MattHolmes123 (changes requested):
Squash to initial migration

PR #1481, opened 2024-04-23
ICMSLST-2598 Return a 403 status code when ProcessError is raised.

MattHolmes123:
ProcessError is raised when a resource is accessed when it doesn't have the expected status or active task.
Returning a 403 instead of a 500 status code is more appropriate.


marcuspp (approved):
Looks good once the tests are passing.

PR #1482, opened 2024-04-23
ICMSLST-2525 - User Reactivate/Deactivate emails

marcuspp:
Screenshot 2024-04-24 at 15 51 55


web/domains/template/context.py
```diff
@@ -332,3 +334,31 @@ def __getitem__(self, item: str) -> str:
...
+        is_importer = self.user.is_importer_user
+        is_exporter = self.user.is_exporter_user
+        match item:
+            case "PLATFORM":
+                if is_exporter == is_importer:
+                    return "Import licence / export certificate"
+                if is_importer:
+                    return "Apply for an import licence"
+                else:
+                    return "Apply for an export certificate"
```
MattHolmes123, line 353: The service names are defined here:
```
class SiteName(TypedTextChoices):
    CASEWORKER = (
        "Manage import licences and export certificates",
        "Manage import licences and export certificates",
    )
    EXPORTER = ("Apply for an export certificate", "Apply for an export certificate")
    IMPORTER = ("Apply for an import licence", "Apply for an import licence")

```

web/domains/template/context.py
```diff
@@ -332,3 +334,31 @@ def __getitem__(self, item: str) -> str:
...
+    def __init__(self, user: User):
+        self.user = user
+
+    def __getitem__(self, item: str) -> str:
+        is_importer = self.user.is_importer_user
+        is_exporter = self.user.is_exporter_user
+        match item:
+            case "PLATFORM":
+                if is_exporter == is_importer:
+                    return "Import licence / export certificate"
```
MattHolmes123, line 349: You should ideally ask Sally what to call both services together.
e.g.
Apply for an import licence or export certificate

MattHolmes123, line 349: `if is_exporter == is_importer:` will also match where both are False, is this what you want?

marcuspp: It is

marcuspp: We have discussed this and we thought having Apply for an ... infront would make it too long when followed by `account deactivated`

web/domains/template/context.py
```diff
@@ -332,3 +334,31 @@ def __getitem__(self, item: str) -> str:
...
+                return settings.ILB_CONTACT_EMAIL
+            case "PLATFORM_LINK":
+                if is_exporter == is_importer:
+                    return f"{get_importer_site_domain()} for import or {get_exporter_site_domain()} for export."
+                if is_importer:
+                    return get_importer_site_domain()
+                else:
+                    return get_exporter_site_domain()
+            case _:
+                raise ValueError(f"{item} is not a valid user management template context value")
```
MattHolmes123, line 364: Alternative using only match and the SiteName labels:
```python
class UserManagementContext:
        is_importer = self.user.is_importer_user
        is_exporter = self.user.is_exporter_user

        match item, is_importer, is_exporter:
            case "PLATFORM", True, False:
                return SiteName.IMPORTER.label
            case "PLATFORM", False, True:
                return SiteName.EXPORTER.label
            case "PLATFORM", _, _:
                return "Import licence / export certificate"
            case "PLATFORM_LINK", True, False:
                return get_importer_site_domain()
            case "PLATFORM_LINK", False, True:
                return get_exporter_site_domain()
            case "PLATFORM_LINK", _, _:
                return f"{get_importer_site_domain()} for import or {get_exporter_site_domain()} for export."
            case "CASE_OFFICER_EMAIL", _, _:
                return settings.ILB_CONTACT_EMAIL
            case _, _, _:
                raise ValueError(f"{item} is not a valid user management template context value")
```

web/domains/user/models.py
```diff
@@ -51,6 +51,14 @@ def full_name(self):
     def account_last_login_date(self):
         return None if self.last_login is None else self.last_login.date()
 
+    @property
+    def is_importer_user(self) -> bool:
+        return self.importer_last_login is not None
+
+    @property
+    def is_exporter_user(self) -> bool:
+        return self.exporter_last_login is not None
```
MattHolmes123, line 60: I wonder if these will be useful elsewhere when you could just as easily use the db columns directly.
I lean towards thin models wherever possible and refactor later when required.

web/domains/user/views.py
```diff
@@ -1,26 +1,43 @@
...
     UserListFilter,
+    UserManagementEmailForm,
     UserPhoneNumberForm,
 )
 
+if TYPE_CHECKING:
+    from django.db import QuerySet
+
+DEACTIVATE = "deactivate"
+REACTIVATE = "reactivate"
```
MattHolmes123, line 39: Earlier on you reference these values:
`choices=[("deactivate", "Deactivate User"), ("reactivate", "Reactivate User")],`
I smell a TypedTextChoice coming along :)

web/domains/user/views.py
```diff
@@ -1,26 +1,43 @@
...
 from .forms import (
     UserDetailsUpdateForm,
     UserEmailForm,
     UserListFilter,
+    UserManagementEmailForm,
     UserPhoneNumberForm,
 )
 
+if TYPE_CHECKING:
+    from django.db import QuerySet
```
MattHolmes123, line 36: No need for TYPE_CHECKING this

web/domains/user/views.py
```diff
@@ -140,6 +156,9 @@ class UsersListView(ModelFilterView):
     page_title = "Maintain Web User Accounts"
     permission_required = Perms.sys.ilb_admin
 
+    def get_queryset(self) -> "QuerySet[User]":
+        return super().get_queryset().exclude(pk=0)
+
```
MattHolmes123, line 161: We should refactor this and reference a a constant instead of 0.
```
def get_anonymous_user_instance(user_model: type[User]) -> User:
    # Change the guardian anonymous user pk to match V1 guest user
    kwargs = {
        "pk": 0,
        user_model.USERNAME_FIELD: guardian_settings.ANONYMOUS_USER_NAME,
    }
    user = user_model(**kwargs)
    user.set_unusable_password()

    return user

```

MattHolmes123: `ANONYMOUS_USER_PK` ?

web/domains/user/views.py
```diff
@@ -154,17 +173,70 @@ class Display:
...
     template_name = "web/domains/user/detail.html"
+    form_class = UserManagementEmailForm
+
+    def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
+        context = super().get_context_data()
+        user = self.get_platform_user()
+        return {
+            "platform_user": user,
+            "action": DEACTIVATE if user.is_active else REACTIVATE,
+        } | context
```
MattHolmes123, line 191: Elsewhere we do this:
```
return context | {"extra": "context}
```

web/domains/template/context.py
MattHolmes123: It's the service name at the end of the day.

web/domains/user/views.py
```diff
@@ -154,17 +173,70 @@ class Display:
...
             "account_last_login_date": {"header": "Last Login Date"},
         }
-        actions = [actions.DeactivateUser(), actions.ActivateUser()]
         select = True
 
 
-class UserDetailView(PermissionRequiredMixin, DetailView):
-    # PermissionRequiredMixin config
+@method_decorator(transaction.atomic, name="post")
+class UserDetailView(PermissionRequiredMixin, FormView):
```
MattHolmes123, line 180: I'm not sure we've done this quite right.
I think the DetailView should remain readonly and we have two new views.
`class ActiveUserFormView`
`class DeactivateUserFormView`
Then the deactive / reactive links would instead take you to the new forms instead of being post only actions

MattHolmes123: Would result in cleaner views that only do one thing, I imagine there could be a base class for both new views.

MattHolmes123 (changes requested):
I think we should split the views

web/domains/template/context.py
MattHolmes123: @marcuspp This fixes  mypy:
```
class UserManagementContext:
    def __init__(self, user: User) -> None:
        self.user = user

        is_importer = self.user.is_importer_user
        is_exporter = self.user.is_exporter_user

        match item, is_importer, is_exporter:
            case "PLATFORM", True, False:
                return SiteName.IMPORTER.label
            case "PLATFORM", False, True:
                return SiteName.EXPORTER.label
            case "PLATFORM", _, _:
                return PLATFORM_NAME
            case "PLATFORM_LINK", True, False:
                return get_importer_site_domain()
            case "PLATFORM_LINK", False, True:
                return get_exporter_site_domain()
            case "PLATFORM_LINK", _, _:
                return f"{get_importer_site_domain()} for import or {get_exporter_site_domain()} for export."
            case "CASE_OFFICER_EMAIL", _, _:
                return settings.ILB_CONTACT_EMAIL

        raise ValueError(f"{item} is not a valid user management template context value")
```

web/domains/user/actions.py
```diff
@@ -1,40 +1,25 @@
...
 
-from web.views.actions import PostAction
+from web.views.actions import LinkAction
 
-from .models import User
 
-
-class ActivateUser(PostAction):
-    action = "activate"
+class ActivateUser(LinkAction):
```
MattHolmes123, line 6: Can we add types for all methods of `ActivateUser` and `DeactivateUser`

web/domains/user/forms.py
```diff
@@ -139,3 +146,12 @@ def save(self, commit=True):
...
         return email
+
+
+class UserManagementEmailForm(forms.Form):
+    subject = forms.CharField(max_length=100, label="Subject")
+    body = forms.CharField(widget=forms.Textarea, label="Body")
+    send_email = forms.BooleanField(required=False)
+
+    class Meta:
+        fields = ["subject", "body", "send_email", "action"]
```
MattHolmes123, line 157: `"action"`doesn't have a field defined, that right?

MattHolmes123: Actually... its a Form not a model form so we can remove this:
```
    class Meta:
        fields = ["subject", "body", "send_email", "action"]
```

web/domains/user/views.py
```diff
@@ -140,31 +159,114 @@ class UsersListView(ModelFilterView):
     page_title = "Maintain Web User Accounts"
     permission_required = Perms.sys.ilb_admin
 
+    def get_queryset(self) -> "QuerySet[User]":
```
MattHolmes123, line 162: `"QuerySet[User]"` -> QuerySet[User]

web/templates/web/domains/user/detail.html
```diff
@@ -1,65 +1,87 @@
...
+          <td>{{ number.get_type_display() }}</td>
+          <td>{{ number.comment or "" }}</td>
         </tr>
       {% endfor %}
       </tbody>
     </table>
-  {% endcall %}
-{% endblock %}
+    {% endcall %}
+{% endblock %}
```
MattHolmes123, line 87: Missing newline

web/templates/web/domains/user/user_management.html
```diff
@@ -0,0 +1,46 @@
...
+                  <a href="{{ icms_url('users-list') }}" class="button link-button">Cancel</a>
+                </li>
+              </ul>
+            </div>
+            <div class="one columns"></div>
+          </div>
+        </div>
+  {% endcall %}
+    {% endcall %}
+{% endblock  %}
```
MattHolmes123, line 46: Same

web/domains/user/views.py
```diff
@@ -140,31 +159,114 @@ class UsersListView(ModelFilterView):
...
     model = User
     pk_url_kwarg = "user_pk"
-    context_object_name = "user"
+    context_object_name = "platform_user"
     http_method_names = ["get"]
-    template_name = "web/domains/user/detail.html"
+
+    def get_object(self) -> User:
+        pk = self.kwargs["user_pk"]
+        return get_object_or_404(User, ~Q(is_superuser=True, pk=ANONYMOUS_USER_PK), pk=pk)
```
MattHolmes123, line 195: The user list view and the two places you do this you could potentially use a reusable function to get you most of the way:
```
def user_list_view_qs() -> QuerySet[User]:
    """Reusable QuerySet that can be used in several views."""
    return User.objects.exclude(pk=ANONYMOUS_USER_PK).exclude(is_superuser=True)
```

web/domains/user/views.py
```diff
@@ -140,31 +159,114 @@ class UsersListView(ModelFilterView):
...
+            "page_title": f"Reactivate {platform_user.full_name}'s account",
+            "platform_user": platform_user,
+            "action": "Reactivate user",
+        }
+
+    def get_platform_user(self) -> User:
+        pk = self.kwargs["user_pk"]
+        return get_object_or_404(
+            User, ~Q(is_superuser=True, pk__in=[ANONYMOUS_USER_PK, self.request.user.pk]), pk=pk
+        )
```
MattHolmes123, line 218: Also SingleObjectMixin supports `queryset`, so if you went with that reusable function above you wouldn't need to reimplement what SingleObjectMixin is good at.

MattHolmes123 (approved):

PR #1490, opened 2024-04-25
ICMSLST-1584 - Manage Case - make response box automatically available once email sent.

chris-pettinga:
Before:
image

After:
image

### DIFF
--- a/web/templates/web/domains/case/manage/case-email-box.html
+++ b/web/templates/web/domains/case/manage/case-email-box.html
@@ -23,12 +23,6 @@
           </form>
         </li>
       </ul>
-    {% else %}
-      <a
-        class="button small-button icon-pencil"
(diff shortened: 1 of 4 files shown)


web/templates/web/domains/case/manage/case-email-box.html
```diff
@@ -196,5 +190,26 @@
...
+                  type="submit"
+                  value="Close"
+                  class="primary-button button" />
+            </li>
+          </ul>
+        </div>
+        <div class="three columns">
+        </div>
+      </div>
+    {% endcall %}
```
MattHolmes123, line 213: Can you fix a related bug while you're in the area please.
Do the following, 
1. Create a cfs app,
2. send a few emails and record responses
3. approve the app
4. Click "Start Authorisation"
5. Click the "View Case" link in the workbasket

Navigate back to the HSE Emails tab and you'll see the "Edit Response" Link is still available.
All caseworker screens can be viewed and they all have access to a `readonly_view` context variable.
Can you check that here please.

PR #1493, opened 2024-04-26
ICMSLST-1612 - Applicant Search - Firearms applications - provide action to "Provide Report".

chris-pettinga:


web/permissions/context_processors.py
```diff
@@ -3,30 +3,32 @@
 from django.http import HttpRequest
 from guardian.core import ObjectPermissionChecker
 
+from ..domains.case.types import ImpOrExp
```
MattHolmes123, line 6: Proper import please

web/tests/helpers.py
```diff
@@ -583,6 +583,11 @@ def fa_create_report(application_pk: int):
         kwargs = {"application_pk": application_pk}
         return reverse("import:fa:create-report", kwargs=kwargs)
 
+    @staticmethod
+    def fa_provide_report(application_pk: int):
```
MattHolmes123, line 587: `-> str`

web/utils/search/actions.py
```diff
@@ -11,6 +11,7 @@
 from web.models import ExportApplication, ImportApplication
 from web.permissions import ExporterObjectPermissions, ImporterObjectPermissions
 
+from ...permissions.context_processors import UserObjectPerms
```
MattHolmes123, line 14: Proper imports

MattHolmes123: Actually you won't need UserObjectPerms (that class is used in templates)

web/utils/search/actions.py
```diff
@@ -95,6 +101,35 @@ def get_action(app: ImportApplication) -> types.SearchAction:
...
 
 
+#
+# Applicant action (Import)
+#
+class ProvideSupplementaryReportAction(ActionProtocol):
+    @staticmethod
+    def show_action(app: ImportApplication, uop: UserOrganisationPermissions) -> bool:
+        user_object_perms = UserObjectPerms(uop.user)
+        can_edit_application = user_object_perms.can_edit_application(app)
```
MattHolmes123, line 111: Loading this every time for every application in the search screen will be really inefficient.

There is already a function in this file that you can use to do the same thing: `_can_edit_org_or_agent`.
The difference is it uses `uop` which only gets initialised once and used for every record in the search results.

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #1509, opened 2024-05-02
ICMSLST-1822 - All applications - Further Information request the from ILB cases prompts decision action to release ownership

chris-pettinga:
After:
image

### DIFF
--- a/web/domains/case/fir/forms.py
+++ b/web/domains/case/fir/forms.py
@@ -1,22 +1,34 @@
 import logging
 
-from django.forms import ChoiceField, ModelForm, Textarea
+from django import forms
 
 from web.models import FurtherInformationRequest
 
 logger = logging.getLogger(__name__)
 
 
-class FurtherInformationRequestForm(ModelForm):
-    status = ChoiceField(choices=FurtherInformationRequest.STATUSES, disabled=True)
+class FurtherInformationRequestForm(forms.ModelForm):
+    status = forms.ChoiceField(choices=FurtherInformationRequest.STATUSES, disabled=True)
+    release_ownership = forms.BooleanField(required=False, initial=True, widget=forms.HiddenInput)
+    send = forms.BooleanField(required=False, initial=False, widget=forms.HiddenInput)
 
     class Meta:
         model = FurtherInformationRequest
         fields = ("status", "request_subject", "request_detail")
 
+    @property
+    def media(self):
+        if self.instance.accessrequest_set.exists():
+            js = ["web/js/pages/edit-fir-access-request.js"]
+        else:
+            js = ["web/js/pages/edit-fir.js"]
 
-class FurtherInformationRequestResponseForm(ModelForm):
+        media = forms.Media(js=js)
+        return media
+
+
+class FurtherInformationRequestResponseForm(forms.ModelForm):
     class Meta:
         model = FurtherInformationRequest
         fields = ("response_detail",)
-        widgets = {"response_detail": Textarea({"rows": 5})}
+        widgets = {"response_detail": forms.Textarea({"rows": 5})}
--- a/web/domains/case/views/utils.py
+++ b/web/domains/case/views/utils.py
@@ -1,7 +1,12 @@
 from typing import TYPE_CHECKING
 
 from web.domains.case.services import case_progress
-from web.domains.case.types import ImpOrExpOrAccess, ImpOrExpOrAccessT, ImpOrExpT
+from web.domains.case.types import (
+    ImpOrExp,
+    ImpOrExpOrAccess,
+    ImpOrExpOrAccessT,
+    ImpOrExpT,
+)
 from web.flow.errors import ProcessError
 from web.models import AccessRequest, ExportApplication, ImportApplication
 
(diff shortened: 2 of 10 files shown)


web/domains/case/views/views_fir.py
```diff
@@ -31,7 +31,12 @@
 from web.types import AuthenticatedHttpRequest
 from web.utils.s3 import get_file_from_s3
 
-from .utils import get_caseworker_view_readonly_status, get_class_imp_or_exp_or_access
+from ..access.models import AccessRequest
```
MattHolmes123, line 34: Full import path please if you are doing `..`

MattHolmes123: Actually, this is a model import so `from web.models import AccessRequest`

web/domains/case/views/utils.py
```diff
@@ -50,3 +55,16 @@ def get_caseworker_view_readonly_status(
         is_readonly = True
 
     return is_readonly
+
+
+def release_ownership_of_application(application: ImpOrExp, model_class: ImpOrExpT) -> ImpOrExp:
+    """Release ownership of the application."""
+
+    if application.status != model_class.Statuses.VARIATION_REQUESTED:
+        application.status = model_class.Statuses.SUBMITTED
```
mikeylockwood, line 64: Are the only status's this would be in `PROCESSING` and `VARIATION_REQUESTED`. Prefer to check the `True` condition rather than the `not False` condition
```
if application.status != model_class.Statuses.PROCESSING:
```

web/domains/case/views/utils.py
```diff
@@ -50,3 +55,16 @@ def get_caseworker_view_readonly_status(
         is_readonly = True
 
     return is_readonly
+
+
+def release_ownership_of_application(application: ImpOrExp, model_class: ImpOrExpT) -> ImpOrExp:
```
mikeylockwood, line 60: Is this just for applications or access requests too?

mikeylockwood: I can see below access requests are explicitly excluded

web/templates/web/domains/case/manage/edit-fir.html
```diff
@@ -11,4 +11,50 @@
...
+        </ul>
+      </div>
+    </div>
+  </div>
 {% endblock %}
+
+
+{% block page_js %}
+  {{ super() }}
+  <script type="text/javascript">
```
mikeylockwood, line 36: Didn't you add `nonce="{{ request.csp_nonce }}"` attributes for inline js before? Would prefer this block in a js file

web/templates/web/domains/case/manage/edit-fir.html
```diff
@@ -11,4 +11,50 @@
...
+
+    document.getElementById('save-fir-request-button').addEventListener("click", function (event) {
+      document.querySelector('input[name=send]').value = "False"
```
mikeylockwood, line 39: End js lines with `;`

mikeylockwood (approved):

web/templates/web/domains/case/manage/edit-fir.html
```diff
@@ -11,4 +11,50 @@
...
+            <li>
+              <button class="link-button" type="button" id="retain-ownership-button">No</button>
+            </li>
+            <li>
+              <button class="link-button" type="button" id="release-ownership-button">Yes</button>
+            </li>
+        </ul>
+      </div>
+    </div>
+  </div>
```
MattHolmes123, line 30: Why does this display when the further information request is for an Access Reqest:
image
It shouldn't be shown

web/domains/case/views/utils.py
MattHolmes123: This code is called earlier Michael:

```
def application_in_processing(application: ImpOrExp) -> None:
    """Check if an application is being processed by a caseworker."""

    expected_status = [ST.SUBMITTED, ST.PROCESSING, ST.VARIATION_REQUESTED]
    expected_task = TT.PROCESS

    check_expected_status(application, expected_status)
    check_expected_task(application, expected_task)
```    

I'd be happy to leave this function as is as its just been extracted from the release_ownership view.

web/templates/partial/case/edit-fir.html
```diff
@@ -90,3 +91,6 @@
   </div>
 </div>
 </fieldset>
+
+
+
```
MattHolmes123, line 96: Only have a single extra line

MattHolmes123 (changes requested):
popup shouldn't display for access requests.

web/domains/case/fir/forms.py
```diff
@@ -1,6 +1,13 @@
...
 
-from django.forms import ChoiceField, ModelForm, Textarea
+from django.forms import (
+    BooleanField,
+    ChoiceField,
+    HiddenInput,
+    Media,
+    ModelForm,
+    Textarea,
+)
```
MattHolmes123, line 10: We import as `from django import forms` more frequently than we do `from django.forms import foo, bar, baz`.
Would you mind doing it the first way (should reduce the overall line length).

web/domains/case/fir/forms.py
```diff
@@ -9,11 +16,23 @@
...
         model = FurtherInformationRequest
         fields = ("status", "request_subject", "request_detail")
 
+    @property
+    def media(self):
+        if self.instance.accessrequest_set.exists():
+            js = ("web/js/pages/edit-fir-access-request.js",)
+        else:
+            js = ("web/js/pages/edit-fir.js",)
+
```
MattHolmes123, line 32: All the django docs define js / css as lists of strings not tuples.
Not sure if it would ever cause issues but at least it would remove the need for the trailing `,`

MattHolmes123 (approved):

web/domains/case/fir/forms.py
chris-pettinga: good spot will do, this is PyCharm doing me dirty

PR #1520, opened 2024-05-09
ICMSLST-2403 - Re-instate the Quick Issue button

chris-pettinga:
Quick-issue button to automatically approve and authorise documents for an accepted application.

Not approved:
image

Approved:
image

### DIFF
--- a/web/domains/case/views/views_misc.py
+++ b/web/domains/case/views/views_misc.py
@@ -41,6 +41,7 @@ from web.permissions import (
     organisation_get_contacts,
 )
 from web.types import AuthenticatedHttpRequest
+from web.utils.lock_manager import LockManager
 from web.utils.pdf.utils import cfs_cover_letter_key_filename
 from web.utils.s3 import delete_file_from_s3, get_s3_client
 from web.utils.validation import ApplicationErrors
@@ -480,61 +481,7 @@ def start_authorisation(
         application_errors: ApplicationErrors = get_app_errors(application, case_type)
 
         if request.method == "POST" and not application_errors.has_errors():
-            task = case_progress.get_expected_task(application, Task.TaskType.PROCESS)
-
-            create_documents = True
-            send_vr_email = False
-
-            if application.status == application.Statuses.VARIATION_REQUESTED:
-                if (
-                    application.is_import_application()
-                    and application.variation_decision == application.REFUSE
-                ):
-                    vr = application.variation_requests.get(status=VariationRequest.Statuses.OPEN)
-                    next_task = None
-                    application.status = model_class.Statuses.COMPLETED
-                    vr.status = VariationRequest.Statuses.REJECTED
-                    vr.reject_cancellation_reason = application.variation_refuse_reason
-                    vr.closed_datetime = timezone.now()
-                    vr.save()
-                    send_vr_email = True
-                    create_documents = False
-                else:
-                    next_task = Task.TaskType.AUTHORISE
-
-            else:
-                if application.decision == application.REFUSE:
-                    next_task = Task.TaskType.REJECTED
-                    application.status = model_class.Statuses.COMPLETED
-                    create_documents = False
-
-                else:
-                    next_task = Task.TaskType.AUTHORISE
-                    application.status = model_class.Statuses.PROCESSING
-
-            application.update_order_datetime()
-            application.save()
-
-            end_process_task(task)
-
-            if next_task:
-                Task.objects.create(process=application, task_type=next_task, previous=task)
-
-            if create_documents:
-                document_pack.doc_ref_documents_create(application, request.icms.lock_manager)
-            else:
-                document_pack.pack_draft_archive(application)
-
-            if (
-                application.decision == application.REFUSE
-                and application.status == model_class.Statuses.COMPLETED
-            ):
-                send_application_refused_email(application)
-
-            if send_vr_email:
-                send_variation_request_email(
-                    vr, EmailTypes.APPLICATION_VARIATION_REQUEST_REFUSED, application
-                )
+            start_application_authorisation(application, request.icms.lock_manager)
             return redirect(reverse("workbasket"))
 
         else:
@@ -565,20 +512,9 @@ def authorise_documents(
         )
 
         case_progress.application_is_authorised(application)
-        task = case_progress.get_expected_task(application, Task.TaskType.AUTHORISE)
 
         if request.method == "POST":
-            end_process_task(task, request.user)
-
-            Task.objects.create(
-                process=application, task_type=Task.TaskType.DOCUMENT_SIGNING, previous=task
-            )
-
-            application.update_order_datetime()
-            application.save()
-
-            # Queues all documents to be created
-            create_case_document_pack(application, request.user)
+            authorise_application_documents(application, request.user)
 
             messages.success(
                 request,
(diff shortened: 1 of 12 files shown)


web/domains/case/utils.py
```diff
@@ -9,13 +9,17 @@
...
+from web.domains.case.services import case_progress, document_pack, reference
 from web.flow.models import ProcessTypes
 from web.models import ExportApplication, ImportApplication, Task, User
 from web.permissions import AppChecker
 from web.types import AuthenticatedHttpRequest
 from web.utils.s3 import get_file_from_s3
 
+from ...mail.constants import EmailTypes
+from ...mail.emails import send_application_refused_email, send_variation_request_email
+from ...utils.lock_manager import LockManager
```
MattHolmes123, line 21: No relative imports above the current directory.

MattHolmes123 (changes requested):
Haven't reviewed this yet but can see this is only partly done.

I thought quick issue does the following:
1. The "start authorisation" step
2. The "Authorise documents" step.

So the outcome for an import application would be that it is with chief (or bypass chief when not enabled).
And the outcome for an export application would be that it is complete.

You'd have to check V1 though.

MattHolmes123:
This banner is missing from the checklist view (not caused by you just noticed as part of this pr)
image
image

web/domains/case/utils.py
```diff
@@ -165,3 +169,67 @@ def application_history(app_reference: str, is_import: bool = True) -> None:
 
     for p in document_pack._get_qm(app).order_by("created_at"):
         print(f"DocumentPack: {p}")
+
+
+def start_application_authorisation(application: ImpOrExp, lock_manager: "LockManager") -> None:
```
MattHolmes123, line 174: No need for quotes.

web/domains/case/views/views_misc.py
```diff
@@ -958,3 +905,30 @@ def _get_copy_recipients(application: ImpOrExp) -> "QuerySet[User]":
...
     else:
         return User.objects.none()
+
+
+class QuickIssueApplication(
+    ApplicationTaskMixin, LoginRequiredMixin, PermissionRequiredMixin, View
+):
+    http_method_names = ["post"]
+    current_status = [ImpExpStatus.PROCESSING]
+    current_task_type = Task.TaskType.PROCESS
```
MattHolmes123, line 915: This doesn't reflect what is checked in the function based view for start authorisation:
`case_progress.application_in_processing(application)`
It won't work for variation requests.

MattHolmes123: When clicking quick issue for an application with a variation request:
image

web/domains/case/views/views_misc.py
```diff
@@ -958,3 +905,30 @@ def _get_copy_recipients(application: ImpOrExp) -> "QuerySet[User]":
...
+        if application_errors.has_errors():
+            return redirect(
+                reverse(
+                    "case:start-authorisation",
+                    kwargs={"application_pk": self.application.pk, "case_type": case_type},
+                )
+            )
+        else:
+            start_application_authorisation(self.application, request.icms.lock_manager)
+            return redirect(reverse("workbasket"))
```
MattHolmes123, line 934: I don't thin you want the whole method to be atomic, we are doing two separate units of work.
We should consider several blocks and let it get as far as it can. 
e.g. the authorise step works but fails on the document signing step.


Something like this
```
    def post(
        self, request: AuthenticatedHttpRequest, *args: Any, case_type: str, **kwargs: Any
    ) -> HttpResponse:
        self.set_application_and_task()

        #
        # Start authorisation step
        #
        with transaction.atomic():
            application_errors: ApplicationErrors = get_app_errors(self.application, case_type)
            if application_errors.has_errors():
                return redirect(
                    reverse(
                        "case:start-authorisation",
                        kwargs={"application_pk": self.application.pk, "case_type": case_type},
                    )
                )
            else:
                start_application_authorisation(self.application, request.icms.lock_manager)
        
        self.application.refresh_from_db()
        # Just did this so I didn't need to change the authorise documents code (as it's an example)
        application = self.application
        
        #
        # Authorise documents step
        #
        with transaction.atomic():
            # Use the authorise documents logic here (including the status/task checking)

            case_progress.application_is_authorised(application)
            task = case_progress.get_expected_task(application, Task.TaskType.AUTHORISE)

            if request.method == "POST":
                end_process_task(task, request.user)

                Task.objects.create(
                    process=application, task_type=Task.TaskType.DOCUMENT_SIGNING, previous=task
                )

                application.update_order_datetime()
                application.save()

                # Queues all documents to be created
                create_case_document_pack(application, request.user)

                messages.success(
                    request,
                    f"Authorise Success: Application {application.reference} has been queued for document signing",
                )

```

MattHolmes123: Also instead of doing this
```
        self.application.refresh_from_db()
        # Just did this so I didn't need to change the authorise documents code (as it's an example)
        application = self.application
```
You probably want to correctly fetch it from the db and select_for_update for the authorise documents step.

web/domains/case/utils.py
```diff
@@ -165,3 +169,88 @@ def application_history(app_reference: str, is_import: bool = True) -> None:
...
+        )
+
+
+def authorise_application_documents(application: ImpOrExp, user: User) -> None:
+    """Authorise the documents for the application.
+
+    :param application: The application to authorise
+    :param user: The user authorising the documents
+    """
+    from .tasks import create_case_document_pack
```
MattHolmes123, line 244: Why the need for this?

web/domains/case/views/views_misc.py
```diff
@@ -958,3 +895,54 @@ def _get_copy_recipients(application: ImpOrExp) -> "QuerySet[User]":
...
+    http_method_names = ["post"]
+    current_status = [
+        ImpExpStatus.SUBMITTED,
+        ImpExpStatus.PROCESSING,
+        ImpExpStatus.VARIATION_REQUESTED,
+    ]
+    current_task_type = Task.TaskType.PROCESS
+    permission_required = [Perms.sys.ilb_admin]
+
+    @transaction.atomic()
```
MattHolmes123, line 912: Remove this

web/templates/web/domains/case/manage/base.html
```diff
@@ -41,6 +41,13 @@
         <input type="submit" class="button" value="Release Ownership"/>
       </form>
       <a href="{{ icms_url('case:reassign-ownership', kwargs={'application_pk': process.pk, 'case_type': case_type}) }}" class="button">Reassign</a>
+      <form
+        method="post"
+        action="{{ icms_url('case:quick-issue', kwargs={'application_pk': process.pk, 'case_type': case_type}) }}"
+        class="form-inline">
+        {{ csrf_input }}
+        <input type="submit" class="button" value="Quick Issue"/>
+      </form>
```
MattHolmes123, line 50: This PR throws an exception when you click quick when you refuse an application / variation request.

If you check V1 it will show that the Quick Issue feature only works when you are approving an application / variation request.


image

web/tests/conftest.py
```diff
@@ -1543,3 +1546,13 @@ def gmp_cat(exporter_one_contact) -> CertificateApplicationTemplate:
...
+
+
+@pytest.fixture()
+def dummy_signature_image():
+    """Generate a dummy signature image for testing"""
+    image = PIL.Image.new("RGBA", size=(50, 50), color=(256, 0, 0))
+    image_file = io.BytesIO()
+    image.save(image_file, "PNG")
+    image_file.seek(0)
+    return PngImageFile(image_file)
```
MattHolmes123, line 1558: A duplicate of this fixture exists in `icms/web/tests/utils/pdf/test_signer.py`

MattHolmes123: You can blame me for adding the fixture in `web/tests/domains/case/test_tasks.py` as I should have created a central fixture then.

web/domains/case/utils.py
MattHolmes123: I don't like to leave circular dependencies if I can help it.
Both functions you've added are only used in `views_misc.py`.

If you move them both back there the circular dependency issue should resolve itself.

MattHolmes123 (changes requested):
A quick test shows this throws an exception when you either refuse an application or variation request.
Checking V1 also shows that the feature should only be enabled when an application or variation request are going to be approved.

web/domains/case/views/views_misc.py
chris-pettinga: @MattHolmes123 - how else are we going to protect the view?

MattHolmes123: We have two atomic blocks in the view (one per step)
image

We don't need the outer one.

MattHolmes123: They should be treated as two seperate units of work and I'd rather it commit after the first block.

chris-pettinga: @MattHolmes123 sorry apologies I thought you were talking about the 
```
current_task_type = Task.TaskType.PROCESS
permission_required = [Perms.sys.ilb_admin]
```

block, have removed transaction.etomic

MattHolmes123 (changes requested):
Still fails when you try to quick issue a refused variation request.
image

chris-pettinga:
@MattHolmes123 - have added some conditionals to check if it's a variation request, ready for re-review

chris-pettinga:
@MattHolmes123 ignore me, need to also take into account Export applications.

web/templates/web/domains/case/manage/base.html
```diff
@@ -46,10 +46,10 @@
         action="{{ icms_url('case:quick-issue', kwargs={'application_pk': process.pk, 'case_type': case_type}) }}"
         class="form-inline">
         {{ csrf_input }}
-        {% if (process.status == "VARIATION_REQUESTED" and process.variation_decision != "APPROVE") or (process.status != "VARIATION_REQUESTED" and process.decision != "APPROVE")  %}
-          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
-        {% else %}
+        {% if process.can_quick_issue()  %}
           <input type="submit" class="button" value="Quick Issue"/>
+        {% else %}
+          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
```
MattHolmes123, line 52:
```html
        {% if (process.status == "PROCESSING" and process.decision == "APPROVE") or (process.status == "VARIATION_REQUESTED" and process.variation_decision == "APPROVE") %}
          <input type="submit" class="button" value="Quick Issue"/>
        {% else %}
          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
        {% endif %}
```

web/domains/case/_import/models.py
```diff
@@ -357,6 +357,18 @@ def get_status_display(self) -> str:
             return f"{status} (Refused)"
         return status
 
+    def can_quick_issue(self) -> bool:
```
MattHolmes123, line 360: We don't need these model methods, the logic is the same for import and export applications.

web/templates/web/domains/case/manage/base.html
```diff
@@ -41,10 +41,26 @@
...
+      <form
+        method="post"
+        action="{{ icms_url('case:quick-issue', kwargs={'application_pk': process.pk, 'case_type': case_type}) }}"
+        class="form-inline">
+        {{ csrf_input }}
+        {% if process.can_quick_issue()  %}
+          <input type="submit" class="button" value="Quick Issue"/>
+        {% else %}
+          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
+        {% endif %}
```
MattHolmes123, line 53:
```
        {% if (process.status == "PROCESSING" and process.decision == "APPROVE") or (process.status == "VARIATION_REQUESTED" and process.variation_decision == "APPROVE") %}
          <input type="submit" class="button" value="Quick Issue"/>
        {% else %}
          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
        {% endif %}
```

web/domains/case/views/views_misc.py
```diff
@@ -958,3 +894,151 @@ def _get_copy_recipients(application: ImpOrExp) -> "QuerySet[User]":
...
     else:
         return User.objects.none()
+
+
+class QuickIssueApplicationView(
+    ApplicationTaskMixin, LoginRequiredMixin, PermissionRequiredMixin, View
+):
+    http_method_names = ["post"]
+    current_status = [
+        ImpExpStatus.SUBMITTED,
```
MattHolmes123, line 904: Remove SUBMITTED as it's not required for this view.
The only reason SUBMITTED is checked in other caseworker views is because caseworkers can review readonly cases that have not been picked up by a caseworker yet.

web/templates/web/domains/case/manage/base.html
```diff
@@ -41,10 +41,26 @@
...
 {% endblock %}
 
 {% block sidebar %}
   {% include "partial/case/" + case_type + "/sidebar-management.html" %}
 {% endblock %}
+
+{% block page_js %}
+  {{ super() }}
+  
+{% endblock %}
```
MattHolmes123, line 66: missing newline

web/domains/case/views/views_misc.py
MattHolmes123: Fixed in my pr:
https://github.com/uktrade/icms/pull/1545

MattHolmes123 (changes requested):
Added a few comments but mostly just fixed them here for you to review and test:
https://github.com/uktrade/icms/pull/1545

web/domains/case/views/views_misc.py
```diff
@@ -956,3 +892,153 @@ def _get_copy_recipients(application: ImpOrExp) -> "QuerySet[User]":
...
+            app_approved = (
+                self.application.status == ImpExpStatus.PROCESSING
+                and self.application.decision == self.application.APPROVE
+            )
+
+            # There is no variation_decision field on export applications.
+            if self.application.is_import_application():
+                variation_approved = (
+                    self.application.status == ImpExpStatus.VARIATION_REQUESTED
+                    and self.application.variation_decision == self.application.REFUSE
```
MattHolmes123, line 924: This should be `and self.application.variation_decision == self.application.APPROVE`

Can you add a test for doing a quick issue on an app with a variation request, it's an important path to test.

marcuspp (approved):

MattHolmes123 (approved):

PR #1522, opened 2024-05-10
ICMSLST-1610 Fix bug when optional telephone_number is not set.

MattHolmes123:


web/domains/constabulary/forms.py
```diff
@@ -63,8 +63,12 @@ class Meta:
         fields = ["name", "region", "email", "telephone_number"]
         widgets = {"region": Select(choices=Constabulary.REGIONS)}
 
-    def clean_telephone_number(self):
+    def clean_telephone_number(self) -> int | None:
         telephone_number = self.cleaned_data["telephone_number"]
+
+        if not telephone_number:
+            return
```
mikeylockwood, line 70: `return None` for mypy

MattHolmes123: Ta

PR #1525, opened 2024-05-13
ICMSLST-2380 - Chunk file upload for large files

marcuspp:


chris-pettinga (approved):
Looks good thank you.

I think what would be super useful is a test ensuring that the `_upload_file_parts_to_s3` function calls S3 with a list of bytes that when put together again, hash to the same value as the entire file as a whole.

mikeylockwood (approved):

PR #1526, opened 2024-05-14
CSP fixes

chris-pettinga:
Adding some google tag manager URLs to the CSP script-src as they keep on getting blocked.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -453,9 +453,11 @@ CSP_DEFAULT_SRC = ("'self'",)
 # JS tags with a src attribute can only be loaded from ICMS itself or the DBT Sentry instance
 CSP_SCRIPT_SRC = (
     "'self'",
+    "'unsafe-eval'",
     "https://sentry.ci.uktrade.digital/",
     "https://cdnjs.cloudflare.com",
     "https://www.googletagmanager.com",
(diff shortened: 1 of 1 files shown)


config/settings.py
```diff
@@ -453,9 +453,11 @@
 # JS tags with a src attribute can only be loaded from ICMS itself or the DBT Sentry instance
 CSP_SCRIPT_SRC = (
     "'self'",
+    "'unsafe-eval'",
```
mikeylockwood, line 456: I don't think we should be allowing `unsafe-eval` in our CSP. Might need to look again at how we load gtm
https://developers.google.com/tag-platform/security/guides/csp

chris-pettinga: https://developers.google.com/tag-platform/security/guides/csp#custom_javascript_variables - unfortunately the M&E team at DBT use custom javascript to check if the user has consented to the cookie policy, so we’re pretty bound to using unsafe-eval unfortunately

mikeylockwood: Can you remind me what these are? My issue main issue is that the code is open source and this would expand the options for an attacker to potentially find a vulnerability

mikeylockwood (approved):

PR #1527, opened 2024-05-15
ICMSLST-2535 - CircleCI test parallelisation

chris-pettinga:
Refactoring our CircleCI config to;

- not do all of our checks within docker, circleci is already running a kind-of docker
- split up code hygiene (black, linters...etc.) and the actual testing jobs as they have no dependencies and can be ran in parallel
- run the test suite itself in parallel between 10 different workers
- store test artifacts in circleCI along with codecov, helping to speed up subsequent runs

### DIFF
--- a/.env.circleci
+++ b/.env.circleci
@@ -3,7 +3,7 @@ APP_ENV=test
 ICMS_SECRET_KEY=local-docker-key
 ICMS_ALLOWED_HOSTS='["caseworker","import-a-licence","export-a-certificate"]'
 ICMS_DEBUG=True
-DATABASE_URL='postgres://postgres:password@db:5432/postgres'  # /PS-IGNORE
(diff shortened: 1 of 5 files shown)


.env.circleci
```diff
@@ -81,3 +81,5 @@ COMPANIES_HOUSE_TOKEN=''
 # PDF signing signatures
 P12_SIGNATURE_BASE_64='' # Value in Parameter Store
 P12_SIGNATURE_PASSWORD='' # Value in Parameter Store
+
+LOCAL_REDIS_URL="redis://localhost:6379"
```
MattHolmes123, line 85: These two settings will change in my PR but I guess that's a me problem

MattHolmes123 (approved):
Assuming it all works then looks good to me 👍

PR #1530, opened 2024-05-15
ICMSLST-2552 - Open / Total Email Counts on BEIS / HSE emails

chris-pettinga:
Before:
image

After:
image

### DIFF
--- a/web/templates/partial/case/export/sidebar-management.html
+++ b/web/templates/partial/case/export/sidebar-management.html
@@ -10,13 +10,15 @@
     {% set vr_label = "Variations (" + open_vr_count|string + "/" + all_vt_count|string + ")" %}
     {{ icms_link(request, icms_url('case:variation-request-manage', kwargs={'application_pk': process.pk, 'case_type': 'export'}), vr_label) }}
   {% endif %}
-
-  {% if process.process_type == "CertificateOfFreeSaleApplication" %}
-    {{ icms_link(request, icms_url('case:manage-case-emails', kwargs={'application_pk': process.pk, 'case_type': 'export'}), 'HSE Emails') }}
-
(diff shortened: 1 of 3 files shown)


web/templates/partial/case/export/sidebar-management.html
```diff
@@ -10,13 +10,15 @@ <h4>Export Case</h4>
...
-  {% if process.process_type == "CertificateOfFreeSaleApplication" %}
-    {{ icms_link(request, icms_url('case:manage-case-emails', kwargs={'application_pk': process.pk, 'case_type': 'export'}), 'HSE Emails') }}
-
-  {% elif process.process_type == "CertificateofGoodManufacturingPractice" %}
-    {{ icms_link(request, icms_url('case:manage-case-emails', kwargs={'application_pk': process.pk, 'case_type': 'export'}), 'BEIS Emails') }}
-  {% endif %}
+  
+  {% with total = process.case_emails.count(), processed = process.case_emails.exclude(status="DRAFT").count() %}
+    {% if process.process_type == "CertificateOfFreeSaleApplication" %}
+      {{ icms_link(request, icms_url('case:manage-case-emails', kwargs={'application_pk': process.pk, 'case_type': 'export'}), 'HSE Emails (' + processed|string + '/' + total|string + ')') }}
```
mikeylockwood, line 16: In V1 the counts are open / total. So if there is one open request it would be 1 / 1. Once that request is closed it changes to 0 / 1

chris-pettinga: that's very unintuitive IMO but if V1 says it's right, then it's right. Have updated

mikeylockwood (approved):

PR #1535, opened 2024-05-16
ICMSLST-2660 Use Section5Clause records when adding a section 5 goods line.

MattHolmes123:


MattHolmes123:
Will be merged via this pr:
https://github.com/uktrade/icms/pull/1538

PR #1545, opened 2024-05-21
ICMSLST-2403 Fix Quick Issue logic.

MattHolmes123:
Changes:
  - Remove SUBMITTED from current_status field as it isn't valid.
  - Add different logic for import and export variation requests.
  - Add comment in Authorise documents step in case we get bugs in the future relating to status changes.
  - Fix incorrect data-modal-message label for export applications.

### DIFF
--- a/web/templates/web/domains/case/manage/base.html
+++ b/web/templates/web/domains/case/manage/base.html
@@ -46,10 +46,20 @@
         action="{{ icms_url('case:quick-issue', kwargs={'application_pk': process.pk, 'case_type': case_type}) }}"
         class="form-inline">
         {{ csrf_input }}
-        {% if process.can_quick_issue()  %}
-          <input type="submit" class="button" value="Quick Issue"/>
+        {# Variation request logic is different for import and export applications. #}
+        {% set app_approved = process.status == "PROCESSING" and process.decision == "APPROVE" %}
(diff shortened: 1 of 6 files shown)


web/templates/web/domains/case/manage/base.html
```diff
@@ -46,10 +46,20 @@
...
+        {% set app_approved = process.status == "PROCESSING" and process.decision == "APPROVE" %}
+        {% if process.is_import_application() %}
+          {% if app_approved or (process.status == "VARIATION_REQUESTED" and process.variation_decision == "APPROVE") %}
+            <input type="submit" class="button" value="Quick Issue" data-confirm="Are you sure you want to issue this licence?"/>
+          {% else %}
+            <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
+          {% endif %}
         {% else %}
-          <input type="button" class="button" value="Quick Issue" data-modal-message="The case must be approved to issue a licence."/>
+          {% if app_approved or (process.status == "VARIATION_REQUESTED") %}
```
chris-pettinga, line 58: So you're alllowed to issue export applications even if there's a variation requested?

MattHolmes123: Discussed offline 👍

PR #1547, opened 2024-05-21
ICMSLST-2631 - Active User Report

marcuspp:


web/reports/interfaces.py
```diff
@@ -1174,10 +1176,19 @@ def get_queryset(self) -> QuerySet:
...
         )
+        _permissions_filter = Q(
+            Q(groups__isnull=True)
+            | Q(
+                groups__name__in=[
+                    Perms.obj.importer.get_group_name(),
+                    Perms.obj.exporter.get_group_name(),
+                ]
+            )
+        )
```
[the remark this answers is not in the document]
MattHolmes123: We also have a `get_users_with_permission` function in `work/icms/web/permissions/service.py` that is used in several other places.

web/reports/interfaces.py
```diff
@@ -1155,3 +1161,66 @@ def extra_fields(self, ia: ImportApplication) -> dict[str, Any]:
...
+
+    def get_queryset(self) -> QuerySet:
+        _filter = Q(
+            date_joined__date__range=(
+                self.filters.date_from,
+                self.filters.date_to,
+            )
+        )
+        _permissions_filter = Q(
+            Q(groups__isnull=True)
```
MattHolmes123, line 1180: Why do we need to return users with no groups?

web/reports/interfaces.py
```diff
@@ -1155,3 +1161,66 @@ def extra_fields(self, ia: ImportApplication) -> dict[str, Any]:
...
+                groups__permissions__codename__in=[
+                    Perms.sys.exporter_access.codename,
+                    Perms.sys.importer_access.codename,
+                ]
+            )
+        )
+        return (
+            user_list_view_qs()
+            .filter(is_active=True)
+            .filter(_filter, _permissions_filter)
```
MattHolmes123, line 1191: `.filter(_filter, _permissions_filter, is_active=True)`

web/reports/interfaces.py
```diff
@@ -1155,3 +1161,66 @@ def extra_fields(self, ia: ImportApplication) -> dict[str, Any]:
...
+                exporters=ArraySubquery(
+                    ExporterUserObjectPermission.objects.filter(user__pk=OuterRef("pk"))
+                    .values_list("content_object__name", flat=True)
+                    .distinct()
+                ),
+                importers=ArraySubquery(
+                    ImporterUserObjectPermission.objects.filter(user__pk=OuterRef("pk"))
+                    .values_list("content_object__name", flat=True)
+                    .distinct()
+                ),
```
MattHolmes123, line 1202: This returns the parent company for agents, is that the desired behaviour?

MattHolmes123 (approved):

web/reports/interfaces.py
marcuspp: Users who have logged in and haven't created an access request need to be included

marcuspp: The filtering is really exclude caseworkers/staff and include everyone else

PR #1548, opened 2024-05-21
ICMSLST-2549 - Constabulary Emails - Expired Authorities Warning

chris-pettinga:
Before:
image

After:
image

### DIFF
--- a/web/domains/case/views/views_email.py
+++ b/web/domains/case/views/views_email.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING
 from django.conf import settings
 from django.contrib.auth.decorators import login_required, permission_required
 from django.db import transaction
-from django.db.models import Q
+from django.db.models import BooleanField, ExpressionWrapper, Q
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
@@ -72,6 +72,27 @@ def manage_case_emails(
     else:
         info_email = ""
 
+    verified_section_5_authorities = []
+    verified_firearms_authorities = []
+
+    if application.process_type in [
+        SILApplication.PROCESS_TYPE,
+        OpenIndividualLicenceApplication.PROCESS_TYPE,
+    ]:
+        # it's a SIL or OIL application, get the verified firearm authorities
+        today_date = timezone.now().date()
+        specific_application = application.get_specific_model()
+        verified_firearms_authorities = specific_application.verified_certificates.all().annotate(
+            is_expired=ExpressionWrapper(Q(end_date__lt=today_date), output_field=BooleanField())
+        )
+
+        if application.process_type == SILApplication.PROCESS_TYPE:
+            verified_section_5_authorities = specific_application.verified_section5.all().annotate(
+                is_expired=ExpressionWrapper(
+                    Q(end_date__lt=today_date), output_field=BooleanField()
+                )
+            )
+
     context = {
         "process": application,
         "page_title": "Manage Emails",
@@ -83,6 +104,8 @@ def manage_case_emails(
         "info_email": info_email,
         "readonly_view": readonly_view,
         "record_response_form": forms.CaseEmailResponseForm(),
+        "verified_section_5_authorities": verified_section_5_authorities,
+        "verified_firearms_authorities": verified_firearms_authorities,
     }
 
     return render(
--- a/web/templates/web/domains/case/manage/case-emails.html
+++ b/web/templates/web/domains/case/manage/case-emails.html
@@ -18,6 +18,135 @@
     <h4>{{ email_subtitle }}</h4>
   {% endif %}
 
+  {% if verified_firearms_authorities %}
+    <div class="info-box {% if verified_firearms_authorities.filter(is_expired=True).exists() %}info-box-warning{% else %}info-box-info{% endif %}">
+      <div class="screen-reader-only">Warning information box,</div>
+      <h4>Verified Firearms Authorities</h4>
+        <p>One or more verified Firearms Authorities have been selected by the applicant. Verified Firearms Authority documents may still be attached to constabulary emails if required.</p>
+
+        {% if verified_firearms_authorities.filter(is_expired=True).exists() %}
+          <br>
+          <span>At least one verified Firearms Authority has expired or has been archived since the application was submitted. These are indicated with a <span class="icon-warning orange-text"></span> icon.</span>
+        {% endif %}
+        <table class="setoutList">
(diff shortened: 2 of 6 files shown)


web/domains/case/views/views_email.py
```diff
@@ -72,6 +72,20 @@ def manage_case_emails(
...
+    if application.process_type == SILApplication.PROCESS_TYPE:
+        # it's a SIL application, get the verified section5 and verified firearm authorities
+        sil_application = application.get_specific_model()
+        verified_section_5_authorities = sil_application.verified_section5.all()
+        verified_firearms_authorities = sil_application.verified_certificates.all()
+
+    elif application.process_type == OpenIndividualLicenceApplication.PROCESS_TYPE:
+        # it's an OIL application, get the verified firearm authorities
+        oil_application = application.get_specific_model()
+        verified_firearms_authorities = oil_application.verified_certificates.all()
```
MattHolmes123, line 87: Do we need to show `user_imported_certificates` as well here? (I haven't checked V1).
The use can upload their own certificate when there are no verified ones for the chosen importer:
image

web/templates/web/domains/case/manage/case-emails.html
```diff
@@ -18,6 +18,139 @@ <h3>{{ email_title }}</h3>
...
+  {% if verified_firearms_authorities %}
+    <div class="info-box info-box-warning">
+      <div class="screen-reader-only">Warning information box,</div>
+      <h4>Verified Firearms Authorities</h4>
+        <p>One or more verified Firearms Authorities have been selected by the applicant. Verified Firearms Authority documents may still be attached to constabulary emails if required.</p>
+
+        {% if verified_firearms_authorities.filter(end_date__lt=today_date) %}
+          <br>
+          <span>At least one verified Firearms Authority has expired or has been archived since the application was submitted. These are indicated with a <span class="icon-warning orange-text"></span> icon.</span>
+        {% endif %}
```
MattHolmes123, line 30: Rather than writing orm code in the template we could pass this to the template as context.
Same with this:
`{% with expired = firearm_authority.end_date < today_date %}`
It could be an annotation on the `verified_firearms_authorities` queryset.

I prefer to keep templates logic simple where possible.

MattHolmes123: Same comments apply to `verified_section_5_authorities`

MattHolmes123: Feel free to do whatever you want though, but in my new world order we'd have real dumb templates and no orm code in them 😄

web/templates/web/domains/case/manage/case-emails.html
```diff
@@ -18,6 +18,139 @@ <h3>{{ email_title }}</h3>
...
+                  <td>
+                    <span class="text-widget">{% if expired %}<p class="icon-warning orange-text"></p>{% else %}Not Expired{% endif %}</span>
+                  </td>
+                  <td>
+                    <span class="text-widget">{% if firearm_authority.is_active %}Not Archived{% else %}Archived{% endif %}</span>
+                  </td>
+                  <td>
+                    <ul class="menu-out flow-across">
+                      <li>
+                        <a target="_blank" href="{{ icms_url('importer-firearms-edit', kwargs={'pk': firearm_authority.pk}) }}">View details</a>
```
MattHolmes123, line 80: Missing ` rel="noopener noreferrer"`

web/templates/web/domains/case/manage/case-emails.html
```diff
@@ -18,6 +18,139 @@ <h3>{{ email_title }}</h3>
...
+                  <td>
+                    <span class="text-widget">{% if expired %}<p class="icon-warning orange-text"></p>{% endif %}</span>
+                  </td>
+                  <td>
+                    <span class="text-widget">{% if section_5_authority.is_active %}Not Archived{% else %}Archived{% endif %}</span>
+                  </td>
+                  <td>
+                    <ul class="menu-out flow-across">
+                      <li>
+                        <a target="_blank" href="{{ icms_url('importer-section5-edit', kwargs={'pk': section_5_authority.pk}) }}">View details</a>
```
MattHolmes123, line 142: Missing  rel="noopener noreferrer"

web/tests/conftest.py
```diff
@@ -62,6 +62,8 @@
 from web.tests.helpers import CaseURLS, get_test_client
 from web.tests.utils.search.conftest import Build, importer_one_fixture_data  # NOQA
 
+from ..domains.firearms.models import FirearmsAuthority
+from ..domains.section5.models import Section5Authority
```
MattHolmes123, line 66: Absolute imports when not using a relative import in the same directory.

MattHolmes123: Actually `from web.models import FirearmsAuthority, Section5Authority`

web/domains/case/views/views_email.py
chris-pettinga: I copied the wording from V1 which explicitly lists 'Verified Firearm Authorities' - I will check if anything appears when they upload their own

web/domains/case/views/views_email.py
```diff
@@ -72,6 +72,20 @@ def manage_case_emails(
     else:
         info_email = ""
 
+    verified_section_5_authorities = []
+    verified_firearms_authorities = []
+
+    if application.process_type == SILApplication.PROCESS_TYPE:
+        # it's a SIL application, get the verified section5 and verified firearm authorities
+        sil_application = application.get_specific_model()
+        verified_section_5_authorities = sil_application.verified_section5.all()
```
mikeylockwood, line 81: You can add some nifty annotations to these to make them easier to handle

```
from django.db.models import BooleanField, ExpressionWrapper, Q
...

sil_application.verified_section5.annotate(
    is_expired=ExpressionWrapper(
        Q(end_date__lt=date.today()),
        output_field=BooleanField()
    )
)
```
Now `is_expired` will be a boolean on the queryset

web/templates/web/domains/case/manage/case-emails.html
mikeylockwood: Didn't see this comment. Added an example for this above :)

chris-pettinga: thanks both

web/domains/case/views/views_email.py
chris-pettinga: @MattHolmes123 just confirmed that V1 doesn't bother showing user_imported_certificates here

MattHolmes123 (approved):

PR #1553, opened 2024-05-23
ICMSLST-2628 Refactor env vars to work with copilot secret init.

MattHolmes123:
Changes:
  - Move CloudFoundryEnvironment to cf_env.py
  - Remove common EnvironmentBase class so each env can diverge.
  - Add env_prefix to DBTPlatformEnvironment class.
  - Refactor all env files to set ICMS_ prefix.
  - Fix incorrect default DJANGO_SETTINGS_MODULE value.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -28,11 +28,11 @@ from web.utils.sentry import init_sentry
 BASE_DIR: Path = Path(__file__).resolve().parent.parent
 
 # Application definition
-DEBUG = env.icms_debug
+DEBUG = env.debug
 WSGI_APPLICATION = "config.wsgi.application"
 APP_ENV = env.app_env
-SECRET_KEY = env.icms_secret_key
-ALLOWED_HOSTS = env.allowed_hosts
+SECRET_KEY = env.secret_key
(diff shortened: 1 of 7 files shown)


mikeylockwood (approved):

config/settings.py
```diff
@@ -407,7 +407,9 @@
 }
 
 # Django Log Formatter ASIM settings
-if is_copilot():
+# is_copilot() evaluates to True locally now so also check env.app_env
+# This was done to use DBTPlatformEnvironment instead of CloudFoundryEnvironment going forward.
+if is_copilot() and env.app_env != "local":
```
chris-pettinga, line 412: Can we not put this check in the `is_copilot()` method? if we're on local we're not in copilot

MattHolmes123: is_copilot is a function in the dbt_copilot_python package so nope.
This line is temporary.
When we are deployed we can remove the `is_copilot` call completely and the line will be this:
`if env.app_env != "local":`

chris-pettinga: @MattHolmes123 - ah got you thanks, thought it was one of ours.

PR #1567, opened 2024-05-30
ICMSLST-2670 - Admin view redirects

marcuspp:
Updated admin screens
 - Sanction Emails
 - Importer
 - Exporter

Updated redirect and added django message popup

Screenshot 2024-05-30 at 13 22 13

### DIFF
--- a/web/domains/exporter/views.py
+++ b/web/domains/exporter/views.py
@@ -1,11 +1,13 @@
 from typing import Any
 
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required, permission_required
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponse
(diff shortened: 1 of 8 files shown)


web/domains/exporter/views.py
```diff
@@ -91,8 +92,8 @@ def create_exporter(request: AuthenticatedHttpRequest) -> HttpResponse:
         form = ExporterForm(request.POST)
         if form.is_valid():
             exporter: Exporter = form.save()
-
-            return redirect(reverse("exporter-edit", kwargs={"pk": exporter.pk}))
+            messages.info(request, "Exporter created successfully."),
+            return redirect(reverse("exporter-list") + f"?name={exporter.name}")
```
mikeylockwood, line 96: You should use `urlencode` for this
```
from django.utils.http import urlencode
...


urlargs = urlencode({"name": exporter.name})
return redirect(f"{reverse('exporter-list')}?f{urlargs}")
```

Use urlencode on all the others as well

MattHolmes123 (approved):

PR #1584, opened 2024-06-11
ICMSLST-2673 Fix section 5 goods having hardcoded subsection in pdf.

MattHolmes123:
Added [another bug](https://uktrade.atlassian.net/browse/ICMSLST-2708) to revisit this as the same problem is in one of the reports.

Once I have written tests and fixed ICMSLST-2708 I hope to remove some of the duplication relating to fetching FA-SIL goods.

I believe there are at least 3 places that fetch FA-SIL goods places that could do with a refactor.

### DIFF
--- a/web/tests/utils/pdf/test_utils.py
+++ b/web/tests/utils/pdf/test_utils.py
@@ -8,6 +8,7 @@ from web.models import SILGoodsSection582Other  # /PS-IGNORE
 from web.models import (
     Country,
     EndorsementImportApplication,
+    Section5Clause,
     SILGoodsSection1,
     SILGoodsSection2,
     SILGoodsSection5,
@@ -15,7 +16,7 @@ from web.models import (
 )
 from web.types import DocumentTypes
 from web.utils.pdf import utils
-from web.utils.pdf.utils import _get_importer_eori_numbers
+from web.utils.pdf.utils import _get_fa_sil_goods, _get_importer_eori_numbers
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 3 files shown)


web/utils/pdf/utils.py
```diff
@@ -223,6 +224,50 @@ def _get_fa_dfl_goods(application: DFLApplication) -> list[str]:
     ]
 
 
+# TODO: Revisit in ICMSLST-2708
+def _get_fa_sil_goods(application: SILApplication) -> Iterable[tuple[str, int | str]]:
+    """Return all related FA SIL goods"""
+
+    for g in application.goods_section1.filter(is_active=True):
+        section = "1"
```
mikeylockwood, line 232: Why set sections as variables instead of just hardcoding into the text if they don't change?

mikeylockwood (approved):

web/tests/utils/pdf/test_utils.py
```diff
@@ -386,3 +387,64 @@ def test__get_importer_eori_numbers_wood(wood_app_submitted):
...
     assert _get_importer_eori_numbers(wood_app_submitted) == ["GB0123456789ABCDE"]
+
+
+def test__get_fa_sil_goods(fa_sil_app_submitted, ilb_admin_user):
+    expected = [
+        ("Section 1 goods to which Section 1 of the Firearms Act 1968, as amended, applies.", 111),
+        ("Section 2 goods to which Section 2 of the Firearms Act 1968, as amended, applies.", 222),
+        (
+            "Section 5 goods to which Section 5(A) of the Firearms Act 1968, as amended, applies.",
+            333,
```
mikeylockwood, line 398: One of these tests should be for an unlimited quantity

mikeylockwood (approved):

web/utils/pdf/utils.py
MattHolmes123: I plan on refactoring it in that other bug I raised.

web/tests/utils/pdf/test_utils.py
MattHolmes123: Will do that in the next ticket.

PR #1591, opened 2024-06-18
ICMSLST-2677 - DIT to DBT notification document

chris-pettinga:


mikeylockwood:
It does look a bit weird in the table on the exporter side, but I don't have any suggestions
image

mikeylockwood (approved):

PR #1593, opened 2024-06-19
ICMSLST-2623 BEIS Case Email Download Link

mikeylockwood:
Adding ability to include download links in case emails. 

This ticket is for the BEIS email but includes a generic solution that should make it easier to provide attachment links in all case emails.

The page contains the downloads as well as details on the exporter
image


web/management/commands/utils/add_email_template_data.py
```diff
@@ -60,6 +60,7 @@
     (EmailTypes.WITHDRAWAL_OPENED, "e3499a7c-ad6e-4ceb-bed3-d0a9a137f6a9"),  # /PS-IGNORE
     (EmailTypes.WITHDRAWAL_REJECTED, "ffbed03f-bebc-4f87-a602-e83df0c72890"),  # /PS-IGNORE
     (EmailTypes.CASE_EMAIL, "657aa427-8dda-496a-b933-fa7be10f16fd"),  # /PS-IGNORE
+    (EmailTypes.CASE_EMAIL_WITH_DOCUMENTS, "d208a30a-f03d-447a-89cf-c29f192b3cc4"),  # /PS-IGNORE
```
MattHolmes123, line 63: @mikeylockwood this isn't corrent

MattHolmes123:
I think you should remove `((document_url))` from the template as it doesn't need the full link displaying:
image
image

Looking at the invite a contact email you need to triple quote it:
`Click this link if you want to [accept the invite](((accept_invite_url))).`

web/management/commands/utils/add_email_template_data.py
MattHolmes123: Nevermind... Didn't see the V2 folder.
My bad

web/templates/web/domains/case/download-case-email-documents.html
```diff
@@ -0,0 +1,48 @@
...
+          <th scope="col">Document</th>
+        </tr>
+        </thead>
+        <tbody>
+          {% for doc in documents %}
+            <td>{{ doc.name }}</td>
+            <td>
+              <a href="{{ doc.url }}">
+                  View Document
+              </a>
```
MattHolmes123, line 40: When clicking this it opens the document in the current tab rather than downloading the document.

web/domains/case/forms.py
```diff
@@ -318,11 +319,11 @@ class Meta:
         fields = ReassignOwnershipBaseForm.Meta.fields
 
 
-class DownloadDFLCaseDocumentsForm(forms.Form):
-    link: ImportApplicationDownloadLink | None
-    # The same error message for every outcome
+class DownloadDocumentsFormBase(forms.Form):
+    link_class: type[ConstabularyLicenceDownloadLink] | type[CaseEmailDownloadLink]
```
MattHolmes123, line 323: `type[ConstabularyLicenceDownloadLink | CaseEmailDownloadLink]`

MattHolmes123: Also from what I can see `link_class` should be typed as a `ClassVar`

web/domains/case/models.py
```diff
@@ -293,6 +296,35 @@ def is_draft(self):
         return self.status == self.Status.DRAFT
 
 
+def create_check_code():
```
MattHolmes123, line 299: `-> int`

web/domains/case/views/views_documents.py
```diff
@@ -60,25 +71,138 @@ def get_initial(self) -> dict:
...
+                link.refresh_from_db()
+
+            if link.failure_count == link.FAILURE_LIMIT:
+                link.expired = True
+                link.save()
+
+        return super().form_invalid(form)
+
+
+class DownloadCaseEmailDocumentsFormView(DownloadCaseDocumentsFormView):
```
MattHolmes123, line 115: I wonder if the method_decorators in the base class are called corrently?
`@method_decorator(ratelimit(key="ip", rate="10/m", method=UNSAFE, block=True), name="post")`
`@method_decorator(ratelimit(key="ip", rate="20/m", method="GET", block=True), name="get")`

I am too tired to think about it :D

web/domains/case/views/views_documents.py
```diff
@@ -60,25 +71,138 @@ def get_initial(self) -> dict:
...
+            case CertificateOfGoodManufacturingPracticeApplication():
+                return context | {
+                    "documents": [
+                        {
+                            "name": GMPFile.Type(doc.gmpfile.file_type).label,
+                            "url": create_presigned_url(doc.path),
+                        }
+                        for doc in link.case_email.attachments.all()
+                    ],
+                }
```
MattHolmes123, line 170: Will this be extended later?

web/domains/case/views/views_documents.py
```diff
@@ -102,44 +226,26 @@ def form_valid(self, form: DownloadDFLCaseDocumentsForm) -> HttpResponse:
...
-                link.save()
-
-        return super().form_invalid(form)
-
 
 @method_decorator(ratelimit(key="ip", rate="10/m", method=UNSAFE, block=True), name="post")
-class RegenerateDFLCaseDocumentsDownloadLinkView(View):
+class RegenerateDownloadLinkView(View):
     http_method_names = ["post"]
+    link_class: type[ConstabularyLicenceDownloadLink] | type[CaseEmailDownloadLink]
```
MattHolmes123, line 233: Same comment as earlier.
`ClassVar[type[Foo | Bar]]`

web/domains/case/views/views_documents.py
```diff
@@ -102,44 +226,26 @@ def form_valid(self, form: DownloadDFLCaseDocumentsForm) -> HttpResponse:
...
+    download_url_name: str
+
+    @staticmethod
+    def resend_email(obj: DFLApplication | CaseEmail) -> None:
```
MattHolmes123, line 237: The type for `obj` doesn't match what RegenerateDFLCaseDocumentsDownloadLinkView or RegenerateCaseEmailDocumentsDownloadLinkView define.

They both have a `link` paramater.

web/mail/emails.py
```diff
@@ -469,10 +470,12 @@ def send_retract_mailshot_email_to_organisations(
 
 
 def send_case_email(case_email: CaseEmailModel) -> None:
-    # TODO: ICMSLST-2333 Gov Notify - Email attachments
     recipients = get_email_addresses_for_case_email(case_email)
     for recipient in recipients:
-        CaseEmail(case_email=case_email, to=[recipient]).send()
+        if case_email.attachments.exists():
```
MattHolmes123, line 475: Do we need to filter for inactive attachments?
There is a `is_active` field.

web/mail/messages.py
```diff
@@ -451,6 +456,25 @@ def get_context(self) -> dict[str, Any]:
         return {"subject": self.case_email.subject, "body": self.case_email.body}
 
 
+class CaseEmailWithDocuments(CaseEmail):
```
MattHolmes123, line 459: Both `CaseEmail` and `CaseEmailWithDocuments` should be defined using `@final` and have a common base class.
`BaseVariationRequestEmail` is an example of this.

web/mail/messages.py
```diff
@@ -433,11 +434,15 @@ def get_context(self) -> dict[str, Any]:
...
-@final
 class CaseEmail(GOVNotifyEmailMessage):
     name = EmailTypes.CASE_EMAIL
 
-    def __init__(self, *args: Any, case_email: CaseEmailModel, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        *args: Any,
+        case_email: CaseEmailModel,
+        **kwargs: Any,
```
MattHolmes123, line 444: Remove `,` and this will be formatted on to a single line

MattHolmes123 (approved):

web/mail/emails.py
mikeylockwood: I don't think so, I would include all attachments attachments that are added to the email

web/domains/case/views/views_documents.py
mikeylockwood: Yes depending on the type of email

PR #1597, opened 2024-06-20
ICMSLST-2715 - CFS PDF - over 25 products causes a nasty page break - COM PDF - Manufacturing process overlaps signature block

chris-pettinga:


mikeylockwood (changes requested):
The space available on the page and the space occupied by the products is too variable. There are a lot of examples of products the span over multiple lines. Some countries will have translation text on the schedule as well. Also the schedule statement text on on the page will vary depending on what is entered in the cfs application. Also what happens when there are 100s of products?

The spacing needs to be worked out in the document css and cannot be done by python, there are too many variables

chris-pettinga:
@mikeylockwood  - happy for me to close this?

PR #1599, opened 2024-06-24
ICMSLST-2696 Show organisation address and postcode for all app types.

MattHolmes123:
Import application view:
caseworker_8080_case_import_24_admin_manage_
Export application view:
caseworker_8080_case_export_4_admin_manage_

### DIFF
--- a/web/templates/web/domains/case/import/partials/process.html
+++ b/web/templates/web/domains/case/import/partials/process.html
@@ -30,8 +30,9 @@
   <div class="two columns">
     <span>
       {{ process.application_type.get_type_display() }}
-      -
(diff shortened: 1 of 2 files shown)


web/templates/web/domains/case/import/partials/process.html
```diff
@@ -30,8 +30,9 @@
   <div class="two columns">
     <span>
       {{ process.application_type.get_type_display() }}
-      -
-      {{ process.application_type.get_sub_type_display() }}
+      {% if process.application_type.type == "FA" %}
```
mikeylockwood, line 33: `== process.application_type.Types.FIREARMS`

mikeylockwood (approved):

PR #1603, opened 2024-06-25
ICMSLST-2708 Fix FA-SIL goods descriptions in reports.

MattHolmes123:
Update the code used to generate goods descriptions in the SILFirearmsLicenceInterface report.

This change does nothing to address the fact that we create fa-sil goods descriptions in 4 separate places.
- Chief serializers 
- Email Template context
- Report data
- PDF licence context

### DIFF
--- a/web/reports/interfaces.py
+++ b/web/reports/interfaces.py
@@ -68,7 +68,7 @@ from web.models import (
     UserImportCertificate,
 )
 from web.permissions import Perms
-from web.utils.pdf.utils import _get_sil_section_labels, get_fa_sil_goods_item
+from web.utils.pdf.utils import get_fa_sil_goods_item
 from web.utils.search.app_data import _add_import_licence_data
 from web.utils.sentry import capture_exception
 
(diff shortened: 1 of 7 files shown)


web/reports/interfaces.py
```diff
@@ -880,7 +884,7 @@ def get_constabularies(self) -> QuerySet:
             Q(oil_application__id=OuterRef("pk")) | Q(sil_application__id=OuterRef("pk"))
         ).values("constabulary__name")
 
-    def get_queryset(self) -> QuerySet:
+    def get_queryset(self) -> dict:
```
mikeylockwood, line 887: `.values()` does not return a dict. It returns a queryset of dicts

MattHolmes123: Yer I mentioned in slack that I was fixing that now

MattHolmes123: Also values returns a QuerySet (as do all QuerySet methods).
It's only when they get iterated over do they actually return something different.

marcuspp (approved):

PR #1619, opened 2024-07-03
ICMSLST-2734 Fixing page breaks and adding page numbers for CFS pdfs

mikeylockwood:
Adding page numbers and fixing running header and footer content for CFS pdfs

Page counts and the reference sit at the top of all pages
image

Bottom of the page has the running footer. Page breaks does not cause text to superimpose
image

### DIFF
--- a/web/utils/pdf/pages.py
+++ b/web/utils/pdf/pages.py
@@ -1,0 +1,50 @@
+import io
+from typing import Any
+
+import pypdf
+from reportlab.lib.pagesizes import A4, portrait
+from reportlab.pdfgen import canvas
+
+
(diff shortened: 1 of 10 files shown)


web/utils/pdf/pages.py
```diff
@@ -0,0 +1,50 @@
...
+    issue_date = context["issue_date"]
+    reference = context["reference"]
+
+    left_margin = 73
+    top_margin = 780
+    page_total = len(reader.pages)
+    for page_number, page in enumerate(reader.pages):
+        packet = io.BytesIO()
+        new_page = canvas.Canvas(packet, pagesize=portrait(A4))
+        new_page.setFont("Helvetica", 10)
```
mikeylockwood, line 26: Will need to check the font when deployed

PR #1625, opened 2024-07-08
ICMSLST-2718 - SIL Goods Prep edit fix

marcuspp:


mikeylockwood (approved):

web/domains/case/_import/fa_sil/forms.py
```diff
@@ -475,11 +475,16 @@ class Meta:
 
 
 class ResponsePrepSILGoodsSection5Form(ResponsePrepBaseForm):
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.fields["quantity"].required = False
```
MattHolmes123, line 481: Should it always be optional or only when unlimited quantity is selected?

MattHolmes123 (approved):

web/domains/case/_import/fa_sil/forms.py
marcuspp: The clean method covers it if unlimited is not selected

PR #1637, opened 2024-07-10
ICMSLST-2709 Data Migration - CFS Schedule product eligibility fix

mikeylockwood:


web/templates/web/domains/case/export/cfs-view.html
```diff
@@ -29,7 +29,7 @@
   {% endcall %}
 
   {% for schedule in schedules %}
-    {% set schedule_legislations = schedule.legislations.filter(is_active=True) %}
+    {% set schedule_legislations = schedule.legislations.all() %}
```
mikeylockwood, line 32: When viewing CFS applications, we want to see all legislations selected for an application regardless whether the legislation is still active

MattHolmes123 (approved):

PR #1638, opened 2024-07-11
ICMSLST-2770 Add view to set new user details when first visiting site.

MattHolmes123:
Changes:
  - Add NewUserUpdateView view.
  - Update UserFullyRegisteredMiddleware to redirect to new view.
  - Remove "one_login_unset" from User.full_name property.
  - Move ONE_LOGIN_UNSET_NAME to constants.py file.
  
New view:
import-a-licence_8080_user_40_new-user_

### DIFF
--- a/web/domains/user/models.py
+++ b/web/domains/user/models.py
@@ -4,6 +4,8 @@ from django.db import models
 from guardian.core import ObjectPermissionChecker
 from guardian.mixins import GuardianUserMixin
 
+from web.one_login.constants import ONE_LOGIN_UNSET_NAME
+
 
 class User(GuardianUserMixin, AbstractUser):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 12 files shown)


web/domains/user/models.py
```diff
@@ -44,8 +44,16 @@ def __str__(self):
         return self.full_name
 
     @property
-    def full_name(self):
-        return " ".join(filter(None, (self.title, self.first_name, self.last_name)))
+    def full_name(self) -> str:
+        """Returns full name of user."""
+        # To prevent circular dependency
+        from web.one_login.backends import ONE_LOGIN_UNSET_NAME
```
mikeylockwood, line 50: Might be nicer to move this to a constants file or to settings

mikeylockwood (approved):

web/domains/user/models.py
MattHolmes123: yer true.

PR #1643, opened 2024-07-11
ICMSLST-2719 Fix view mailshot detail sent to check boxes

mikeylockwood:
Bugfix in mailshot detail view, sent to checkboxes are always checked

### DIFF
--- a/web/templates/web/domains/mailshot/partials/detail.html
+++ b/web/templates/web/domains/mailshot/partials/detail.html
@@ -36,14 +36,14 @@
         <div class="row">
           <div class="twelve columns">
             <label class="radio-label">
-              <input type="checkbox" aria-label="Importers and Agents" checked="checked" disabled="disabled">Importers and Agents
+              <input type="checkbox" aria-label="Importers and Agents" disabled {% if object.is_to_importers %}checked{% endif %}>Importers and Agents
             </label>
           </div>
         </div>
         <div class="row">
           <div class="twelve columns">
             <label class="radio-label">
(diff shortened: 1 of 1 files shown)


chris-pettinga (approved):

web/templates/web/domains/mailshot/partials/detail.html
```diff
@@ -36,14 +36,14 @@
         <div class="row">
           <div class="twelve columns">
             <label class="radio-label">
-              <input type="checkbox" aria-label="Importers and Agents" checked="checked" disabled="disabled">Importers and Agents
+              <input type="checkbox" aria-label="Importers and Agents" disabled="disabled" {{ 'checked="checked"' if object.is_to_importers else ""}}>Importers and Agents
```
chris-pettinga, line 39: my measly personal opinion is that the below is clearer but up to you:

```suggestion
              <input type="checkbox" aria-label="Importers and Agents" disabled="disabled" {% if  object.is_to_importers %}checked="checked"{% endif %}>Importers and Agents
```

chris-pettinga: also no need to do `checked="checked"`, the presence of the `checked` attribute is fine

MattHolmes123: I never knew those kind of attributes are called [boolean attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML).

You learn something new every day :)

MattHolmes123 (approved):

PR #1644, opened 2024-07-12
ICMSLST-2707 Data Migration - Sovereign Territory fix

mikeylockwood:


data_migration/models/reference/country.py
```diff
@@ -13,6 +13,14 @@ class Country(MigrationBase):
     commission_code = models.CharField(max_length=20, null=False)
     hmrc_code = models.CharField(max_length=20, null=False)
 
+    @classmethod
+    def data_export(cls, data: dict[str, Any]) -> dict[str, Any]:
+        name = data["name"]
+        if name in ["Mali", "Liberia", "Aruba", "Cameroon", "Gambia"]:
+            data["type"] = "SOVEREIGN_TERRITORY"
```
mikeylockwood, line 20: Using text as I would rather not import web models into the data migration code. The test checks the constant so will flag if there is a mismatch

marcuspp (approved):

PR #1653, opened 2024-07-16
ICMSLST-2742 Add Application Update Withdrawn email.

MattHolmes123:
Send an email to the applicant when an Application Update has been withdrawn by an ILB admin.

New email:
image


web/mail/emails.py
```diff
@@ -350,6 +351,15 @@ def send_application_update_email(update_request: UpdateRequest) -> None:
         ).send()
 
 
+def send_application_update_withdrawn_email(update_request: UpdateRequest) -> None:
+    application = (
```
marcuspp, line 355: Pass the application in, as the function this is called from already has it.

marcuspp (approved):

web/mail/emails.py
MattHolmes123: I did consider that but I was intentionally keeping it the same as the `send_application_update_email` function.
Also If I change it to instead use application then this function could be misused.
At least this forces the need for an update_request to exist.
What you think?

mikeylockwood: I'm assuming the suggesting was to pass application request in as well as the update request. I am happy with it as it is

mikeylockwood (approved):

web/mail/emails.py
marcuspp: Okay, I see what you mean that the update request is only used to obtain the application, so only passing in the application could mean this could be triggered incorrectly.

marcuspp (approved):

web/mail/emails.py
MattHolmes123: Yer that's right Marcus

PR #1656, opened 2024-07-17
ICMSLST-2745 - Complete DFL Licence PDF document

chris-pettinga:


web/templates/pdf/import/fa-dfl-licence.html
```diff
@@ -1,22 +1,236 @@
...
+          <ol>
+            <li>If 10 consignments have been imported but the licence is not exhausted, the importer or his agent should
+              return it to Import Licensing Branch with a request for issue of a new licence for the outstanding
+              balance.
+            </li>
+            <li class="mt-1">When the licence is no longer required (whether it has expired or not), or if it is
+              revoked, then
+              return
+              it to<br/>
+              Department for International Trade<br/>
```
mikeylockwood, line 225: Department for Business and Trade

web/templates/pdf/import/fa-dfl-licence.html
```diff
@@ -1,22 +1,236 @@
...
+              Queensway House, West Precinct<br/>
+              Billingham, United Kingdom TS23 2NF<br/>{# /PS-IGNORE #}
+              Email: enquiries.ilb@trade.gov.uk{# /PS-IGNORE #}
```
mikeylockwood, line 228: Probably pass this into context from settings so we don't have to update it in the certificate when the email changes

web/templates/pdf/import/fa-dfl-licence.html
```diff
@@ -1,22 +1,236 @@
...
+        may
+        be necessary to effect importation will be granted.
+      </p>
+      <br/>
+      <div class="avoid-break">
+        <p class="content mb-1">This licence</p>
+        <ul id="this-licence-bullets">
+          <li>is not transferable and may be used only for importations by the holder named above.</li>
+          <li>may be revoked at any time without reason given.</li>
+          <li>may be altered only by the Department for International Trade.</li>
```
mikeylockwood, line 145: Business and Trade

web/templates/pdf/import/fa-dfl-licence.html
mikeylockwood: `"ilb_contact_email": settings.ILB_CONTACT_EMAIL` is already passed into `get_sanctions_context`. Just move this into `get_licence_context`

web/utils/pdf/pages.py
```diff
@@ -2,6 +2,7 @@
 from typing import Any
 
 import pypdf
+from PyPDF2 import PageObject
```
mikeylockwood, line 5: Use`pypdf.PageObject` or

```
if TYPE_CHECKING:
    from pypdf import PageObject
```

web/utils/pdf/pages.py
```diff
@@ -143,3 +144,65 @@ def format_wood_pages(pdf_data: bytes) -> bytes:
...
+            if page_number >= 1:
+                # add the top closing line to this page if it's not the first.
+                # the first page already has the top line in CSS
+                page = add_top_line(page)
+                writer.add_page(page)
+                should_add_top_line = False
+                continue
+            else:
+                # don't bother with any borders if the entire licence is on 1 page
+                return pdf_data
```
mikeylockwood, line 193: If you're not modifying single page pdfs, return after line 154
```
if page_total == 1:
    return pdf_data
```

web/utils/pdf/pages.py
```diff
@@ -143,3 +144,65 @@ def format_wood_pages(pdf_data: bytes) -> bytes:
...
+        new_page.line(left_margin, 5, right_margin, 5)
+        new_page.save()
+        packet.seek(0)
+        new_pdf = pypdf.PdfReader(packet)
+        page.merge_page(new_pdf.pages[0])
+        return page
+
+    should_add_top_line = True
+
+    for page_number, page in enumerate(reader.pages):
```
mikeylockwood, line 181: This is all a bit confusing, if you add an early return before the for loop then there will always be a need to add a line on the page. Just draw a canvas once per page and add what lines you need to it

```
for page_number, page in enumerate(reader.pages, start=1):
    packet = io.BytesIO()
    new_page = canvas.Canvas(packet, pagesize=portrait(A4))

    if page_number > 1:
        new_page.line(left_margin, 5, right_margin, 5)

    if page_number < page_total:
        new_page.line(left_margin, 5, right_margin, 5)

    new_page.save()
    packet.seek(0)
    new_pdf = pypdf.PdfReader(packet)
    page.merge_page(new_pdf.pages[0])
```

web/utils/pdf/pages.py
mikeylockwood: I would rewrite the for loop that uses the functions that use these types and remove them altogether. See suggestion below

web/utils/pdf/pages.py
mikeylockwood: After merging the pages, just add
```
 if "This licence will cease to be valid for importations" in page.extract_text():
     break
```

mikeylockwood: I see, you won't want a bottom line on that page either

```
signature_page = "This licence will cease to be valid for importations" in page.extract_text()
...
if not signature_page and page_number < page_total:
    new_page.line(left_margin, 5, right_margin, 5)
...
if signature_page:
     break
```

chris-pettinga: @mikeylockwood but sometimes there's a nice little disclaimer after the signature page which we want to include in the new PDF but without lines

chris-pettinga: so we need to continue the for loop to get all the pages in

mikeylockwood (approved):

PR #1657, opened 2024-07-17
Log counts of records when running data migration scripts

mikeylockwood:


data_migration/tests/utils/__init__.py
```diff
@@ -49,6 +49,10 @@ def fetch_rows(self):
         self.fetched = False
         yield None
 
+    @property
+    def rowcount(self):
```
MattHolmes123, line 53: `-> int`

MattHolmes123 (approved):

PR #1660, opened 2024-07-18
ICMSLST-2748 - Sanctions PDF

marcuspp:


web/templates/pdf/import/sanctions-licence.html
```diff
@@ -1,157 +1,130 @@
...
-            <p class="address-line">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
-            <p>Email: {{ ilb_contact_email }}</p>
+          </div>
+          <div class="content-block row flex-row wrap">
+            <p class="row title"><strong>4. </strong>Authority responsible for issue</p>
+            <p class="address-line row">Department for Business and Trade</p>
+            <p class="address-line row">Queensway House, West Precinct</p>
+            <p class="address-line row">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
+            <p class="row min-height-40">Email: {{ ilb_contact_email }}</p>
           </div>
```
MattHolmes123, line 56: Noticed this address appears to be hardcoded in at least two other files:
web/templates/pdf/import/sanctions-licence.html
web/templates/pdf/import/wood-licence.html

coded slightly differently and with a slightly different label for the first line

web/utils/pdf/pages.py
```diff
@@ -143,3 +143,79 @@ def format_wood_pages(pdf_data: bytes) -> bytes:
...
+            new_page.drawRightString(165, top_margin, "ELECTRONIC LICENCE")
+            new_page.setFont("Helvetica", 10)
+            new_page.drawRightString(
+                455,
+                top_margin,
+                f"issued on {start_date} and sent to HM Revenue and Customs",
+            )
+
+            # Numbers in holders copy box
+            new_page.setFont("Helvetica", 14)
```
mikeylockwood, line 174: Doesn't matter but you can you `setFontSize` if you're not changing the font style

web/static/web/css/pdfs/sanctions-licence-style.css
```diff
@@ -1,118 +1,130 @@
...
     size: A4;
-    margin: 48px 36px 60px 72px;
-
+    margin: 68px 40px 60px 70px;
 
-    @top-left {
+    @top-center {
         content: element(header);
+        width: 100%;
     }
```
mikeylockwood, line 8: As mentioned, this currently isn't supported in browsers. Probably best to remove it in case it does someday get supported and adds extra elements to the documents

mikeylockwood (approved):

web/templates/pdf/import/sanctions-licence.html
marcuspp: I will add a ticket to look at maybe creating a macro / shared partial template that they can share and leave it for now.

PR #1663, opened 2024-07-22
ICMSLST-2747 - Complete OIL Licence PDF Document

chris-pettinga:


web/utils/pdf/pages.py
```diff
@@ -190,3 +189,13 @@ def format_dfl_pages(pdf_data: bytes, context: dict[str, Any]) -> bytes:
...
+
+
+def format_dfl_pages(pdf_data: bytes, context: dict[str, Any]) -> bytes:
+    """Function to draw the top/bottom border lines on DFL licenses"""
+    return format_firearm_license_pages(pdf_data, context)
+
+
+def format_oil_pages(pdf_data: bytes, context: dict[str, Any]) -> bytes:
+    """Function to draw the top/bottom border lines on OIL licenses"""
+    return format_firearm_license_pages(pdf_data, context)
```
mikeylockwood, line 201: Need to be careful with this if the use the same format function but their css is independent of each other. We should make sure we add visual regression tests for all licence and certificates that contain page breaks (probably in a new ticket)

chris-pettinga: @mikeylockwood yes agreed, right now they're the same but could change in the future so that's why I put them in their own functions.

re. the 2nd point I agree, the benchmark PDFs / PDFs used to test should be very long. I"ll create a ticket now

chris-pettinga: @mikeylockwood  - https://uktrade.atlassian.net/browse/ICMSLST-2791

mikeylockwood (approved):

mikeylockwood:
Tested single page and different page breaks around the signature block. All looked fine to me

PR #1664, opened 2024-07-22
General CSP fixes

chris-pettinga:
Removing HTTPS:// and www. from the directives to make them more accepting

Removing all onclick handlers

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -471,10 +471,10 @@ CSP_DEFAULT_SRC = ("'self'",)
 CSP_SCRIPT_SRC = (
     "'self'",
     "'unsafe-eval'",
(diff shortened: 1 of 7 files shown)


web/templates/web/utils/show-hide-modal-event-handler.html
```diff
@@ -0,0 +1,20 @@
+<script nonce="{{ request.csp_nonce }}">
```
mikeylockwood, line 1: Why is this not in a js file?

chris-pettinga: @mikeylockwood  - yeah that's a good point, I have no idea lol it's still early. Have changed to JS

mikeylockwood (approved):

PR #1673, opened 2024-07-23
ICMSLST-2755 -  Migrate Reports

marcuspp:


data_migration/management/commands/config/run_order/files.py
```diff
@@ -372,6 +372,16 @@
             "secure_lob_ref_id": DEFAULT_SECURE_LOB_REF_ID,
         },
     ),
+    QueryModel(
+        queries.schedule_reports,
+        "Schedule Reports",
```
marcuspp, line 377: Not sure about this one being here as these aren't files

data_migration/queries/reports.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+
+
+schedule_reports = """
+SELECT
+    rs.id AS id,
+    rr."DOMAIN" AS report_domain,
+    rs.title,
+    rs.DESCRIPTION AS notes,
+    rr.PARAMETER_XML as parameters_xml,
```
marcuspp, line 33: The xml is no longer needed to retrieve the parameters but thought leaving it here for reference was sensible.

web/reports/utils.py
```diff
@@ -116,6 +126,8 @@ def format_parameters_used(schedule_report: ScheduleReport) -> dict[str, str]:
                 if value
                 else "All"
             )
+        elif desc == "is_legacy_report":
+            value = "Yes" if value else "No"
```
mikeylockwood, line 130: SO is_legacy_report won't be populated or will be False for V2 generated reports?

data_migration/management/commands/config/run_order/files.py
mikeylockwood: I think it's fine there

mikeylockwood (approved):

mikeylockwood (approved):

PR #1678, opened 2024-07-24
ICMSLST-2746 SIL Licence PDF Styling

mikeylockwood:


web/static/web/css/pdfs/fa-sil-licence-style.css
```diff
@@ -0,0 +1,242 @@
...
+    margin-top: 40px;
+    margin-bottom: 40px;
+}
+
+@page:first {
+    margin-top: 5px;
+}
+
+* {
+    color: black;
```
marcuspp, line 14: No mention of fonts. Won't this override the `*` setting in `fonts.css`.

web/templates/pdf/import/fa-dfl-licence.html
```diff
@@ -10,9 +10,11 @@
     <style>
       {{ get_css_rules_as_string("web/css/pdfs/fonts.css")|safe }}
       {{ get_css_rules_as_string("web/css/pdfs/constants.css")|safe }}
-    </style>
-    <style>
       {{ get_css_rules_as_string("web/css/pdfs/fa-dfl-licence-style.css")|safe }}
+      {% if paper_licence_only %}
+        {{ get_css_rules_as_string("web/css/pdfs/fa-paper-licence-style.css")|safe }}
+        {{ get_css_rules_as_string("web/css/pdfs/fa-paper-licence-quantity-table-style.css")|safe }}
+      {% endif %}
```
chris-pettinga, line 17: Can you make these files specific to SIL, I know it means we'll have a bunch of probably identical CSS files for each application but it means we're certain that changes down the line to one won't affect others

mikeylockwood: The licence quantity table is a separate component on a separate page and isn't going to change. More likely it'll just be removed, I'd rather not make a separate component for it. `fa-paper-licence-style` has to override styles in the header, and again I would expect this to be removed entirely rather than changed. I would instead encourage adding visual regression test to catch any unwanted changes.

More importantly though, the SIL, DFL and OIL are all subtypes of the firearms and ammunition licence and they should never diverge; a change in one will be a change in all. I don't think they should separate at all.

chris-pettinga: @mikeylockwood so long as the regression tests include paper licenses (which we plan to) then it should be okay, I just think 'they should never diverge' is a risky assumption

PR #1682, opened 2024-07-26
ICMSLST-2858 - Sanctions licence PDF -> Styling bugs

chris-pettinga:
Also included the feedback from the doc here - https://dbis.sharepoint.com/:w:/r/sites/dit253/_layouts/15/Doc.aspx?sourcedoc=%7B5C06AA06-06EB-4EDB-A659-4899C1F3720C%7D&amp=&file=Sanctions%20and%20Adhoc%20PDF%20Formatting%20Issues.docx&action=default&mobileredirect=true&wdsle=0

### DIFF
--- a/web/templates/pdf/import/sanctions-licence.html
+++ b/web/templates/pdf/import/sanctions-licence.html
@@ -43,15 +43,12 @@
             <p class="title"><strong>3. </strong>Proposed place and date of import</p>
             
           <div class="flex-row space-between">
-               <p style="align-self: flex-start">Any UK Port</p>
-               <p style="align-self: flex-end">{{ licence_start_date }}</p>
+               <p class="flex-start">Any UK Port</p>
+               <p id="licence-start-date">{{ licence_start_date }}</p>
           </div>
           </div>
           <div class="content-block row flex-row wrap">
             <p class="row title"><strong>4. </strong>Authority responsible for issue</p>
-            <p class="address-line row">Department for Business and Trade</p>
-            <p class="address-line row">Queensway House, West Precinct</p>
-            <p class="address-line row">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
             <p class="row min-height-40">Email: {{ sanctions_contact_email }}</p>
           </div>
         </div>
@@ -121,7 +118,7 @@
                   {{ document_signature_placeholder(signature_file, signature.content_type, signature.signatory)  }}
               {% endif %}
           </div>
-          <div class="dbt-logo-block half">
+          <div class="half">
               {{ dbt_logo(classes="header-block") }}
           </div>
         </div>
--- a/web/utils/pdf/pages.py
+++ b/web/utils/pdf/pages.py
@@ -220,6 +220,45 @@ def format_sanctions_pages(pdf_data: bytes, context: dict[str, Any]) -> bytes:
         packet = io.BytesIO()
         new_page = canvas.Canvas(packet, pagesize=portrait(A4))
 
+        # holders copy box
+        new_page.setLineWidth(0.5)
+
+        # Numbers in holders copy box
+        new_page.setFontSize(14)
+        new_page.drawRightString(left_margin - 8, 770, "1")
+        new_page.drawRightString(left_margin - 8, 489, "1")
+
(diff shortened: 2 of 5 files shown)


web/templates/pdf/import/sanctions-licence.html
```diff
@@ -44,15 +44,12 @@
...
+               <p style="align-self: flex-end;margin-right:35px;">{{ licence_start_date }}</p>
           </div>
           </div>
           <div class="content-block row flex-row wrap">
             <p class="row title"><strong>4. </strong>Authority responsible for issue</p>
-            <p class="address-line row">Department for Business and Trade</p>
-            <p class="address-line row">Queensway House, West Precinct</p>
-            <p class="address-line row">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
-            <p class="row min-height-40">Email: {{ ilb_contact_email }}</p>
+            <p class="row min-height-40">Email: <a href="mailto:importcontrols@businessandtrade.gov.uk">importcontrols@businessandtrade.gov.uk</a></p>{# /PS-IGNORE #}
```
mikeylockwood, line 52: I don't think this should be hardcoded in. We can just change the value in the test environment if they really want to see it in the test pdfs

chris-pettinga: @mikeylockwood unless I'm mistaken this is the only place that this specific email address is used? If so then it feels a bit unnecessary to do the whole env var rigamarole

MattHolmes123: elsewhere we do this: `<a href="mailto:{{ ilb_contact_email }}">{{ ilb_contact_email }}</a>`
Might as well be a context variable reading from a setting in my book

chris-pettinga: @MattHolmes123 @mikeylockwood this email address is different from the typical value set by `ilb_contact_email` so it would mean the inclusion of another env var

marcuspp: This has a separate ticket which i am doing today and will create a settings for it.

pii-ner-exclude.txt
```diff
@@ -4862,3 +4862,4 @@ hide_modal_buttons[i].addEventListener('click
 Registered Number
 class="paper-header"><strong
 div class="mb-0-5
+importcontrols@businessandtrade.gov.uk">importcontrols@businessandtrade.gov.uk</a></p
```
marcuspp, line 4865: Remove as this no longer part of the PR

web/templates/pdf/import/sanctions-licence.html
```diff
@@ -44,14 +44,11 @@
             
           <div class="flex-row space-between">
                <p style="align-self: flex-start">Any UK Port</p>
-               <p style="align-self: flex-end">{{ licence_start_date }}</p>
+               <p style="align-self: flex-end;margin-right:35px;">{{ licence_start_date }}</p>
```
marcuspp, line 47: Change to class functions as these are the only two inline styles and this file has its only dedicated css so we should use it

web/utils/pdf/pages.py
```diff
@@ -220,6 +220,47 @@ def format_sanctions_pages(pdf_data: bytes, context: dict[str, Any]) -> bytes:
...
+        new_page.rotate(90)
+        new_page.setFontSize(12)
+        new_page.drawString(610, -45, "Holder's copy")
+        new_page.restoreState()
+
+        # holders copy box
+        new_page.setLineWidth(0.4)
+        # holders copy box - left vertical line. Margin is different on first page
+        if page_number == 0:
+            new_page.line(left_margin - 25, 791, left_margin - 25, 467)
```
marcuspp, line 239: I think this would read better if it was 
```
top_margin = 794
if page_number == 0:
    top_margin = 791
```    
Less if statements

marcuspp (changes requested):
The holders box is no longer in line at the bottom with the start of section 10.

web/utils/pdf/pages.py
chris-pettinga: @marcuspp good idea thank you

marcuspp (approved):

PR #1684, opened 2024-07-29
ICMSLST-2814 - Sanctions email address

marcuspp:


chris-pettinga (approved):

web/templates/pdf/import/sanctions-licence.html
```diff
@@ -52,7 +52,7 @@
             <p class="address-line row">Department for Business and Trade</p>
             <p class="address-line row">Queensway House, West Precinct</p>
             <p class="address-line row">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
-            <p class="row min-height-40">Email: {{ ilb_contact_email }}</p>
+            <p class="row min-height-40">Email: {{ sanctions_contact_email }}</p>
```
chris-pettinga, line 55: do we want to add `mailto:` here?

```suggestion
            <p class="row min-height-40">Email: <a href="mailto:{{sanctions_contact_email }}">{{ sanctions_contact_email }}</a></p>
```

web/templates/pdf/import/sanctions-licence.html
```diff
@@ -52,7 +52,7 @@
             <p class="address-line row">Department for Business and Trade</p>
             <p class="address-line row">Queensway House, West Precinct</p>
             <p class="address-line row">Billingham, United Kingdom, TS23 2NF</p>  <!-- /PS-IGNORE -->
```
mikeylockwood, line 54: We should review the addresses to make sure they're correct. I was asked to change the firearms paper licence one to darlington. Sanctions team may be different though

marcuspp: It is different

mikeylockwood (approved):

PR #1696, opened 2024-08-02
ICMSLST-2840 - GMP Application > Manual Address Entry > Character Limit

chris-pettinga:
Before:
image

After:
image


mikeylockwood (approved):
Checked the max length of the data in prod and it's 114 so happy for it to be a model change 👍

web/domains/case/export/forms.py
```diff
@@ -731,7 +731,9 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        if self.instance.pk:
+        # if the instance is already saved, and the form is being loaded with no data, then we need to disable the
+        # address fields if they are set to SEARCH
+        if self.instance.pk and not self.data:
```
mikeylockwood, line 736: ?

mikeylockwood: Sorry I get it, it fixes the form error disabling the address box

PR #1707, opened 2024-08-07
ICMSLST-2908 - Import V1 files to S3 created_datetime fix

marcuspp:


data_migration/management/commands/import_v1_files_to_s3.py
```diff
@@ -153,11 +154,11 @@ def save_run_data(
             number_of_files_processed, data_dict["number_of_files_to_be_processed"]
         ):
             data_dict["number_of_files_processed"] = number_of_files_processed
+            data_dict[limited_by_field] = last_file_processed[limited_by_field.upper()]
```
marcuspp, line 157: `limited_by_field`  for `Supplementary Report Goods Uploaded Files` is now `created_datetime` which caused the issue. Moving this line before the following one. Means the date format is properly set

MattHolmes123 (approved):

PR #1712, opened 2024-08-08
ICMSLST-2846 Prevent any form posting twice.

MattHolmes123:


web/static/web/js/components/form-double-submit.js
```diff
@@ -25,8 +25,18 @@ const blockDoubleSubmit = function(event) {
...
-    let elements = this.document.getElementsByClassName(doubleSubmitClass.formSelector);
 
+    // Add doubleSubmitClass.formSelector to all forms with method=post
+    let forms = this.document.getElementsByTagName("form")
+    Array.from(forms).forEach(function(el) {
+        // Form.method is normalized, i.e. <form method='POST'> === 'post'.
+        if (el.method === 'post' && !el.classList.contains(doubleSubmitClass.formSelector)) {
+            el.classList.add(doubleSubmitClass.formSelector)
+        }
+    });
```
mikeylockwood, line 36: Find every form, add a class to it if it doesn't have class already and then add a double submit blocker to all forms with that class? The class seems a bit redundant.

MattHolmes123: I suppose we're now dynamically adding the class we might as well directly add the eventhandler instead.
I guess I just wanted to leave it working how it was before.
I will simplify it

mikeylockwood (approved):

PR #1714, opened 2024-08-08
ICMSLST-2846 Add better error handling when a user presses back button.

MattHolmes123:
image
image

### DIFF
--- a/web/views/views.py
+++ b/web/views/views.py
@@ -13,7 +13,7 @@ from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db.models import QuerySet
 from django.http import HttpRequest, HttpResponse, HttpResponseForbidden
 from django.shortcuts import redirect, render
-from django.urls import reverse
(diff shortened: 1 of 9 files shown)


web/views/views.py
```diff
@@ -214,13 +216,85 @@ def handler403_capture_process_error_view(
...
+    """Returns True if the status / view_name can be safely ignored.
+
+    This doesn't capture every view possible, just the views that change the status.
+    e.g. a user presses submit and then presses back.
+
+    :param exception: ProcessStatusError instance raised
+    :param view_name: name of the view being requested
+    """
+
+    # Importer / Exporter user has pressed back after pressing submit
```
mikeylockwood, line 263: Include export apps too?

mikeylockwood (approved):

web/views/views.py
MattHolmes123: Oh yer good point!
I just copied that list directly from the ticket when I first added the example.

PR #1727, opened 2024-08-16
ICMSLST-2859 Replace all naive datetime instances in web and data_migration directories.

MattHolmes123:


data_migration/tests/utils/ea_data.py
```diff
@@ -133,11 +135,11 @@
                 "GA/2022/9901",  # reference
                 "IN PROGRESS",  # status
                 2,  # created_by_id
-                dt.datetime(2022, 4, 27),  # create_datetime
-                dt.datetime(2022, 4, 27),  # created
+                dt.datetime(2022, 4, 27, tzinfo=dt.UTC),  # create_datetime
```
mikeylockwood, line 138: The data in these `_data` files represents the format the data comes from the V1 query so this should not be changed. The datetimes from V1 don't come with any timezone information.

The format should instead be fixed after the data is pulled from V1 using `adjust_icms_v1_datetime` `data_migration/utils/format.py`.  Probably need to undo the change made in [ this commit ](https://github.com/uktrade/icms/commit/412cd0f5fdc2142c870f85a914d0c76a9b7540ac) as it was done to correct the datetimes displayed in V2 frontend to match V1

MattHolmes123: "The datetimes from V1 don't come with any timezone information."
☝️ Good point I will revert this then.

Let's have a call about the second paragraph

mikeylockwood (approved):

PR #1729, opened 2024-08-19
ICMSLST-2830 Fix app withdrawal bugs relating to status refactor.

MattHolmes123:
Changes:
  - Update several places WithdrawApplication.Statuses is used.
  - Add a missing form label
  - Refactor how app withdrawals are shown.

You can no longer add multiple open withdrawal requests and can withdraw if needed:
import-a-licence_8080_case_import_2_applicant_withdraw_

### DIFF
--- a/web/templates/partial/case/export/sidebar-management.html
+++ b/web/templates/partial/case/export/sidebar-management.html
@@ -21,7 +21,7 @@
   {% endwith %}
 
   {% with total = process.withdrawals.count(),
-          processed = process.withdrawals.exclude(status="open").count() %}
+          processed = process.withdrawals.exclude(status=process.withdrawals.model.Statuses.OPEN).count() %}
     {{ icms_link(request, icms_url('case:manage-withdrawals', kwargs={'application_pk': process.pk, 'case_type': 'export'}),
                  'Application Withdrawals (' + processed|string + '/' + total|string + ')') }}
   {% endwith %}
--- a/web/templates/partial/case/import/sidebar-management.html
+++ b/web/templates/partial/case/import/sidebar-management.html
@@ -49,7 +49,7 @@
 
   {# Common import application links #}
   {% with total = process.withdrawals.count(),
-          processed = process.withdrawals.exclude(status="open").count() %}
+          processed = process.withdrawals.exclude(status=process.withdrawals.model.Statuses.OPEN).count() %}
     {{ icms_link(request, icms_url('case:manage-withdrawals', kwargs={'application_pk': process.pk, 'case_type': 'import'}),
                   'Application Withdrawals (' + processed|string + '/' + total|string + ')') }}
   {% endwith %}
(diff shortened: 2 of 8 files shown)


web/templates/partial/case/export/sidebar-management.html
```diff
@@ -21,7 +21,7 @@ <h4>Export Case</h4>
   {% endwith %}
 
   {% with total = process.withdrawals.count(),
-          processed = process.withdrawals.exclude(status="open").count() %}
+          processed = process.withdrawals.exclude(status="OPEN").count() %}
```
chris-pettinga, line 24: if you really fancy

```suggestion
          processed = process.withdrawals.exclude(status=process.withdrawals.rel.to.Statuses.OPEN).count() %}
```

web/templates/partial/case/import/sidebar-management.html
```diff
@@ -49,7 +49,7 @@ <h4>Case</h4>
 
   {# Common import application links #}
   {% with total = process.withdrawals.count(),
-          processed = process.withdrawals.exclude(status="open").count() %}
+          processed = process.withdrawals.exclude(status="OPEN").count() %}
```
chris-pettinga, line 52: same as above

chris-pettinga (approved):

web/templates/partial/case/export/sidebar-management.html
mikeylockwood: To be fair, this file is absolutely littered with hardcoded strings rather than enums

mikeylockwood: Maybe there is something we could do in jinja to make it easier to access the enums anywhere? 
```
{{ process_types().SIL }}
{{ withdraw_statuses().OPEN }}
```

mikeylockwood (approved):

web/templates/partial/case/export/sidebar-management.html
MattHolmes123: I mean I could just pass it down as context.
Will have a think now

MattHolmes123: I'm just going to set a variable in the template:
e.g.
```
{% set WithdrawStatus = withdrawals.model.Statuses %}
{{ WithdrawStatus.OPEN }}
```

MattHolmes123: > Maybe there is something we could do in jinja to make it easier to access the enums anywhere?
> 
> ```
> {{ process_types().SIL }}
> {{ withdraw_statuses().OPEN }}
> ```

I agree @mikeylockwood , its something to consider in the V3 code.
Not all enums will be required everywhere though.

web/templates/partial/case/import/sidebar-management.html
MattHolmes123: `process.withdrawals.model.Statuses.OPEN` 👍

PR #1730, opened 2024-08-20
ICMSLST-2890 - Permanently deleting all versions of files from S3 if a virus is found

chris-pettinga:


web/upload_handlers.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def file_complete(self, file_size: int) -> UploadedFile:
+        # we wrap this in a try/except and also check if the file is a FileWithVirus in case
+        # CHUNK_UPLOADER_RAISE_EXCEPTION_ON_VIRUS_FOUND is set to True for whatever reason - we need to be sure.
+        try:
+            file = super().file_complete(file_size)
+            if isinstance(file, FileWithVirus):
+                # permanently delete the virus-infected file from S3
+                self.permanently_delete_file()
```
mikeylockwood, line 24: How will the user know that a virus was found so the file was deleted?

mikeylockwood (approved):

web/upload_handlers.py
chris-pettinga: django-chunk-uploader deals with this:

```
def validate_virus_check_result(file):
    try:
        file.readline()
    except VirusFoundInFileException:
        raise ValidationError(
            _('A virus was found'),
        )
```

PR #1732, opened 2024-08-20
ICMSLST-2874 - Mailshot search -> Using wrong reference

chris-pettinga:


config/settings_local.py
```diff
@@ -10,6 +10,10 @@
 AWS_STORAGE_BUCKET_NAME = "icms.local"
 AWS_S3_ENDPOINT_URL = env.local_aws_s3_endpoint_url
 
+# we set this environment variable only in the context of the Django server process because trying to change
+# AWS_ENDPOINT_URL in your general .env file causes some real issues when you're trying to interact with actual AWS
+os.environ["AWS_ENDPOINT_URL"] = AWS_S3_ENDPOINT_URL
```
MattHolmes123, line 15: Still want to see what commands you can't run / see if you can run them outside of the icms repo before we need to commit this.

MattHolmes123 (changes requested):
Feature is fine, not sure about this:
`os.environ["AWS_ENDPOINT_URL"] = AWS_S3_ENDPOINT_URL`

chris-pettinga:
@MattHolmes123  - makes sense and not urgent, so will remove and re-visit in another PR

PR #1739, opened 2024-08-22
ICMSLST-2926 Store original descriptions on DFL, Sanctions and SIL goods models

mikeylockwood:
* Remove the _override values on Sanctions and DFL Goods models
* Add _original values on Sanctions, DFL and SIL goods models
* Add migration to populate _original values to prepare to set `null=False` (will only affect non prod environments as prod will be wiped prior to running a data migration)


web/migrations/0032_dfl_sanctions_sil_goods_store_applicant_original.py
```diff
@@ -0,0 +1,186 @@
...
+
+def set_goods_original(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    set_sanctions_goods_original(apps, db_alias)
+    set_dfl_goods_original(apps, db_alias)
+    set_sil_goods_original(apps, db_alias)
+
+
+def set_sanctions_goods_original(apps, db_alias):
+    SanctionsAndAdhocApplicationGoods = apps.get_model("web", "sanctionsandadhocapplicationgoods")
```
MattHolmes123, line 14: Why not use the class name instead of all lowercase when you call apps.get_model?
e.g. sanctionsandadhocapplicationgoods -> SanctionsAndAdhocApplicationGoods

web/migrations/0032_dfl_sanctions_sil_goods_store_applicant_original.py
```diff
@@ -0,0 +1,186 @@
...
+
+    SanctionsAndAdhocApplicationGoods.objects.using(db_alias).all().update(
+        goods_description_original=models.F("goods_description"),
+        quantity_amount_original=models.F("quantity_amount"),
+        value_original=models.F("value"),
+    )
+
+
+def set_dfl_goods_original(apps, db_alias):
+    DFLGoodsCertificate = apps.get_model("web", "sanctionsandadhocapplicationgoods")
```
MattHolmes123, line 24: This looks like a copy / paste error?

MattHolmes123 (changes requested):
.

web/migrations/0032_dfl_sanctions_sil_goods_store_applicant_original.py
mikeylockwood: Good spot

web/migrations/0032_dfl_sanctions_sil_goods_store_applicant_original.py
mikeylockwood: I always thought it needed to be lowercase. Class name would be more readable, I will test

MattHolmes123 (approved):

PR #1746, opened 2024-08-28
ICMSLST-2914 - Keep standard OIL Licence document to one page

chris-pettinga:


web/static/web/css/pdfs/import/fa-paper-licence-style.css
```diff
@@ -1,5 +1,5 @@
 @page:first {
-    margin-top: 32px;
+    margin-top: 25px;
```
marcuspp, line 2: Pinching a few pixels here means it does affect the other fa certs

mikeylockwood (approved):

PR #1754, opened 2024-08-30
ICMSLST-2912 Add user group form validation to admin site.

MattHolmes123:
If an existing group is removed:
image

If a new group is added:
image

### DIFF
--- a/web/admin.py
+++ b/web/admin.py
@@ -1,6 +1,9 @@
+from typing import Any
+
 from django import forms
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.forms import UserChangeForm
 from django.contrib.auth.models import Group, Permission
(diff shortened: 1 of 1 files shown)


web/admin.py
```diff
@@ -85,13 +88,63 @@ def has_delete_permission(
...
+    Certain groups should only be modified in ICMS directly because there is extra business logic
+    required.
+    """
+
+    # List of groups that shouldn't be modified in the Admin site.
+    object_permission_groups: list[str] = [
+        Perms.obj.constabulary.get_group_name(),
+        Perms.obj.exporter.get_group_name(),
+        Perms.obj.importer.get_group_name(),
+    ]
```
mikeylockwood, line 105: Could we just always exclude all object perms instead of a set list? Say if we added more? Object perms will always related to a specific object so will have extra logic

mikeylockwood (approved):

PR #1758, opened 2024-09-04
ICMSLST-2942 Data Migration - Include legacy licence documents without files

mikeylockwood:


web/templates/web/domains/case/search/import/case_details.html
```diff
@@ -26,7 +26,12 @@
       <div class="twelve columns"><span class="prompt north disabledLink">Licence Reference</span></div>
     </div>
     <div class="row">
-      <div class="twelve columns"><a href="{{ cs.licence_reference_link }}">{{ cs.licence_reference }}</a></div>
+      <div class="twelve columns">
+        {% if cs.legacy_case_flag %}
+          <span>{{ cs.licence_reference }}</span>
+        {% else %}
+          <a href="{{ cs.licence_reference_link }}">{{ cs.licence_reference }}</a></div>
```
MattHolmes123, line 33: There is a closing div here as part of the else clause.
Seems like a bug

MattHolmes123 (approved):

web/templates/web/domains/case/search/import/case_details.html
mikeylockwood: Ah yeah, the closing div belongs outside the if block

PR #1760, opened 2024-09-05
ICMSLST-2946 - Application Update - Summary of Changes fix

marcuspp:
Screenshot 2024-09-05 at 13 17 59

Screenshot 2024-09-05 at 13 18 33


web/domains/case/models.py
```diff
@@ -87,9 +87,9 @@ class Status(TypedTextChoices):
     request_subject = models.CharField(max_length=100, null=True)
     request_detail = models.TextField(null=True)
     response_detail = models.TextField(
-        verbose_name="Response to Request",
+        verbose_name="Summary of Changes",
         null=True,
-        help_text="Please either confirm acceptance of the updates requested or your reason for no updates being made",
+        help_text="Please enter a summary of the changes made",
     )
```
MattHolmes123, line 93: These changes were requested by ILB and have deviated from V1

https://uktrade.atlassian.net/browse/ICMSLST-2740

MattHolmes123 (changes requested):

web/templates/partial/update-request-detail.html
```diff
@@ -98,7 +98,7 @@
   <div class="row">
     <div class="three columns">
       <label class="prompt west">
-        Response to Request
+        Summary of Changes
```
MattHolmes123, line 101: Same here

pii-ner-exclude.txt
```diff
@@ -4963,3 +4963,4 @@ noreferrer">DIT
 25th September 2024
 web.support.urls
 Primary Email Address
+Django 4.2.16
```
MattHolmes123, line 4966: Guess we don't need this as it was from the migration?

mikeylockwood (approved):

PR #1782, opened 2024-09-13
ICMSLST-2822 Restrict country queryset to real countries.

MattHolmes123:


marcuspp (approved):
A test could be added for this one.

PR #1789, opened 2024-09-16
ICMSLST-2966 - Update packages

marcuspp:


.env.circleci
```diff
@@ -2,6 +2,7 @@
 # Environemnt variables not loaded by DBTPlatformEnvironment directly
 #
 COPILOT_ENVIRONMENT_NAME="local_docker"  # To ensure DBTPlatformEnvironment class is used
+ECS_CONTAINER_METADATA_URI_V4="local_docker"  # To ensure DBTPlatformEnvironment class is used
```
MattHolmes123, line 5: is this required because of updating dbt-copilot-python?

MattHolmes123 (approved):

.env.circleci
marcuspp: Correct

PR #1791, opened 2024-09-17
ICMSLST-2913 Add versioning to admin templates

mikeylockwood:
Editing a template will now create a new template version and set the previous version inactive

Templates now display which version they are on and details on who / when template was modified.

System generated test data uses the anonymous user, so created by details appear blank

image


web/domains/template/models.py
```diff
@@ -55,25 +57,31 @@ class Template(models.Model):
...
-    # everything except CFS schedules uses this; CFS schedules use "paragraphs" (see CFSScheduleParagraph, below)
-    template_content = models.TextField(blank=False, null=True)
-
     countries = models.ManyToManyField("web.Country")
     country_translation_set = models.ForeignKey(
         "web.CountryTranslationSet", on_delete=models.SET_NULL, blank=False, null=True
     )
 
+    @property
+    def current_version(self) -> Union["TemplateVersion", None]:
```
MattHolmes123, line 71: Can't you do this: `TemplateVersion | None`

MattHolmes123: ahh it's because the model comes after this line.
In that case you can define the type like this:
`"TemplateVersion | None"`

web/domains/template/models.py
```diff
@@ -96,6 +104,17 @@ class Meta:
         )
 
 
+class TemplateVersion(models.Model):
+    template = models.ForeignKey(Template, on_delete=models.CASCADE, related_name="versions")
+    start_datetime = models.DateTimeField(auto_now_add=True, blank=False, null=False)
```
MattHolmes123, line 109: `null=False` is the default and can be removed.
`blank=False` is the default and can be removed.

web/domains/template/models.py
```diff
@@ -96,6 +104,17 @@ class Meta:
...
+    end_datetime = models.DateTimeField(blank=True, null=True)
+    is_active = models.BooleanField(blank=False, null=False, default=True)
+    version_number = models.PositiveIntegerField(default=1)
+    title = models.CharField(max_length=4000, blank=False, null=True)
+    content = models.TextField(blank=False, null=True)
+    created_by = models.ForeignKey("web.User", on_delete=models.PROTECT)
```
MattHolmes123, line 115: This should use this: `settings.AUTH_USER_MODEL`

MattHolmes123: I refactored all the models ages ago but a couple have snuck back in.

web/templates/web/domains/template/detail.html
```diff
@@ -20,6 +20,12 @@
 {% endblock %}
 
 {% block main_content %}
+  {% if object and object.current_version %}
+    <span class="extra-info">
+      Currently viewing version {{ object.current_version.version_number }}.
+      Last updated on {{ object.current_version.start_datetime.strftime('%d-%b-%Y %H:%M:%S') }} by {{ object.current_version.created_by }}.
```
MattHolmes123, line 26: We should never format datetime instances directly in templates.
All datetimes should use `|datetime_format(` to correctly localise the datetime to the user.

web/templates/web/domains/template/edit-cfs-declaration-translation.html
```diff
@@ -2,5 +2,9 @@
 
 {% block main_content %}
   <h3>Edit '{{ object.template_name }}' CFS Declaration Translation</h3>
+  <span class="extra-info">
+    Currently editing version {{ object.current_version.version_number }}.
+    Last updated on {{ object.current_version.start_datetime.strftime('%d-%b-%Y %H:%M:%S') }} by {{ object.current_version.created_by }}.
```
MattHolmes123, line 7: The lines you have added use the default format so no need to pass the format.

MattHolmes123 (approved):

web/domains/template/views.py
```diff
@@ -116,6 +122,23 @@ def view_cfs_schedule_translation(request, pk):
...
+        current_version.is_active = False
+        current_version.save()
+
+    TemplateVersion.objects.create(
+        template=template,
+        created_by=user,
+        title=title,
+        content=content,
+    )
+
```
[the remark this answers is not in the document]
mikeylockwood: It does need more test coverage following the changes. I will add a ticket for this

web/domains/template/views.py
```diff
@@ -116,6 +122,23 @@ def view_cfs_schedule_translation(request, pk):
...
+    if current_version:
+        current_version.end_datetime = timezone.now()
+        current_version.is_active = False
+        current_version.save()
+
+    TemplateVersion.objects.create(
+        template=template,
+        created_by=user,
+        title=title,
+        content=content,
```
marcuspp, line 138: As mentioned add in the version number

marcuspp (approved):

PR #1799, opened 2024-09-18
ICMSLST-2949 - Legacy Report parameters missing

marcuspp:


data_migration/models/report.py
```diff
@@ -33,18 +33,32 @@ def data_export(cls, data: dict[str, Any]) -> dict[str, Any]:
...
         if data["deleted_by_id"]:
             data["status"] = "DELETED"
             data["deleted_by_id"] = ANONYMOUS_USER_PK
         return data
 
+    @classmethod
+    def filter_parameters(cls, data: dict[str, Any], report_id: int) -> dict[str, Any]:
+        parameters = {"is_legacy_report": True}
+        for key, value in data.items():
+            if key == "application_type" and report_id in [1, 3]:
```
MattHolmes123, line 52: will values `1` and `3` benefit from being made in to constants?
I have no idea what they are.

MattHolmes123 (approved):

data_migration/models/report.py
marcuspp: Actually i will change it to pass in the report domain `IMP_ISSUED_CERTIFICATES` and `IMP_DATA_EXTRACT` to improve readability

PR #1802, opened 2024-09-18
ICMSLST-2956 - Firearms licence missing column

marcuspp:


mikeylockwood (approved):

web/reports/interfaces.py
```diff
@@ -436,7 +436,8 @@ def serialize_row(self, ia: dict) -> BaseModel:
             eori_number=ia["importer_eori_number"],
             importer_address=utils.format_importer_address(ia),
             first_submitted_date=ia["submit_datetime"].date(),
-            final_submitted_date=ia["last_submit_datetime"].date(),
+            final_submission_date=ia["last_submit_datetime"].date(),
+            licence_authorisation_date=licence.start_date,
```
MattHolmes123, line 440: This will output incorrect values as it doesn't convert to local time first.
e.g. using datetime_format as an example:
```
>>> import datetime as dt
>>> from web.utils import datetime_format
>>> val = dt.datetime(2024, 8, 3, 23, 30, tzinfo=dt.UTC)
>>> val.date()
datetime.date(2024, 8, 3)
>>> datetime_format(val)
'04-Aug-2024 00:30:00'
```

MattHolmes123: Whenever we show datetime instances to the user we need to localise them first.

MattHolmes123: As your code requires a date in the serializer (instead of a str that datetime_format returns) you can do the following:
```python
>>> from django.utils import timezone
>>> localtime_val = timezone.localtime(val)
>>> localtime_val.date()
datetime.date(2024, 8, 4)
```

MattHolmes123: Just realised my "Comment on lines..." is wrong.
I am referring to this line: `final_submission_date=ia["last_submit_datetime"].date(),`

PR #1804, opened 2024-09-19
Template Version bug fixes

mikeylockwood:


web/domains/template/views.py
```diff
@@ -193,7 +195,7 @@ def form_valid(self, form):
         """
         Sets readonly fields for this template type and validates other inputs.
         """
-        template = form.instance
+        template = form.save()
```
mikeylockwood, line 198: Needs to be a template object created to be able to set a fk to it in `TemplateVersion`

marcuspp (approved):

PR #1805, opened 2024-09-19
ICMSLST-2941 Display agent info when viewing case.

MattHolmes123:
Import Application showing agent data:
image

Export Application showing agent data:
image

I kept the `Agent -` prefix in the workbasket after removing it from the model `__str__` methods:
image

### DIFF
--- a/web/templates/web/domains/case/export/partials/process.html
+++ b/web/templates/web/domains/case/export/partials/process.html
@@ -52,10 +52,14 @@
     <span>{{ process.get_status_display() }}</span>
   </div>
   <div class="two columns">
-    <label class="prompt west contextItemName">Exporter Address</label>
+    {% if process.agent %}
+      <label class="prompt west contextItemName">Agent</label>
+    {% endif %}
   </div>
   <div class="two columns">
-    <span>{{ process.exporter_office.address }} {{ process.exporter_office.postcode or "" }}</span>
+    {% if process.agent %}
+      <span>{{ process.agent }}</span>
+    {% endif %}
   </div>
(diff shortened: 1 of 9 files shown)


web/templates/web/domains/case/export/partials/process.html
```diff
@@ -52,10 +52,14 @@
     <span>{{ process.get_status_display() }}</span>
   </div>
   <div class="two columns">
-    <label class="prompt west contextItemName">Exporter Address</label>
+    {% if process.agent %}
+      <label class="prompt west contextItemName">Agent</label>
+    {% endif %}
   </div>
   <div class="two columns">
-    <span>{{ process.exporter_office.address }} {{ process.exporter_office.postcode or "" }}</span>
```
MattHolmes123, line 58: This previous value was something I decided to add and isn't in V1. 
It should have always been the agent.

web/templates/web/domains/case/import/partials/process.html
```diff
@@ -60,10 +60,14 @@
     </span>
   </div>
   <div class="two columns">
-    <label class="prompt west contextItemName">Importer EORI</label>
+    {% if process.agent %}
+      <label class="prompt west contextItemName">Agent</label>
+    {% endif %}
   </div>
   <div class="two columns">
-    <span>{{ process.importer.eori_number|default("N/A", True) }}</span>
```
MattHolmes123, line 66: This previous value was something I decided to add and isn't in V1. 
It should have always been the agent.

mikeylockwood (approved):

PR #1824, opened 2024-09-27
ICMSLST-2987 Show access request response and response reason.

MattHolmes123:
image
image

### DIFF
--- a/web/templates/web/domains/case/access/access-request-history.html
+++ b/web/templates/web/domains/case/access/access-request-history.html
@@ -1,4 +1,6 @@
 {% extends "layout/no-sidebar.html" %}
+{% from "display/fields.html" import application_field %}
+
 
 {% block main_content %}
   <!-------------------->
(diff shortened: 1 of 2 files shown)


mikeylockwood (approved):

web/templates/web/domains/case/access/access-request-history.html
```diff
@@ -8,6 +10,12 @@
     {% include "web/domains/case/access/partials/access-request-section.html" %}
   {% endwith %}
 
+  {# Extra access request fields to show when viewing a historical record. #}
+  {{ application_field(access_request.get_response_display(), "Response") }}
+  {% if access_request.response == access_request.REFUSED %}
+    {{ application_field(access_request.response_reason, "Response Reason") }}
```
mikeylockwood, line 16: Should this be "Refuse Reason"?

MattHolmes123: Yes probably 👍

PR #1827, opened 2024-09-30
ICMSLST-2984 Add case email workbasket message.

MattHolmes123:


mikeylockwood (approved):

web/migrations/0037_alter_exportapplication_case_emails_and_more.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.2.16 on 2024-09-30 15:59
```
mikeylockwood, line 1: It's nice to rename the django migrations to something more useful than the autogenerated name. I use `--name` when generating but can just rename the file afterwards

web/domains/case/_import/models.py
```diff
@@ -247,7 +247,7 @@ class Meta:
     update_requests = models.ManyToManyField("web.UpdateRequest")
     case_notes = models.ManyToManyField("web.CaseNote")
     commodity_group = models.ForeignKey("web.CommodityGroup", on_delete=models.PROTECT, null=True)
-    case_emails = models.ManyToManyField("web.CaseEmail", related_name="+")
+    case_emails = models.ManyToManyField("web.CaseEmail", related_name="import_applications")
```
mikeylockwood, line 250: It's a bit strange to pluralise import_application as there can only possibly be one import application per case_email. I know it does return a queryset however due to the M2M

MattHolmes123: I'm just following djangos convensions:
image

and like you said it's a M2M (even though we know there will be only one app)

web/migrations/0037_alter_exportapplication_case_emails_and_more.py
MattHolmes123: renamed

PR #1834, opened 2024-10-02
ICMSLST-2986 Add endpoint to link agents to access requests.

MattHolmes123:
Manage Access Request page with importer and agent foms:
caseworker_8080_access_case_11_importer_link-access-request_

Close access request message when agent hasn't been linked:
caseworker_8080_access_case_11_importer_close-access-request_

Manage Access Request page with complete approval showing agent can still be set after approval:
caseworker_8080_access_case_11_importer_link-access-request_ (1)

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -2,14 +2,19 @@ from typing import Any, Literal
 
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required, permission_required
-from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.contrib.auth.mixins import (
+    LoginRequiredMixin,
+    PermissionRequiredMixin,
+    UserPassesTestMixin,
+)
 from django.db import transaction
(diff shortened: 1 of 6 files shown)


web/domains/case/access/views.py
```diff
@@ -237,9 +244,14 @@ def link_access_request(
                 )
 
             form = form_cls(request.POST, instance=access_request)
-
+            original_link = copy.copy(access_request.link.pk) if access_request.link else None
```
mikeylockwood, line 247: Why do you need to use `copy` to copy a pk?

mikeylockwood (approved):

web/domains/case/access/views.py
MattHolmes123: Ahh I don't anymore! 
I had that from when I was comparing records (the form instance is the same object as the returned form.save() value) so it was picking up the new value everytime.
Good spot.

PR #1835, opened 2024-10-03
ICMSLST-3005 CFS Spreadsheet Upload less strict for non biocide products

mikeylockwood:
Made the spreadsheet uploaded less strict
- Doesn't require a sheet name. Will just take the first sheet if there are no sheets named "CFS Products"
- Will skip blank rows rather than raising a validation error
- As long as the first column is "Product Name" for non biocide, it will upload the first column of data regardless of whats in the other columns

Added a message to the screen to use the .xlsx template provided

image

### DIFF
--- a/web/domains/case/export/utils.py
+++ b/web/domains/case/export/utils.py
@@ -127,18 +127,21 @@ def _extract_sheet_header(sheet: Worksheet, is_biocidal: bool = False) -> list[s
 
     header = _get_header(is_biocidal)
     header_data: list[str] = list(next(sheet.values))  # type:ignore[arg-type]
-
     header_row = [col.strip() if col else "" for col in header_data]
 
-    # Check the number of columns with data in xlsx file matches the length of the header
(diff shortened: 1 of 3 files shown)


web/domains/case/export/utils.py
```diff
@@ -127,18 +127,21 @@ def _extract_sheet_header(sheet: Worksheet, is_biocidal: bool = False) -> list[s
...
+            )
+        # Check the header in the xlsx file matches the expected template header
+        if header != header_row:
+            raise ValidationError("Spreadsheet header does not match the template header")
 
-    # Check the header in the xlsx file matches the expected template header
-    if header != header_row:
-        raise ValidationError("Spreadsheet header does not match the template header")
+    else:
+        if header[0:1] != header_row:
```
MattHolmes123, line 143: `header[:1]`

MattHolmes123 (approved):

PR #1842, opened 2024-10-08
ICMSLST-2977 - Add access requests missing fir files

marcuspp:


data_migration/management/commands/add_access_request_fir_files.py
```diff
@@ -0,0 +1,151 @@
...
+        """Finds the relevant access request fir in the V2 database add adds the fir file attachment data."""
+        access_request_pk = dm.AccessRequest.objects.get(iar_id=row["ACCESS_REQUEST_ID"]).pk
+        ar_data = dict(reference=row["REQUEST_REFERENCE"], pk=access_request_pk)
+        if self.dry_run:
+            self.stdout.write(json.dumps(ar_data, indent=2, cls=DjangoJSONEncoder))
+        access_request = AccessRequest.objects.get(**ar_data)
+
+        fir_data = dict(
+            request_subject=row["FIR_REQUEST_SUBJECT"],
+            requested_datetime=adjust_icms_v1_datetime(row["FIR_REQUESTED_DATETIME"]),
```
marcuspp, line 132: Is the `adjust_icms_v1_datetime` required?

MattHolmes123: If you're extracting v1 data to save in to a V2 model I would say so.

```
def adjust_icms_v1_datetime(dt_val: dt.datetime) -> dt.datetime:
    """Adjust an ICMS V1 naive datetime to an aware UTC datetime.

    Assumption: For ambiguous datetime values we are using the default fold of 0.
    E.g. We have no way of knowing if this `naive_dt` is GMT or BST
    >>> naive_dt = dt.datetime(2022, 10, 30, 1, 30, 0)
    >>> london = zoneinfo.ZoneInfo("Europe/London")
    >>> bst_val = naive_dt.replace(tzinfo=london, fold=0)
    >>> gmt_val = naive_dt.replace(tzinfo=london, fold=1)
    >>> print(bst_val)
        2022-10-30 01:30:00+01:00
    >>> print(gmt_val)
        2022-10-30 01:30:00+00:00
    >>> print(bst_val.astimezone(dt.UTC))
        2022-10-30 00:30:00+00:00
    >>> print(gmt_val.astimezone(dt.UTC))
        2022-10-30 01:30:00+00:00
    """
```

mikeylockwood (approved):

PR #1844, opened 2024-10-09
ICMSLST-3016 Allow users to search for "." character.

MattHolmes123:
image

### DIFF
--- a/web/domains/case/forms_search.py
+++ b/web/domains/case/forms_search.py
@@ -25,8 +25,8 @@ from web.utils.search import get_export_status_choices, get_import_status_choice
 # % (The wildcard character)
 # \s Any whitespace
 # \w Matches any alphanumeric character; this is equivalent to the class [a-zA-Z0-9_].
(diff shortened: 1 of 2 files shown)


web/domains/case/forms_search.py
```diff
@@ -26,7 +26,7 @@
 # \s Any whitespace
 # \w Matches any alphanumeric character; this is equivalent to the class [a-zA-Z0-9_].
 # Any of the following symbols: ' " , / & @ - \ ( )
```
marcuspp, line 28: Comment missing `.`

marcuspp (approved):
Any tests for this that could be updated?

MattHolmes123:
> Any tests for this that could be updated?

When it used to use [__regex](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#regex) under the hood I would have added a test for this.

There are already many tests for the wildcard search and it uses the `__ilike` field lookup now which is way simpler so didn't feel the need to add more tests.

PR #1845, opened 2024-10-09
ICMSLST-3015 Add/change bought from details when app is out for update

mikeylockwood:


web/domains/case/_import/fa/views.py
```diff
@@ -124,9 +125,19 @@ def create_import_contact(
...
+                    application, [ImpExpStatus.IN_PROGRESS, ImpExpStatus.PROCESSING]
+                )
+            except errors.ProcessStatusError as e:
+                # Check if application is out for update with ownership released
+                if not (
+                    application.status == ImpExpStatus.SUBMITTED
+                    and not application.case_owner
+                    and application.current_update_requests().exists()
+                ):
+                    raise e
```
MattHolmes123, line 139: Can you do this check in a single function in this file please?
I will be refactoring it soon enough anyway but no need for the duplication

MattHolmes123: Actually Michael... can it just use this: `case_progress.application_in_progress(application)`

MattHolmes123: image

MattHolmes123 (approved):

web/domains/case/_import/fa/views.py
```diff
@@ -234,11 +231,24 @@ def delete_import_contact(
...
-        )
-
+        try:
+            case_progress.check_expected_status(
+                application,
+                [
+                    ImpExpStatus.IN_PROGRESS,
+                    ImpExpStatus.PROCESSING,
+                    ImpExpStatus.VARIATION_REQUESTED,
+                    ImpExpStatus.COMPLETED,
```
mikeylockwood, line 241: This also has a requirement fro completed as the view is used in the supplementary report

PR #1848, opened 2024-10-10
ICMSLST-2867 - DFL UI Improvements

marcuspp:


web/templates/partial/know-bought-from.html
```diff
@@ -1,6 +1,6 @@
-{% set display_warning = process.know_bought_from == False %}
-
-<div id="know-bought-from-info-box" class="info-box info-box-info" style="display: {{ 'block' if display_warning else 'none' }}">
-  You will be asked to provide this information after the import licence has been issued.
-  An export authority cannot issue a firearms export licence until this information has been provided.
-</div>
+{% if not process.know_bought_from and process.known_bought_from != False %}
```
mikeylockwood, line 1: There is `know_bought_from` and `known_bought_from` in this condition. Is that right?

marcuspp: Good spot! i'll amend

web/templates/partial/know-bought-from.html
```diff
@@ -1,6 +1,6 @@
...
-<div id="know-bought-from-info-box" class="info-box info-box-info" style="display: {{ 'block' if display_warning else 'none' }}">
-  You will be asked to provide this information after the import licence has been issued.
-  An export authority cannot issue a firearms export licence until this information has been provided.
-</div>
+{% if not process.know_bought_from %}
+  <div id="know-bought-from-info-box" class="info-box info-box-warning" style="display: {{ 'block' if process.know_bought_from == False else 'none' }}">
+    You will be asked to provide this information after the import licence has been issued.
+    An export authority cannot issue a firearms export licence until this information has been provided
+  </div>
+{% endif %}
```
MattHolmes123, line 6: This won't work with this now will it:
`icms/web/static/web/js/pages/bought-from.js`
As the template isn't always included the javascript won't trigger?

PR #1849, opened 2024-10-10
ICMSLST-3014 Refactor function views to use case_progress functions.

MattHolmes123:


web/domains/case/_import/views.py
```diff
@@ -494,11 +494,6 @@ def bypass_chief(
             ImportApplication.objects.select_for_update(), pk=application_pk
         )
 
-        case_progress.check_expected_status(
-            application,
-            [ImpExpStatus.PROCESSING, ImpExpStatus.VARIATION_REQUESTED, ImpExpStatus.REVOKED],
-        )
-
```
MattHolmes123, line 501: I have just removed this as it's only the bypass chief link

marcuspp (approved):

PR #1851, opened 2024-10-10
ICMSLST-3014 Refactor several places ApplicationTaskMixin is used.

MattHolmes123:


marcuspp (approved):
Some Mixins have descriptions others don't. It would be good if this was consistent

PR #1854, opened 2024-10-11
ICMSLST-2874 Show response prep content before approval to caseworker.

MattHolmes123:


web/templates/web/domains/case/manage/prepare-response.html
```diff
@@ -52,8 +52,8 @@ <h3>Response Preparation</h3>
     {% endcall %}
   {% endif %}
 
-
-  {% if process.application_approved %}
+  {# Show editable content to caseworker yet to approve or approved or readonly content when decision is approved #}
+  {% if (not readonly_view and process.decision in [None, process.APPROVE]) or (readonly_view and process.decision == process.APPROVE) %}
```
mikeylockwood, line 56: Could this be simplified to
```
{% if process.application_approved or (not readonly_view and process.decision is None) %}
```

mikeylockwood (approved):

web/templates/web/domains/case/manage/prepare-response.html
MattHolmes123: `== None`
`is` isn't valid jinja syntax

PR #1873, opened 2024-10-18
ICMSLST-2995 - Fixed naive datetime, excluded config from coverage

marcuspp:


.coveragerc
```diff
@@ -11,14 +11,15 @@ omit =
...
     web/end_to_end/*
     web/tests/*
+    web/migrations/*
     web/management/commands/utils/*
-    web/management/commands/add_dummy_data.py
-    web/management/commands/add_test_data.py
-    web/management/commands/countries.py
-    web/management/commands/drop_all_tables.py
-    web/management/commands/query_task_result.py
+    web/management/commands/*
```
marcuspp, line 16: Do you think that its okay to do this?

mikeylockwood: The reason I didn't do this is I added tests for management commands that I created

marcuspp: Yeah i know some have cmds have tests and others don't. Was potentially thinking of moving the cmds that load test data to a sub dir and only those commands are excluded and any command that does something real, needs to be tested?

chris-pettinga (approved):

.coveragerc
mikeylockwood: Sub dirs don't work for management commands unfortunately, else I would have already moved them

PR #1887, opened 2024-10-30
ECIL-222 Schedule EU labels for pre-brexit applications

mikeylockwood:


web/domains/case/views/views_view_case.py
```diff
@@ -332,6 +333,7 @@ def _view_cfs(
         "application_countries": app_countries,
         "schedules": application.schedules.all().order_by("created_at"),
         "page_title": get_case_page_title("export", application, "View"),
+        "eu_datetime_cutoff": dt.datetime(2021, 1, 1, tzinfo=dt.UTC),
```
MattHolmes123, line 336: You might as well just do the logic here if it's the only place used.
e.g.
`"show_pre_brexit_fields": process.submit_datetime < dt.datetime(2021, 1, 1, tzinfo=dt.UTC)`

MattHolmes123 (approved):

PR #1897, opened 2024-11-04
ECIL-274 Normalize GMP manufacturer address

mikeylockwood:
This is to fix an issue with a the manufacturer address appearing on the GMP certificate with black squares when non html characters are used. In the case of the bug it was for a fullwidth comma.

Also now making the address lines on the certificate and the space between the postcode include commas as is consistent with V1

image


web/domains/case/export/forms.py
```diff
@@ -907,3 +931,17 @@ def save(
...
+    address = address.strip()
+
+    # Normalize string to remove unwanted characters
+    address = unicodedata.normalize("NFKC", address)
+
+    # Strip trailing comma
+    address = address.rstrip(",")
+
+    # Replace comma without a following whitespace charatcer with a comma and a space
+    return re.sub(r",(?=\S)", ", ", address)
```
MattHolmes123, line 947: Should we have a more detailed comment in the docstring detailing why we are doing this?

web/tests/domains/case/export/test_forms.py
```diff
@@ -16,47 +17,102 @@
...
+        form.cleaned_data = form.data
+        form.clean()
+        assert form.cleaned_data["manufacturer_address"] == data["manufacturer_address"]
+        assert form.cleaned_data["responsible_person_address"] == data["responsible_person_address"]
+
+    def test_gmp_edit_form_normalize_manual_address(self):
+        data = self.form_data | {
+            "manufacturer_address": "Buckingham Palace，London，",
+            "responsible_person_address": "Buckingham Palace，London，",
+        }
```
MattHolmes123, line 93: I would even add an explicit comment here.
e.g. 
`# Test we can normalise the following unicode character: https://www.codetable.net/decimal/65292`

MattHolmes123 (approved):

PR #1905, opened 2024-11-07
ECIL-285 Add web-internal to local docker-compose.yml.

MattHolmes123:
Run two versions of ICMS to better reflect the future production infrastructure.
Added so that our end to end tests will run on different servers and test the dynamic urls.

### DIFF
--- a/config/settings_test.py
+++ b/config/settings_test.py
@@ -16,6 +16,9 @@ RATELIMIT_ENABLE = False
 FILE_UPLOAD_HANDLERS = ("web.tests.file_upload_handler.DummyFileUploadHandler",)  # type: ignore[assignment]
 APP_ENV = "test"
 
+# Ensure all endpoints are available when testing.
+INCLUDE_PRIVATE_URLS = True
+
 # Add so we can test the bypass chief views.
 ALLOW_BYPASS_CHIEF_NEVER_ENABLE_IN_PROD = True
 
--- a/web/menu/menu.py
+++ b/web/menu/menu.py
@@ -2,10 +2,10 @@ import logging
 from typing import Any
 
 from django.conf import settings
-from django.urls import resolve, reverse
+from django.urls import NoReverseMatch, resolve, reverse
 
 from web.permissions import Perms, can_user_view_search_cases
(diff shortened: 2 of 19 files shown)


config/settings_test.py
```diff
@@ -16,6 +16,9 @@
 FILE_UPLOAD_HANDLERS = ("web.tests.file_upload_handler.DummyFileUploadHandler",)  # type: ignore[assignment]
 APP_ENV = "test"
 
+# Ensure all endpoints are available when testing.
+IS_PRIVATE_APP = True
```
MattHolmes123, line 20: TODO: At at least one unittest that changes this value and reloads the URL config.

chris-pettinga (approved):

web/tests/test_urls.py
```diff
@@ -0,0 +1,32 @@
...
+def test_ilb_admin_client_cant_access_private_urls(ilb_admin_client):
+    reload_urlconf()
+    reload_urlconf("web.urls")
+    clear_url_caches()
+
+    # Check we can't use reverse.
+    with pytest.raises(NoReverseMatch):
+        reverse("signature-list")
+
+    response = ilb_admin_client.get("/signature/")
```
marcuspp, line 31: This is should use reverse not the hardcoded path

web/menu/menu.py
```diff
@@ -41,11 +41,16 @@ def has_view_access(
...
+                view_url = reverse(view_name)
+                view = resolve(view_url).func.view_class()
+                view.request = request
 
-            has_perm = view.has_permission()
+                has_perm = view.has_permission()
+
+            # Now some urls are only available on private app this code can throw exceptions.
+            except NoReverseMatch:
+                has_perm = False
```
marcuspp, line 53: The coverage report says this is not being hit.

marcuspp (approved):
Nice one for finding a way to test the public / private urls :+1:

web/menu/menu.py
MattHolmes123: Yes that's because all tests (apart from the one I added) run with all urls enabled.

PR #1908, opened 2024-11-08
ECIL-173 - First name added to emails

marcuspp:
Introduces a RecipientDetails object that contains the email address the users first name.
 - The logic that uses the `ICMS_SEND_ALL_EMAILS_TO` setting is moved to with the dataclass
 - Handles if a users name is not set and uses defaults if user/recipient is unknown


web/tests/helpers.py
```diff
@@ -84,6 +84,7 @@ def check_gov_notify_email_was_sent(
 
 
 def assert_common_email_personalisation(personalisation: dict, exp_subject: str, exp_in_body: str):
+    assert personalisation.pop("first_name")
```
marcuspp, line 87: Simple check for the first_name is in the personalisation data. The exact value is checked within `test_emails.py` tests

MattHolmes123 (approved):

PR #1916, opened 2024-11-13
ECIL-270 - File uploads to temp storage

marcuspp:
- Adds second temp s3 bucket to local install
 - Files initially uploaded to temp storage.
 - After virus scan gives all clear
 - Files are moved to main storage
 - Removed check for versions and their removal as no versions of files should exist


web/domains/file/models.py
```diff
@@ -24,6 +25,7 @@ class File(models.Model):
     path = models.CharField(max_length=4000)
     created_datetime = models.DateTimeField(auto_now_add=True)
     created_by = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.PROTECT)
+    clam_av_results = models.JSONField(encoder=DjangoJSONEncoder, default=list)
```
marcuspp, line 28: Not sure if this is neccessary

local_deployment/s3/lifecycle.json
```diff
@@ -0,0 +1,13 @@
+{
+    "Rules": [
+        {
+            "ID": "Delete files older than 1 day",
+            "Prefix": "",
+            "Status": "Enabled",
+            "Expiration":
+                {
+                    "Days": 1
+                }
```
MattHolmes123, line 10: Do we want to change this to a few days?
Just incase we are working on something relating to files that takes a few days

web/domains/file/models.py
MattHolmes123: It doesn't harm anything, just adds support for extra date / time types:
https://docs.djangoproject.com/en/5.1/topics/serialization/#django.core.serializers.json.DjangoJSONEncoder

local_deployment/s3/lifecycle.json
marcuspp: This only applies to the temp storage. Files will be copied to the main storage and still be there so all files applications wise will still be present so should not impact test applications.

web/upload_handlers.py
```diff
@@ -1,44 +1,32 @@
...
-        except VirusFoundInFileException:
-            self.permanently_delete_file()
-            raise
+class S3FileUploadHandler(DjangoChunkS3FileUploadHandler):
+
+    def file_complete(self, file_size: int) -> S3Boto3StorageFile:
+        file = super().file_complete(file_size)
+        file.content_type_extra = self.content_type_extra
+        if isinstance(file, S3Boto3StorageFile):
+            self.move_file_to_permanent_storage()
```
chris-pettinga, line 21: might be best to hardcode `CHUNK_UPLOADER_RAISE_EXCEPTION_ON_VIRUS_FOUND=False` so the upload handler will always return some sort of file

chris-pettinga (approved):

PR #1931, opened 2024-11-29
ECIL-201 Email Verification

marcuspp:


web/migrations/0048_email_is_verified_alter_emailtemplate_name_and_more.py
```diff
@@ -0,0 +1,167 @@
...
+    EmailTemplate.objects.using(db_alias).filter(
+        name=EmailTypes.EMAIL_VERIFICATION, gov_notify_template_id=GOV_NOTIFY_TEMPLATE_ID
+    ).delete()
+
+
+def mark_primary_email_address_as_verified(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+
+    Email = apps.get_model("web", "Email")
+    Email.objects.using(db_alias).filter(is_primary=True).update(is_verified=True)
```
marcuspp, line 36: This could maybe only set `is_verified` if the primary email address matches the user's email address

web/domains/user/views.py
```diff
@@ -168,37 +171,102 @@ def get_queryset(self) -> QuerySet[PhoneNumber]:
...
+        return qs.filter(user=self.request.user)
+
+    def post(self, request: AuthenticatedHttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        email = self.get_object()
+        send_and_create_email_verification(email, self.request.site)
+        messages.info(
+            self.request,
+            f"Email sent to {email.email}, Please check your inbox to complete adding the email address to your account",
+        )
+        return super().post(request, *args, **kwargs)
```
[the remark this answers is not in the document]
marcuspp: This requires a the user to be logged in so don't think a rate limit is required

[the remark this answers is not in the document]
MattHolmes123: Logged in can literally be anyone though.
Any bad actor can easily log in via One Login.

web/domains/user/views.py
```diff
@@ -168,37 +171,102 @@ def get_queryset(self) -> QuerySet[PhoneNumber]:
...
+        if self.object:
+            return reverse(
+                "user-email-edit",
+                kwargs={"user_pk": self.request.user.pk, "email_pk": self.object.email.pk},
+            )
+        else:
+            return reverse("user-edit", kwargs={"user_pk": self.request.user.pk})
+
+
+class SendVerifyEmailView(UserBaseMixin, SingleObjectMixin, RedirectView):
```
MattHolmes123, line 218: Need to rework this as it can leave a lot if records lying around:
image

web/migrations/0048_email_is_verified_alter_emailtemplate_name_and_more.py
MattHolmes123: Might be safer to consider this a new feature going forward though and just leave it how you did it 🤔

mikeylockwood (commented):
These objects should expire after a period of time, not just when clicked. I would changed `processed` on the `EmailVerification` to `expired`, clicking the link works in the same way where `expired` is set to True. But also clicking the link after x amount of time (2 weeks?) causes it to expire, but verification fails and a new object has to be sent out.

Timed expiry logic could be done in another ticket as long as the model changes to support it are done in this one. I would want an `expiry_datetime` field on the model whether that is set during creation or if it's a poperty on the model based off `created_at`

web/migrations/0048_email_is_verified_alter_emailtemplate_name_and_more.py
```diff
@@ -0,0 +1,168 @@
+# Generated by Django 5.1.3 on 2024-12-06 12:19
```
mikeylockwood, line 1: Rename the migration file 
`0048_user_email_verification.py` is more concise

chris-pettinga (approved):

mikeylockwood (approved):

PR #1933, opened 2024-12-05
ECIL-325 Fix error summary links for several form fields.

MattHolmes123:
Changes:
  - Fix error summary link for GovUKDateInputField.
  - Fix error summary link for GovUKCheckboxesField.
  - Fix error summary link for GovUKRadioInputField.
  - Add test for error_summary_kwargs.
  - Rename GDSMixin to GDSFieldMixin.

### DIFF
--- a/web/ecil/gds/forms/forms.py
+++ b/web/ecil/gds/forms/forms.py
@@ -47,13 +47,27 @@ class GDSFormMixin:
         if not self.errors:
             return {}
 
-        # TODO: Revisit in ECIL-325 to fix checkbox, date and radio input fields.
+        error_list = []
+
+        for field_name, error in self.errors.items():
+            match self.fields[field_name]:
+                case gds_fields.GovUKDateInputField():
+                    # Default the date input link to the first field.
+                    href = f"#id_{field_name}_0"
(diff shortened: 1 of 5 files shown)


web/ecil/gds/forms/forms.py
```diff
@@ -47,13 +47,27 @@ def error_summary_kwargs(self) -> dict[str, Any]:
...
+            match self.fields[field_name]:
+                case gds_fields.GovUKDateInputField():
+                    # Default the date input link to the first field.
+                    href = f"#id_{field_name}_0"
+                case gds_fields.GovUKCheckboxesField():
+                    # Default to the first checkbox item
+                    href = f"#id_{field_name}_0"
+                case gds_fields.GovUKRadioInputField():
+                    # Default to the first radio item
+                    href = f"#id_{field_name}_0"
```
MattHolmes123, line 62: Even though the current fix is the same for all three I didn't feel the need to condense them in to a single block.
The GovUKDateInputField should actually go to the field with the error (e.g. you've populated the first two but not the third field) but I couldn't find a clean way to do it.

However just linking to the first field is fine for now (as stated in the GDS docs)
image

mikeylockwood (approved):

PR #1936, opened 2024-12-09
ECIL-192 - Sanitise importer address field

chris-pettinga:


web/domains/office/model_fields.py
```diff
@@ -0,0 +1,24 @@
...
+from .form_fields import AddressLineFormField
+
+
+class AddressLineField(models.CharField):
+    """Address line field for office addresses"""
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("max_length", 100)
+        kwargs.setdefault("null", True)
+        kwargs.setdefault("blank", True)
```
MattHolmes123, line 13: This is the opposite to djangos defaults for `null` and `blank`, do we even need these two lines as they are the opposite to the expected behaviour?

web/domains/office/models.py
```diff
@@ -12,28 +14,28 @@ class Office(models.Model):
...
-    address_6 = models.CharField(
+    address_6 = AddressLineField(
         max_length=100, null=True, blank=True, verbose_name="Address line 6"
     )
-    address_7 = models.CharField(
+    address_7 = AddressLineField(
         max_length=100, null=True, blank=True, verbose_name="Address line 7"
     )
-    address_8 = models.CharField(
+    address_8 = AddressLineField(
```
MattHolmes123, line 38: I wonder if we actually want to add this logic here at all.
The encoding issue may go away completely with the new API so therefore it should be a form change 🤔 ?

MattHolmes123: If this was added to the form logic it would be way less code too:
```python
    def clean(self):
        cleaned_data = super().clean()
        
        for f in ["address_1", "address_2", "address_3", "address_4", "address_5"]:
            address = cleaned_data.get(f)
            cleaned_data[f] = normalize_address(f)
    
    
def normalize_address(f):
    """Fix the address"""
    ...    
```

MattHolmes123 (changes requested):
I think this logic should live in the form clean method

web/domains/office/models.py
MattHolmes123: As I mentioned in standup normalize_address is really just `normalize_chief_field`

PR #1942, opened 2024-12-24
ECIL-419 - Mailshot should have the name of the recipient auto entered

chris-pettinga:


web/mail/messages.py
```diff
@@ -783,7 +785,7 @@ def __init__(self, *args: Any, mailshot: Mailshot, site_domain: str, **kwargs: A
     def get_context(self) -> dict[str, Any]:
         context = super().get_context()
         context["subject"] = self.get_subject()
-        context["body"] = self.get_body()
```
mikeylockwood, line 786: It looks like the purpose of calling `self.get_body()` is the different definitions of `get_body` on the subclass. I think it would be better to pass the result of `get_body()` to the function rather than the mailshot class itself.

mikeylockwood: Something like `get_mailshot_content(body, recipient)`

web/domains/template/utils.py
```diff
@@ -18,6 +18,7 @@
     User,
 )
 
+from ...mail.types import RecipientDetails
```
mikeylockwood, line 21: Only use relative imports in the same directory. Change this to an absolute path

web/mail/messages.py
```diff
@@ -79,6 +80,7 @@ class GOVNotifyEmailMessage(EmailMessage):
     name: ClassVar[EmailTypes]
 
     def __init__(self, *args: Any, recipient: RecipientDetails, **kwargs: Any) -> None:
+        self.recipient = recipient
         kwargs["to"] = [recipient.email_address]
         super().__init__(*args, **kwargs)
         self.template_id = self.get_template_id()
```
mikeylockwood, line 86: I don't think the `self.first_name` below this line and in the context is actually used for anything

mikeylockwood (approved):

web/mail/messages.py
chris-pettinga: @mikeylockwood used on line 106 of the same file

mikeylockwood: Yes, in the context, but this won't be used anyway as the template variables are replaced before sending the email to gov notify

PR #1945, opened 2024-12-27
ECIL-419 - Mailshot should have the name of the recipient auto entered

chris-pettinga:


web/mail/messages.py
```diff
@@ -79,6 +80,7 @@ class GOVNotifyEmailMessage(EmailMessage):
     name: ClassVar[EmailTypes]
 
     def __init__(self, *args: Any, recipient: RecipientDetails, **kwargs: Any) -> None:
+        self.recipient = recipient
         kwargs["to"] = [recipient.email_address]
         super().__init__(*args, **kwargs)
         self.template_id = self.get_template_id()
```
mikeylockwood, line 86: Remove `self.first_name` from line 86 and `first_name": self.first_name` from line 106. This looks like is for context variables in govuk notify templates, but the variables are replaced before being sent to govuk notify. The govuk notify template just uses the body message and it's own footer.

mikeylockwood (approved):

web/mail/messages.py
chris-pettinga: @mikeylockwood removing this makes 115 tests fail. Looking at the code it looks like `get_personalisation()` is called explicitly to populate the gov notify email template - https://github.com/uktrade/icms/blob/9cf92ed095fa145b0e4c0fafa8285dec4b88911b/web/mail/backends.py#L17

PR #1947, opened 2025-01-06
ECIL-421 - Change colour of banner for lower environments

chris-pettinga:
Turns the header a different colour on lower environments depending on site.

Caseworker:
image

Exporter:
image

Importer:
image

### DIFF
--- a/web/misc/context_processors.py
+++ b/web/misc/context_processors.py
@@ -1,0 +1,18 @@
+from django.conf import settings
+from django.http import HttpRequest
+
+from web.sites import SiteName
+
+
+def header_colour(request: HttpRequest) -> dict[str, str]:
+    """Add the required header colour to the context."""
+    header_colours = {
+        SiteName.IMPORTER: "green",
+        SiteName.EXPORTER: "grey",
+        SiteName.CASEWORKER: "red",
+    }
+
+    return {
+        "CHANGE_HEADER_COLOUR": settings.APP_ENV != "production",
+        "HEADER_COLOUR": header_colours[request.site.name],
+    }
--- a/web/templates/base.html
+++ b/web/templates/base.html
@@ -36,6 +36,16 @@
     {% endcompress %}
 
     {% block css %}{% endblock %}
+    {% if CHANGE_HEADER_COLOUR %}
+      {#
+      changing the colour of the header if we're not on production and depending on the site. 
+      #}
+      <style>
+        #menu-bar {
+          background: {{ HEADER_COLOUR }} !important;
+        }
+      </style>
+    {% endif %}
   </head>
 
   <body>
--- a/config/settings.py
+++ b/config/settings.py
@@ -111,6 +111,7 @@ TEMPLATES = [
                 "web.auth.context_processors.auth",
                 "web.sites.context_processors.sites",
                 "web.ecil.context_processors.govuk_frontend_jinja_template",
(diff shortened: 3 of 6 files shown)


config/cf_env.py
```diff
@@ -173,6 +173,9 @@ class CloudFoundryEnvironment(BaseSettings):
     # Flag to decide if we want to save the PDFs generated as part of the visual regression tests - useful for debugging
     save_generated_pdfs: bool = True
 
+    # Current environment
+    current_environment: str = "local"
```
MattHolmes123, line 177: We do already have `APP_ENV` defined in settings.py that we use throughout ICMS.
I don't think we need another setting to do this

web/misc/context_processors.py
```diff
@@ -0,0 +1,18 @@
...
+def header_colour(request: HttpRequest) -> dict[str, str]:
+    """Add the required header colour to the context."""
+    header_colours = {
+        SiteName.IMPORTER: "green",
+        SiteName.EXPORTER: "grey",
+        SiteName.CASEWORKER: "red",
+    }
+
+    return {
+        "CHANGE_HEADER_COLOUR": settings.CURRENT_ENVIRONMENT != "prod",
```
mikeylockwood, line 16: Why not use` settings.APP_ENV`?

web/templates/base.html
```diff
@@ -36,6 +36,16 @@
...
+    {% if CHANGE_HEADER_COLOUR %}
+      {#
+      changing the colour of the header if we're not on production and depending on the site. 
+      #}
+      <style>
+        #menu-bar {
+          background: {{ HEADER_COLOUR }} !important;
+        }
+      </style>
+    {% endif %}
```
MattHolmes123, line 48: This is the only place we want to do this therefore I think the context processor is overkill.

You could still write tests and check the template for the correct content.

config/settings_local.py
```diff
@@ -51,3 +51,5 @@
 # sometimes we want to run celery tasks synchronously to help with debugging
 CELERY_TASK_ALWAYS_EAGER = env.celery_task_always_eager
 CELERY_EAGER_PROPAGATES_EXCEPTIONS = env.celery_eager_propagates_exceptions
+
+CURRENT_ENVIRONMENT = "local"
```
mikeylockwood, line 55: `CURRENT_ENVIRONMENT ` also have to be added to `settings_test` and `settings`. But best to just remove it and use `APP_ENV`

MattHolmes123 (changes requested):
I like the change but I think we only need to edit a single file (base.html) and use `APP_ENV` instead of using a new setting.

web/templates/base.html
mikeylockwood: `app_env` is globally accessible in templates. See globals in `web/jinja2.py`

chris-pettinga: @mikeylockwood @MattHolmes123 

Agreed with `APP_ENV` however I like the context processor approach as we can test it as a unit of functionality, rather than rely on the frontend tests

mikeylockwood: You can test the content of the html that has been returned in the view. So in the unit tests check the content of the request contains the following for example
 
```
      <style>
        #menu-bar {
          background: green !important;
        }
      </style>
```

mikeylockwood (approved):

PR #1948, opened 2025-01-07
ECIL-347 User Feedback Survey

mikeylockwood:
User feedback survey after submitting an application

Pop up after submitting
image

Survey page
image

Pop up after completing survey
image


web/templates/messages/messages.html
```diff
@@ -1,9 +1,16 @@
 {% macro flash_message(message) %}
+  {% set class_list = ["info-box"] %}
+  {% for tag in message.tags.split() %}
+    {{ class_list.append('info-box-' + tag) }}
+  {% endfor %}
   <div class="flash-message" style="display: block;">
-    <div role="alert" class="info-box info-box-{{message.tags}} " tabindex="0">
```
mikeylockwood, line 3: This broke the stying if more than one tag was applied to a message

web/templates/messages/messages.html
```diff
@@ -1,9 +1,16 @@
...
+    <div role="alert" class="{{ class_list|join(' ') }}" tabindex="0">
       <button class="flash-message-close icon-cross" aria-label="Close this message"></button>
-      {# Do not use the "safe" template filter here #}
```
mikeylockwood, line 5: Need the option of being able to mark a message as safe to provide a link

web/domains/case/utils.py
```diff
@@ -156,11 +156,14 @@ def submit_application(app: ImpOrExp, request: AuthenticatedHttpRequest, task: T
...
+    survey_url = reverse("user-feedback-survey", kwargs={"process_pk": app.pk})
+    survey_message = f'<br/>Help us improve our service by <a class"info-box-link" href="{survey_url}">providing feedback</a>.'
+
     msg = (
         "Your application has been submitted."
-        f" The reference number assigned to this case is {app.get_reference()}."
+        f" The reference number assigned to this case is {app.get_reference()}. {survey_message}"
     )
-    messages.success(request, msg)
+    messages.success(request, msg, extra_tags="safe")
```
mikeylockwood, line 166: Could maybe add a setting to turn this off or on.

web/domains/survey/models.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    issues = ArrayField(
+        models.CharField(max_length=16, choices=IssuesChoices),
+        default=list,
+        blank=True,
+        verbose_name="Did you experience any of the following issues?",
+    )
+    issue_details = models.TextField(
+        blank=True,
+        null=True,
```
MattHolmes123, line 51: The ship has sailed a little bit on this one with our models but django recommends not having two values for no data with text fields:
https://docs.djangoproject.com/en/5.1/ref/models/fields/#null

MattHolmes123, line 51: The ship has sailed a little bit on this one with our models but django recommends not having two values for False with text fields:
https://docs.djangoproject.com/en/5.1/ref/models/fields/#null

MattHolmes123: same applies to all null text fields you have described

web/domains/survey/models.py
```diff
@@ -0,0 +1,78 @@
...
+    service_improvements = models.TextField(
+        blank=True, null=True, verbose_name="How could we improve the service?"
+    )
+    future_contact = models.CharField(
+        max_length=3,
+        choices=YesNoChoices,
+        verbose_name="Would you be happy for us to contact you in the future to help us improve the service?",
+    )
+    site = models.CharField(max_length=60)
+    process = models.ForeignKey("web.Process", on_delete=models.SET_NULL, null=True)
```
MattHolmes123, line 76: Do we want feedback from deleted applications?
Can you even delete a submitted application?

MattHolmes123: withdrawn applications get a status of withdrawn and nothing is deleted.
So this can safely be set to PROTECT

web/domains/survey/urls.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .views import UserFeedbackView
+
+urlpatterns = [
+    path(
+        "<int:process_pk>/",
+        UserFeedbackView.as_view(),
+        name="user-feedback-survey",
+    )
+]
```
MattHolmes123, line 11: namespace this with `app_name`.

MattHolmes123: e.g.  
`app_name = "survey"`
then name can be `"user-feedback" 

`reverse("survey:user-feedback")`

MattHolmes123 (changes requested):
Need to checkout code to suggest a change.

web/templates/web/domains/survey/user-feedback-survey.html
```diff
@@ -0,0 +1,45 @@
...
+              <input
+                type="submit"
+                name="action"
+                class="primary-button button"
+                value="Submit" />
+            </li>
+          </ul>
+        </div>
+      </div>
+    </div>
```
MattHolmes123, line 42: We do have a submit_button macro now.
{{ forms.submit_button(btn_label="Submit") }}

web/static/web/js/pages/user-survey.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  noIssuesCheckbox.addEventListener("change", function (e) {
+    if (noIssuesCheckbox.checked === true) {
+      for (let i = 0; i < 5; i++) {
+        let checkbox = document.querySelector('#id_issues_' + i);
+        checkbox.checked = false;
+      }
+    }
+  });
```
MattHolmes123, line 35: You do not clear the other details (Tell us more about the issue you had) field.
This can preserve previously entered data.
Maybe this is by design?

web/domains/survey/views.py
```diff
@@ -0,0 +1,54 @@
...
+
+from .forms import UserFeedbackForm
+
+
+class UserFeedbackView(LoginRequiredMixin, PermissionRequiredMixin, FormView):
+    template_name = "web/domains/survey/user-feedback-survey.html"
+    form_class = UserFeedbackForm
+    success_url = "/workbasket"
+    page_title = "Survey"
+    permission_required = (Perms.sys.importer_access, Perms.sys.exporter_access)
```
MattHolmes123, line 22: This isn't used if you are overriding has_permission

web/domains/survey/views.py
```diff
@@ -0,0 +1,54 @@
...
+            raise Http404("Survey does not exist.")
+        except AssertionError:
+            raise Http404("Survey does not exist.")
+
+        try:
+            UserFeedbackSurvey.objects.get(process=self.process)
+            messages.success(request, "This survey has already been submitted.")
+            return redirect(reverse("workbasket"))
+        except UserFeedbackSurvey.DoesNotExist:
+            return super().dispatch(request, *args, **kwargs)
```
MattHolmes123, line 39: Most of this could live in has_permission if you wanted to refactor it.

You'd probablly have to override handle_no_permission if you wanted to redirect to the workbasket for all errors though.

web/domains/survey/views.py
```diff
@@ -0,0 +1,54 @@
...
+from web.models import UserFeedbackSurvey
+from web.permissions import AppChecker, Perms
+
+from .forms import UserFeedbackForm
+
+
+class UserFeedbackView(LoginRequiredMixin, PermissionRequiredMixin, FormView):
+    template_name = "web/domains/survey/user-feedback-survey.html"
+    form_class = UserFeedbackForm
+    success_url = "/workbasket"
```
MattHolmes123, line 20: reverse_lazy instead of hardcoding urls that can change.

web/domains/survey/views.py
```diff
@@ -0,0 +1,54 @@
...
+            raise Http404("Survey does not exist.")
+
+        try:
+            UserFeedbackSurvey.objects.get(process=self.process)
+            messages.success(request, "This survey has already been submitted.")
+            return redirect(reverse("workbasket"))
+        except UserFeedbackSurvey.DoesNotExist:
+            return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
```
MattHolmes123, line 41: missing types (for what it's worth).
same in has_permission

MattHolmes123 (commented):
We should add end to end tests for this

web/domains/survey/views.py
```diff
@@ -0,0 +1,52 @@
...
+    def has_permission(self) -> bool:
+        process_pk = self.kwargs["process_pk"]
+        try:
+            process = Process.objects.get(pk=process_pk)
+            app = process.get_specific_model()
+            assert app.submitted_by == self.request.user
+        except AssertionError:
+            return False
+
+        application = Process.objects.get(pk=process_pk).get_specific_model()
```
MattHolmes123, line 51: you already have `app` defined in the try block.

However I think it woulld be better written like this:

```python
    def has_permission(self) -> bool:
        application = Process.objects.get(pk=self.kwargs["process_pk"]).get_specific_model()
        submitted_by_user = application.submitted_by == self.request.user
        can_edit = AppChecker(self.request.user, application).can_edit()

        return submitted_by_user and can_edit
```

MattHolmes123: assert lines can be disabled when running production python so shouldn't be used this way

MattHolmes123 (approved):
Approved if you fix that last comment please

PR #1951, opened 2025-01-08
ECIL-219 Add warning in response prep for importers with no EORI

mikeylockwood:
image

### DIFF
--- a/web/templates/web/domains/case/manage/prepare-response.html
+++ b/web/templates/web/domains/case/manage/prepare-response.html
@@ -6,6 +6,23 @@
 
 {% block main_content %}
   <h3>Response Preparation</h3>
+  {% if case_type == "import" and not process.importer.eori_number and not process.importer_office.eori_number %}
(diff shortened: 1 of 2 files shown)


web/templates/web/domains/case/manage/prepare-response.html
```diff
@@ -6,6 +6,17 @@
 
 {% block main_content %}
   <h3>Response Preparation</h3>
+  {% if case_type == "import" and not process.importer.eori_number %}
```
MattHolmes123, line 9: When sending the eori to CHIEF this is the logic:
```python
def _get_eori_number(importer: "Importer", office: "Office") -> str:
    # Use office override if set
    main_eori_number = office.eori_number or importer.eori_number

    return main_eori_number
```

This code doesn't check if the application office has an override eori set.

MattHolmes123 (approved):
Approved but the template logic should check for the office eori

PR #1964, opened 2025-01-21
ECIL-343 Prevent access requests being closed with open FIRs.

MattHolmes123:
Message shown to ILB if they try to close an AR with open FIRs
caseworker_8008_access_case_6_exporter_close-access-request_

Post example after another fir is created (edge case)
caseworker_8008_access_case_6_exporter_close-access-request_ (1)

### DIFF
--- a/web/domains/case/access/views.py
+++ b/web/domains/case/access/views.py
@@ -373,7 +373,16 @@ def close_access_request(
 
         has_open_approval_request = access_request.approval_requests.filter(
             is_active=True, status=ApprovalRequest.Statuses.OPEN
-        )
+        ).exists()
+
+        has_open_firs = access_request.further_information_requests.filter(
+            is_active=True,
+            status__in=[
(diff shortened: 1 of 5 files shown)


web/domains/case/access/views.py
```diff
@@ -373,6 +373,18 @@ def close_access_request(
 
         has_open_approval_request = access_request.approval_requests.filter(
             is_active=True, status=ApprovalRequest.Statuses.OPEN
+        ).exists()
+
+        has_open_firs = (
+            access_request.further_information_requests.filter(is_active=True)
+            .filter(
```
mikeylockwood, line 380: Can just be in the same filter as `is_active=True`

mikeylockwood (approved):

web/domains/case/access/views.py
MattHolmes123: oops! that was me combing the queryset with code from elsewhere.
Good spot

PR #1967, opened 2025-01-23
ECIL-457 Add session based multistep form and summary view.

MattHolmes123:
Changes:
  - Add base classes to support session based multistep forms.
  - Add example views using the base classes.
  - Add tests
  - Refactor ECIL menu items to appear in subfolder.

Notes:
  - Added several todo to revisit unfinished work.


web/ecil/gds/views/base.py
```diff
@@ -0,0 +1,186 @@
...
+                    },
+                }
+            )
+
+        # TODO: Use pydantic class for summary_list_kwargs
+        return context | {"summary_list_kwargs": {"rows": rows}}
+
+    def form_valid(self, form: django_forms.Form | django_forms.ModelForm) -> HttpResponseRedirect:
+        record = form.save(commit=False)
+        record.created_by = self.request.user
```
MattHolmes123, line 148: This should be in the base class and will be fixed the first time this class actually gets used.

chris-pettinga: @MattHolmes123  I would put in a hook that can be overriden by child classes depending on the behaviour they want to implement. e.g.

```  
record = form.save(commit=False)
record = self.record_post_save_hook(instance)
record.save()
```

web/ecil/gds/views/base.py
```diff
@@ -0,0 +1,187 @@
...
+
+        match value:
+            case True | "True":
+                return "Yes"
+            case False | "False":
+                return "No"
+            case None:
+                return ""
+            case _:
+                return value
```
chris-pettinga, line 168: This should be done (I think) in the HTML template itself using a templatetag - https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#yesno

web/ecil/gds/views/base.py
```diff
@@ -0,0 +1,187 @@
...
+            field_error = f"{field.label}: {','.join(error)}"
+
+            error_list.append({"text": field_error})
+
+        # TODO: This should use a python serializer from gds.components.serializers
+        context["error_summary_kwargs"] = {
+            "titleText": "There is a problem",
+            "errorList": error_list,
+        }
+        return self.render_to_response(context)
```
chris-pettinga, line 187: this function doesn't account for `form.non_field_errors()`

web/ecil/gds/views/utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def save_session_form_data(
+    request: HttpRequest, prefix: str, step: str, form: django_forms.Form | django_forms.ModelForm
+) -> None:
+    # When updating a dict value do this
+    # request.session["foo"]["bar"] = 1
+    # request.session.modified = True
+
+    for field, value in form.cleaned_data.items():
```
chris-pettinga, line 23: I think you'll need to use `form.dirty_data` here because Django has already done some pre-processing to make it cleaned_data which can't be stored in the session, e.g. datetimes.

chris-pettinga (changes requested):
Some comments and questions, thank you.

web/ecil/gds/views/utils.py
MattHolmes123: Yer I've got plans to return to this very soon `# TODO: This will not store all data types`
It's not the final solution.

MattHolmes123: You can also write custom session serializers to handle python types how you want to (instead of storing dirty data)

web/ecil/gds/views/base.py
MattHolmes123: Yep was going to resolve for access requests.

web/ecil/gds/views/base.py
MattHolmes123: Added a comment to return to

PR #1989, opened 2025-02-10
 ECIL-600 - User Survey - Redirect users to user survey page after completing an application

chris-pettinga:


web/domains/survey/models.py
```diff
@@ -55,24 +55,31 @@ class IssuesChoices(TypedTextChoices):
         max_length=16,
         choices=EaseFindChoices,
         verbose_name="How was the process of finding the service?",
+        null=True,
```
MattHolmes123, line 58: having two values to mean empty is an anti pattern for text based fields.

https://medium.com/@kasperjuunge/django-10-anti-patterns-to-avoid-7a2a7a49ba4e#eb0d

MattHolmes123: `default=""` is more idiomatic

MattHolmes123: Same applies to all text based fields.

web/domains/survey/models.py
```diff
@@ -55,24 +55,31 @@ class IssuesChoices(TypedTextChoices):
...
         max_length=3,
         choices=YesNoChoices,
         verbose_name="Would you be happy for us to contact you in the future to help us improve the service?",
+        blank=True,
+        null=True,
     )
+    referrer_path = models.CharField(max_length=255, blank=True, null=True)
     site = models.CharField(max_length=60)
-    process = models.ForeignKey("web.Process", on_delete=models.PROTECT)
+    process = models.ForeignKey("web.Process", on_delete=models.PROTECT, null=True, blank=True)
```
MattHolmes123, line 83: Why is the link now optional? 
Is that to allow feedback not related to an application

chris-pettinga: @MattHolmes123  -  correct

web/domains/survey/models.py
chris-pettinga: @MattHolmes123 - good spot thanks

mikeylockwood (approved):

PR #1991, opened 2025-02-12
ECIL-504 - Ability to upload document the same time as creating a Section 5 Authority

chris-pettinga:
image

image

### DIFF
--- a/web/domains/file/utils.py
+++ b/web/domains/file/utils.py
@@ -52,15 +52,23 @@ class ICMSFileField(forms.FileField):
     mark_help_text_safe = True
 
     def __init__(
-        self, *, validators: Iterable[VALIDATOR] = (), help_text: str = HELP_TEXT, **kwargs: Any
+        self,
+        *,
+        validators: Iterable[VALIDATOR] = (),
+        help_text: str = HELP_TEXT,
+        show_default_help_text: bool = True,
+        **kwargs: Any,
     ) -> None:
-        field_help_text = render_to_string(
-            template_name="forms/icms_file_field_helptext.html",
-            context={
-                "help_text": help_text,
-                "ilb_contact_email": settings.ILB_CONTACT_EMAIL,
-            },
-        )
+        if show_default_help_text:
+            field_help_text = render_to_string(
+                template_name="forms/icms_file_field_helptext.html",
+                context={
+                    "help_text": help_text,
+                    "ilb_contact_email": settings.ILB_CONTACT_EMAIL,
+                },
+            )
+        else:
+            field_help_text = help_text
 
         super().__init__(
             # order is important: validate_file_extension can delete the file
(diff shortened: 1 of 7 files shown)


MattHolmes123 (commented):
If you only fill out the file field and press Create you will see a load of errors for the mandatory fields.

The file you previously chose is now gone and a user will probably fail to add this the second time.

caseworker_8008_importer_1_section5_create_

caseworker_8008_importer_1_section5_create_ (1)

MattHolmes123:
Having an inline file field that you have to use and press the save buttton of the main form is different to everywhere else in ICMS.

Currently file uploads are always separate forms and I'm not sure we should change that even if it is ILB feedback
image

MattHolmes123 (changes requested):
I think we should reject this change.
On one hand it does make adding a single file as part of creating a section 5 authority better but it makes adding subsequent files a lot worse and behave differently to the rest of ICMS.

MattHolmes123:
An alternative might be to only include the file upload as an optional extra field to the create form but not when editing.
After it's created the list of documents, and the ability to add another, should revert to where they were placed previously

chris-pettinga:
@MattHolmes123 I suppose the end-goal here to convert all caseworker file-upload pages to the above pattern so there is consistency, there is a direct need from ILB as we've seen so it feels somewhat obstructive to block just for sake of inconsistency. 

I agree the flow can be improved, I wonder if along with the 'create-only' pattern we also allow for multiple file uploads

MattHolmes123:
@chris-pettinga I don't think we shouldn't change it just for consistency, I just think this is worse than what was there before.
I don't know of a UI where I've been asked to save a set of semi related data just to upload a file.

If it appears on the create only form and then displays the list where it did previously I think that would be better.
And obviously fix the bug where the file is lost.

chris-pettinga:
@MattHolmes123 but in order to create the object in the first place you need to provide those fields? Maybe I'm misunderstanding but in the current world you need to provide all that data before you can upload the file in the first place and so this is no different.

MattHolmes123:
> @MattHolmes123 but in order to create the object in the first place you need to provide those fields? Maybe I'm misunderstanding but in the current world you need to provide all that data before you can upload the file in the first place and so this is no different.

@chris-pettinga I am referring to subsequent uploads.
To add a second file it doesn't feel right to submit all the section 5 data again as part of that edit view.

Including the ability to add a document when you're initially creating the record is a good change.
Subsequent files should be added like they were before this PR is what I am saying.

chris-pettinga:
@MattHolmes123 - I have implemented as suggested, users can now only upload files as part of the form when creating (they can also upload multiple)

web/templates/web/domains/importer/edit-section5-authority.html
```diff
@@ -64,13 +65,11 @@
...
     </div>
   {% endif %}
-
   <div class="container">
     <div class="row">
       <a class="button small-button" href="{{ icms_url('importer-section5-add-document', kwargs={'pk': section5.pk}) }}">
         Add document
       </a>
     </div>
   </div>
```
MattHolmes123, line 74: While you're in the area can you align the Add Document button with the file column please

image

web/domains/file/utils.py
```diff
@@ -125,6 +130,27 @@ def validate_file_extension(file: S3Boto3StorageFile) -> None:
...
+        kwargs.setdefault("widget", MultipleFileInput())
+        super().__init__(*args, **kwargs)
+
+    def clean(self, data, initial=None):
+        single_file_clean = super().clean
+        if isinstance(data, (list, tuple)):
+            result = [single_file_clean(d, initial) for d in data]
+        else:
+            result = [single_file_clean(data, initial)]
+        return result
```
MattHolmes123, line 151: It might be a good idea to include some help text telling the user to select multiple files at the same time:
image

They may not realise they are just overriding the same file if they try and use the input multiple times.

web/templates/web/domains/importer/edit-section5-authority.html
MattHolmes123: Save button looks a little dodgy too if you wouldn't mind fixing that :)

web/templates/web/domains/importer/edit-section5-authority.html
```diff
@@ -41,13 +41,14 @@
...
                   <form method="post"
+                        id="{{ document.pk }}-delete-form"
                         action={{ icms_url(
                           'importer-section5-delete-document',
                           kwargs={
                             'section5_pk': section5.pk,
                             'document_pk': document.pk}) }}>
                     {{ csrf_input }}
-                    <button type="submit" class="link-button icon-bin button">Delete</button>
+                    <button type="submit" class="link-button icon-bin button" form="{{ document.pk }}-delete-form" onclick="">Delete</button>
```
MattHolmes123, line 51: Why do we need this?
`onclick=""`

MattHolmes123: Also noticed it's easy to delete a file without a confirm.
Can we add one of these please:
`data-confirm="Are you sure you want to delete this record?"`

MattHolmes123 (approved):

PR #2002, opened 2025-02-18
Ecil 464

mikeylockwood:


web/data_workspace/urls.py
```diff
@@ -0,0 +1,5 @@
+from django.urls import include, path
+
+app_name = "data-workspace"
+
+urlpatterns = [path("v0/", include("web.data_workspace.v0.urls"))]
```
MattHolmes123, line 5: Curious whats going on here?

web/tests/data_workspace/v0/test_views.py
```diff
@@ -0,0 +1,76 @@
...
+            "email": email,
+            "primary_email_address": email,
+            "department": "prototype_user_dep",
+            "job_title": "prototype_user_job_title",
+            "organisation": "prototype_user_org",
+            "date_joined": "2024-01-20T00:00:00Z",  # /PS-IGNORE
+            "last_login": None,
+            "exporter_ids": [],
+            "importer_ids": [],
+            "group_names": ["ECIL Prototype User"],
```
MattHolmes123, line 75: Shouldn't we add tests for a few users who have linked importers and exporters to check your annotations work?

web/utils/api/auth.py
```diff
@@ -0,0 +1,170 @@
...
+) -> dict[str, Any]:
+    hawk_auth_id = settings.HAWK_AUTH_ID
+    if not constant_time_compare(access_id, hawk_auth_id):
+        raise mohawk.exc.HawkFail(f"Invalid Hawk ID {access_id}")
+
+    return {
+        "id": hawk_auth_id,
+        "key": settings.HAWK_AUTH_KEY,
+        "algorithm": HAWK_ALGO,
+    }
```
MattHolmes123, line 49: This will need changing before we expose this API.

A credentials map should return the hawk_auth_id for a particular application.
We should not be sharing a common `HAWK_AUTH_ID` for ICMS-HMRC and dataworkspace.

Create a ticket for now but add a todo please.

MattHolmes123 (approved):

web/utils/api/auth.py
mikeylockwood: Yes I did have a separate id and key for data workspace originally which is why I added kwargs to this function, but need a bit of a rework to get it to work in the current implementation. Will add a ticket

web/data_workspace/urls.py
mikeylockwood: We'll need versions for the urls to manage breaking changes. This is just creating a `/data-workspace/v0/` path. If we want to make changes to the api in the future we will make a `/data-workspace/v1/` path.

MattHolmes123: My gut is saying we don't bake versioning in at this time (as we don't know if we will need it).

PR #2007, opened 2025-02-24
ECIL-607 Add new access request fields to manage access request view.

MattHolmes123:
**Changes:**
  - Add new fields to manage access request view
  - Fix nullable char fields on access request model

**Screenshots:**
  - ECIL exporter access request
caseworker_8008_access_case_1_exporter_link-access-request_

  - ECIL agent access request
caseworker_8008_access_case_3_exporter_link-access-request_


mikeylockwood (changes requested):
You're making fields that contain null values in prod not null. You'll need to write a custom migration to populate these values before merging

```>>> AccessRequest.objects.filter(organisation_trading_name__isnull=True).count()
14019
>>> AccessRequest.objects.filter(agent_name__isnull=True).count()
11491
>>> AccessRequest.objects.filter(response__isnull=True).count()
6```

mikeylockwood (approved):

mikeylockwood:
> You're making fields that contain null values in prod not null. You'll need to write a custom migration to populate these values before merging
> 
> ```
> >>> AccessRequest.objects.filter(organisation_trading_name__isnull=True).count()
> 14019
> >>> AccessRequest.objects.filter(agent_name__isnull=True).count()
> 11491
> >>> AccessRequest.objects.filter(response__isnull=True).count()
> 6
> ```

Actually the default value should be applied to the existing records, so shouldn't be an issue

MattHolmes123:
> You're making fields that contain null values in prod not null. You'll need to write a custom migration to populate these values before merging
> 
> ```
> >>> AccessRequest.objects.filter(organisation_trading_name__isnull=True).count()
> 14019
> >>> AccessRequest.objects.filter(agent_name__isnull=True).count()
> 11491
> >>> AccessRequest.objects.filter(response__isnull=True).count()
> 6
> ```

I created some records before changing the model fields and running the migrations.
I could successfully apply the migration and view the records in the front end.

If you make a nullable field non null django asks for a default when you create a migration.
This didn't happen this time.
I don't think postgres actually stores Null for string columns.

I am going to test it again though just to be sure we are ok.

PR #2009, opened 2025-02-24
ECIL-465 Add UserFeedbackSurveyData to data workspace api

mikeylockwood:


web/tests/data_workspace/test_views.py
```diff
@@ -4,13 +4,17 @@
...
 
+from web.models import UserFeedbackSurvey
 from web.tests.api_auth import JSON_TYPE, make_testing_hawk_sender
 
 
 def at_example(prefix: str) -> str:
     return f"{prefix}@example.com"  # /PS-IGNORE
 
 
+DT_STRING = "%Y-%m-%dT%H:%M:%S.%fZ"
```
MattHolmes123, line 15: What's the `Z` on the end actually doing?
Looks like it's just hardcoded on the end.
Are you trying to use isoformat because there is a method in the datetime libarary for that.
```python
>>> now = dt.datetime.now()
>>> now.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
'2025-02-25T08:42:36.859620Z'
>>> now.isoformat()
'2025-02-25T08:42:36.859620'
```

web/tests/data_workspace/test_views.py
```diff
@@ -112,3 +115,53 @@ def test_authetication(self):
...
+    def _setup(self, cw_client, cfs_app_submitted, exporter_one_contact):
+        self.client = cw_client
+        self.url = reverse("data-workspace:user-survey-data", kwargs={"version": "v0"})
+        self.app = cfs_app_submitted
+        self.user = exporter_one_contact
+        self.survey = UserFeedbackSurvey.objects.create(
+            satisfaction="happy",
+            issues=["some"],
+            issue_details="",
+            find_service="easy",
```
MattHolmes123, line 131: `easy` isn't a valid value:
```python
class EaseFindChoices(TypedTextChoices):
        VERY_EASY = ("VERY_EASY", "Very Easy")
        EASY = ("EASY", "Easy")
        ...
```
We should create a test record with every column filled in

MattHolmes123: Neither is `satisfaction="happy",`

MattHolmes123: please use the TypedTextChoices

web/tests/data_workspace/test_views.py
```diff
@@ -112,3 +115,53 @@ def test_authetication(self):
...
+            content_type=JSON_TYPE,
+            HTTP_HAWK_AUTHENTICATION=sender.request_header,
+        )
+        assert response.status_code == HTTPStatus.OK
+
+        result = response.json()
+
+        assert len(result) == 1
+        expected = list(UserFeedbackSurvey.objects.values())
+        expected[0]["created_datetime"] = self.survey.created_datetime.strftime(DT_STRING)
```
MattHolmes123, line 165: Do we not want to add a thorough test that at least tests every field for a row?

web/tests/data_workspace/test_views.py
MattHolmes123: So I've realised you've added this because of pydantic.
It doesn't include the `Z` when I've tested it though?
```python
>>> class S(BaseModel):
...     dt_val: dt.datetime
...
>>> foo = S(dt_val=dt.datetime.now())
>>> foo.model_dump(mode="json")
{'dt_val': '2025-02-25T08:54:04.227897'}
```

MattHolmes123: Ahh it's for a timezone aware datetime
```python
>>> tz_aware_dt = timezone.now()
>>> foo = S(dt_val=tz_aware_dt)
>>> foo.model_dump(mode="json")
{'dt_val': '2025-02-25T08:57:11.242223Z'}
```

MattHolmes123 (approved):

web/tests/data_workspace/test_views.py
mikeylockwood: Didn't think it was important as the form to enter data isn't being tested here. Just data being returned from the model

PR #2022, opened 2025-03-03
ECIL-620 Separate icms-hmrc and data workspace hawk api creds

mikeylockwood:
- Split api credentials for icms-hmrc and data workspace
- Subclassed the `HawkAuthMixin` so only the credentials for the configured api are valid


config/settings.py
```diff
@@ -380,6 +380,23 @@
 HAWK_AUTH_ID = env.hawk_auth_id
 HAWK_AUTH_KEY = env.hawk_auth_key
 
+HAWK_ALGORITHM = "sha256"
+HAWK_ICMS_HMRC_API_ID = env.hawk_icms_hmrc_api_id
+HAWK_DATA_WORKSPACE_API_ID = env.hawk_data_workspace_api_id
```
MattHolmes123, line 385: Do the IDs need to be secrets as well as the keys?

MattHolmes123 (approved):

config/settings.py
mikeylockwood: The ids don't need to be secrets, I can take them out of the secrets and just set them directly in settings

PR #2023, opened 2025-03-04
ECIL-636 Add several model field to gds/form/fields.py.

MattHolmes123:
Changes:
  - Add GovUKCheckboxesModelField.
  - Add GovUKRadioInputModelField.
  - Add GovUKSelectModelField.
  - Add GovUKSelectModelField field example.
  - Add ECILExampleListView list view.
  - Add tests for new fields.


web/ecil/forms/forms_example.py
```diff
@@ -136,6 +140,26 @@ def clean(self) -> dict[str, Any]:
...
+        error_messages={"required": "Select a commodity code"},
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["foreign_key_field"].queryset = Commodity.objects.filter(
+            commodity_code__startswith="500500"
+        )
+
+        # hardcode unsupported fields so save works.
```
mikeylockwood, line 156: What's going on here?

web/ecil/gds/component_serializers/checkboxes.py
```diff
@@ -31,7 +34,7 @@ class CheckboxItem(TextOrHTMLMixin, BaseModel):
     # Specific name for the checkbox item. If omitted, then component global name string will be applied.
     name: str | None = None
     # Required. Value for the checkbox input.
-    value: str
+    value: str | int | ModelChoiceIteratorValue
```
mikeylockwood, line 37: Worth making a type for input fields so it only needs to be updated in one place?

mikeylockwood (approved):

web/ecil/forms/forms_example.py
MattHolmes123: So I created a model with every field django supports and made them all required...
So I just needed to save some random values to get the record to save in the db.

PR #2027, opened 2025-03-07
WIP ECIL-647 Add nuclear material app type

MattHolmes123:
Public branch containing all NMIL changes.


web/domains/case/_import/models.py
```diff
@@ -32,42 +33,59 @@ class SubTypes(TypedTextChoices):
...
+        on_delete=models.PROTECT,
+        blank=False,
+        null=False,
+        related_name="declaration_application_types",
+    )
+    endorsements = models.ManyToManyField(
+        "web.Template", related_name="endorsement_application_types"
+    )
+
+    # Field not used by ICMS - Will be deleted in future
```
mikeylockwood, line 52: Thanks for looking into all these redundant fields

PR #2028, opened 2025-03-07
Mark new IAT record as inactive for tests.

MattHolmes123:


web/management/commands/add_nuclear_material_app.py
```diff
@@ -8,14 +8,11 @@ class Command(BaseCommand):
     via django migrations due to the test data using fixtures that a populated only once the migrations have run"""
 
     def handle(self, *args, **options):
-        try:
-            ImportApplicationType.objects.get(type="NMIL")
+        if ImportApplicationType.objects.filter(type="NMIL").exists():
```
mikeylockwood, line 11: I do prefer try: .get for a single record, but thats just preference

PR #2031, opened 2025-03-11
ECIL-656 Make eori_number field editable for individual importers and no longer allow GBPR

mikeylockwood:


web/domains/importer/forms.py
```diff
@@ -17,35 +17,29 @@
...
-                raise ValidationError(
-                    f"'{eori_number}' must be set to {self.EORI_INDIVIDUAL} for individual importers."
-                )
+            # Example value: GB123456789012345
+            eori_number_length = len(eori_number[2:])
+
+            if eori_number_length != 12 and eori_number_length != 15:
+                raise ValidationError("Must start with 'GB' followed by 12 or 15 numbers")
 
             return eori_number
```
MattHolmes123, line 44: The ImporterOrganisationForm already has a `clean_eori_number` method:

```python
    def clean_eori_number(self):
        """Make sure eori number starts with GB."""
        if eori_number := self.cleaned_data.get("eori_number"):
            prefix = "GB"

            if not eori_number.lower().startswith(prefix.lower()):
                raise ValidationError(f"'{eori_number}' doesn't start with {prefix}")

            # Example value: GB123456789012345
            eori_number_length = len(eori_number[2:])

            if eori_number_length != 12 and eori_number_length != 15:
                raise ValidationError("Must start with 'GB' followed by 12 or 15 numbers")

            return eori_number
```

That version checks for the prefix, both versions should be the same now right?

web/tests/domains/importer/test_forms.py
```diff
@@ -45,15 +45,23 @@ def test_required_fields_importer_individual_form():
...
-    assert importer.eori_number == ImporterIndividualForm.EORI_INDIVIDUAL
+    assert importer.eori_number == "GB123456789012"
+
+
+@pytest.mark.parametrize("eori", [("",), ("GBPR",), ("GB12345678901",), ("GB1234567890123")])
+@pytest.mark.django_db()
+def test_individual_importer_eori_validation(importer_one_contact, eori):
+    data = {"user": importer_one_contact.pk, "eori_number": eori}
+    form = ImporterIndividualForm(data)
+    assert form.is_valid() is False
```
MattHolmes123, line 63: `assert not form.is_valid()` ?

MattHolmes123 (changes requested):

MattHolmes123 (approved):

PR #2038, opened 2025-03-31
ECIL-663 Nuclear materials email template

mikeylockwood:


web/admin.py
```diff
@@ -189,6 +190,8 @@ class ICMSSiteAdmin(SiteAdmin):
 admin.site.register(SanctionsAndAdhocApplication)
 admin.site.register(SanctionsAndAdhocApplicationGoods)
 admin.site.register(EmailTemplate, EmailTemplateAdmin)
+# TODO: Remove NuclearMaterialEmail from admin site when merged with SanctionEmail
```
mikeylockwood, line 193: Just adding `NuclearMaterialEmail` model to the admin site for now to quickly be able to manage email address tot send nuclear material case emails to. My idea is to merge the data with SanctionEmail in a new model called CaseEmailAddress and manage all case email addresses there rather than have a model fro each.

MattHolmes123 (approved):

PR #2049, opened 2025-04-08
ECIL-690 Adding NMIL commodities to import licence report

mikeylockwood:


web/reports/interfaces.py
```diff
@@ -869,13 +884,18 @@ def get_ima_type_title(self, ima_type: str) -> str:
             return ima_type
 
     def get_commodity_codes(self, ia: dict) -> str:
-        commodity_code = ia["commodity_code"]
+        commodity_code = ia["commodity_code"]  #
```
MattHolmes123, line 887: Remove this:
`  #`

PR #2052, opened 2025-04-15
ECIL-698 Update office model to only have 5 addres lines.

MattHolmes123:
Changes:
  - Remove address_6, address_7 & address_8 from all places they are used.
  - Update V2 form to not include extra fields.
  - Update ECIL new exporter office form.


web/migrations/0069_remove_office_address_6_remove_office_address_7_and_more.py
```diff
@@ -0,0 +1,25 @@
...
+            name="address_6",
+        ),
+        migrations.RemoveField(
+            model_name="office",
+            name="address_7",
+        ),
+        migrations.RemoveField(
+            model_name="office",
+            name="address_8",
+        ),
```
mikeylockwood, line 24: Will this temporarily break anything that accesses the fields?The migrations are run in the db before the code is updated

MattHolmes123: Practically I don't think anything will go wrong as these fields are only used in a few places.
To be safe though I will spilt the removal of the fields to another PR to be merged in the next release.

mikeylockwood (approved):

PR #2077, opened 2025-05-09
ECIL-628 Add export application data to data workspace api

mikeylockwood:


web/data_workspace/urls.py
```diff
@@ -17,4 +17,39 @@
...
+    path(
+        "<dwversion:version>/cfs-products/",
+        views.CFSProductDataView.as_view(),
+        name="cfs-product-data",
+    ),
+    path(
+        "<dwversion:version>/legislations/",
+        views.LegislationDataView.as_view(),
+        name="legislation-data",
+    ),
```
MattHolmes123, line 54: Consider refactor to remove dwversion duplication?
`path("<dwversion:version>/", include([...])`

web/data_workspace/views/case.py
```diff
@@ -0,0 +1,46 @@
...
+            ),
+            "application_id": Case(
+                When(
+                    export_application_certificates__export_application_id__isnull=True,
+                    then="import_application_licences__import_application_id",
+                ),
+                default="export_application_certificates__export_application_id",
+                output_field=IntegerField(),
+            ),
+        }
```
MattHolmes123, line 46: I think you should test this with an application with several variation requests.
One application can have multiple document packs and you might only want the active one?
```
# From web/domains/case/services/document_pack.py

def pack_active_get(application: ImpOrExp) -> DocumentPack:
    """Get the current active document pack."""

    return _get_qm(application).get(status=PackStatus.ACTIVE)
    
def _get_qm(application: ImpOrExp) -> Manager[DocumentPack]:
    if application.is_import_application():
        return application.licences
    else:
        return application.certificates    
```

MattHolmes123 (approved):
Added a few comments.

PR #2079, opened 2025-05-13
ECIL-756 Add import application model to data workspace api

mikeylockwood:


web/data_workspace/serializers/base.py
```diff
@@ -114,6 +114,9 @@ class ApplicationBaseSerializer(BaseSerializer):
 
     # Application fields
     status: str
+    applicant_reference: str
```
MattHolmes123, line 117: I thought applicant_reference was an optional field?

PR #2081, opened 2025-05-15
ECIL-757 Add NMIL and Sanctions to data workspace api

mikeylockwood:


web/data_workspace/serializers/base.py
```diff
@@ -25,12 +25,15 @@
     str | None: ("String", True),
 }
 
+ASDECIMAL_ANNOTATIONS = (Decimal, Decimal | None)
```
MattHolmes123, line 28: should the following have an extra underscore:
  - ASDECIMAL_ANNOTATIONS
  - asdecimal
To this:
  - AS_DECIMAL_ANNOTATIONS
  - as_decimal

MattHolmes123 (approved):

web/data_workspace/serializers/base.py
MattHolmes123: Ahh ok

